PR #4, opened 2016-06-07
[SofaKernel] Fix XML export

guparan:
As suggested by Alex Bilger in SOFA forum: https://www.sofa-framework.org/community/forum/topic/duplicated-components-in-xml-export/

> Hello guys,
> 
> I noticed a bad behavior when exporting a scene in xml. The components are duplicated within the same node. The node components are written, then the sub-nodes, and then the same node components again.
> You can get the same behavior by opening any sofa scene and exporting it (File -> Save As in runSofa), then open it in a text editor.
> I see that the xml export uses visitors. I found that the objects are processed twice: once in processNodeTopDown and once in processNodeBottomUp. So I removed the processing part in processNodeBottomUp, and kept only the node closure. I get the expected behavior. Could you guys confirm this is the right thing to do?
> I tested on Windows with a very recent commit.

### DIFF
--- a/SofaKernel/framework/sofa/simulation/XMLPrintVisitor.cpp
+++ b/SofaKernel/framework/sofa/simulation/XMLPrintVisitor.cpp
@@ -123,17 +123,6 @@ Visitor::Result XMLPrintVisitor::processNodeTopDown(simulation::Node* node)
 
 void XMLPrintVisitor::processNodeBottomUp(simulation::Node* node)
 {
-    for (simulation::Node::ObjectIterator it = node->object.begin(); it != node->object.end(); ++it)
-    {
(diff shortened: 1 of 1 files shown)


guparan:
Comment from Alex Bilger on SOFA Forum:

> I also would like to highlight a problem with xml export: when exporting, we loop over the objects and write all the data in this object. This is not correct as some component compute things differently if some data are provided or not. This means the initialization is not the same with the exported scene than in the initial one. I don’t have a fix for that. Maybe by writing only the provided data? But that would mean to write file names instead of the position/faces in the case of a MeshLoader for instance.

matthieu-nesme:
I think I created trouble in the commit ef8b7f08.
Thanks for letting us notice that!

Should be fixed by commit b80b8777

PR #10, opened 2016-07-27
[framework] FIX : StateMask method clean needs to resize m_size to 0

Ngautier:
Commit 9b8dcf05 cleaning the code of StateMask forgot to implement clean method, just add resize at 0


Ngautier:
Will be redone, wrong commit

PR #12, opened 2016-08-22
Fix: inclusion test in MeshNewProximityIntersection

RomainTestylier:


guparan:
One remark about commented blocks: couldn't they be simply removed if not used anymore? 
Would be better to avoid introducing commented/unmaintained/why-is-this-here code.

matthieu-nesme:
I agree with Guillaume, but it is a branch so the consequences are limited.
In a near future, this branch will only be managed in a fork and deleted from the public repo.

PR #16, opened 2016-08-30
Remove bezier high order

hdeling:
Removal of Bezier tetrahedron elements and Bezier triangle elements : move to a new plugin SofaHighOrderTopology


matthieu-nesme:
I guess this will break the Flexible plugin compilation.
Do you plane to commit it back in a public plugin? In which case, Flexible could link with this plugin.

hdeling:
Hi,

In fact, I have created an external plugin SOFA-HighOrderTopology which 
is intended to be made public.

Either this plugin can be added in the list of "official plugins" such 
as Compliant or handled completely separately from the SOFA git hub 
repository.

I would like to hear your opinion on where this should go. Either 
situations is fine with me.

Best,

Herve

matthieu-nesme:
Hi Hervé,

In general, I think it is a good idea to have a repo per plugin, even if it becomes more difficult to update them when the sofa core is modified.
But in this specific case, it would be great to add this plugin to this github repo, so Flexible can still benefit from your code.

Another solution, could be to add a third plugin in your independent repository, that would link with both SOFA-HighOrderTopology and Flexible, and that would contain the high order shape functions.

What do you think?

thomas-lemaire:
yet an other option is to conditionnaly compile HighOrderTopology code in Flexible based on whether the HighOrderTopology plugin is existing or not (in cmakelists package_require(HighOrderTopology) if (HighOrderTopology_FOUND) ... ): 
- each plugin can live in its own repository 
- no need for a third plugin 
- should work whether all the plugins are compiled in the same cmake build project or are compiled/installed separately (well HighOrderTopology must be available when compiling Flexible) 

thomas 

----- Mail original -----

hdeling:
OK I agree that this allows to compile Flexible without the SOFA-HighOrder.

Before moving SOFA-HighOrder to the gitHub repo, do any of you know if 
there are plans to move those plugins outside of github ?

My understanding is that at some point SOFA was going to be cut into 
several pieces that could be optionally compiled.

Is this still on the radar ?

Hervé

thomas-lemaire:
The questions of conditional compilation (plugin selection at cmake), and repository split are distinct, the idea is flying in the air, but nothing concrete, the main issue (I think) is integrating the several repositories in a single continuous integration platform, not a big deal, but it has to be done 

thomas 

----- Mail original -----

damienmarchal:
Hi all,

I see no conclusion on this specific pull-request ... does it still have to be opened or doest it have to be closed (or merged) ? 

DM.

matthieu-nesme:
This merge is not possible.

Hervé, first you have to know your own fork (or a more global fork you could create for asclepios) can be compiled in the sofa continuous integration system.

To include your work about high order topology in the sofa-public w/o breaking existing stuff, my suggestion is to create a new plugin in your fork, then to propose it here as a pull request. Once you'll have created the PR, we'll be able to possibly fix the glue with Flexible.

PR #18, opened 2016-09-19
fixes compilation with SOFA_NO_OPENGL flag

raphaeldeimel:
This commit fixes compilation of a basic configuration with the additional SOFA_NO_OPENGL flag being set, including compilation of Compliant and SofaPython plugins

Tested for runSofa -gui batch


matthieu-nesme:
This could help Frederick who is cleaning the opengl part.

fredroy (commented):
The first 3 modifications should be useless by now. (using drawtools)
The last one is fixing a bug though so you could you just commit PickHandler.cpp ?

PR #19, opened 2016-09-19
FIX: explicitily link rt library when on unix systems

raphaeldeimel:
Not sure if this is correct, but it seems that sofa uses the rt library  (high resolution clock?) on unix, but does not explicitly specify the library when linking.  Because of that, compilation may fail on some systems.

This patch just makes dependency on the rt library explicit in cmake


matthieu-nesme:
By curiosity, which are the concerned systems?

raphaeldeimel:
The system where I noticed it is CentOS 6.3.

The compile system is custom though, gcc is 6.1.0 and binutils 2.27 to be specific.

I wasn't really digging into the cause, the linker complained and I just added the corresponding flag.

matthieu-nesme:
Ok thanks.

`build start`

guparan:
[ci-build]

guparan:
[ci-build]

guparan:
[ci-build]

matthieu-nesme:
I do not know how to modify the PR.

It does not link on macos.
The cmake condition could be replaced with something like 
`if(UNIX AND NOT APPLE)`

matthieu-nesme:
tried in 03b042feb4483a1ec4caa0e525b4ef95fa93d7b1

PR #22, opened 2016-09-22
[Sensable] Fix compilation error when using a const context node for the creation of the visual node

jnbrunet:
This is a small correction to fix the compilation of the Sensable plugin.

Also, any idea why we keep the OmniDriver when there is a NewOmniDriver available?


FabienPean:
@hugtalbot Is plugin Sensable now obsolete with the new Geomagic plugin out ?
Otherwise these fix were solving my issues at the time.

PR #46, opened 2016-10-21
[Compliant] added easy python mappings

maxime-tournier:
This PR introduces nice python wrappers for mappings. 

It enables easy definition of pure python mappings similar to c++ with little overhead, since all the Eigen sparse matrices are aliased as SciPy sparse matrices (yay!).

Here is a small taste:

``` python
class DistanceMapping(easy.Mapping):
    '''a hooke-like distance mapping: 

    f(p1, p2) = norm(p2 - p1) - rest_length
    '''

    def __init__(self, node, **kwargs):
        easy.Mapping.__init__(self, node, **kwargs)

        self.rest_length = 0

    def apply(self, out, at):

        p1 = at[0][0]
        p2 = at[1][0]

        delta = p2 - p1
        norm = math.sqrt(delta.dot(delta))

        out[0][0] = norm - self.rest_length


    def jacobian(self, js, at):

        p1 = at[0][0]
        p2 = at[1][0]

        delta = p2 - p1
        norm = math.sqrt(delta.dot(delta))

        if norm > 1e-14:
            u = delta / norm
            js[0][0] = -u
            js[1][0] = u
        else:
            # warning
            js[0][0] = np.zeros(3)
            js[1][0] = np.zeros(3)


    def geometric_stiffness(self, gs, at, force):

        p1 = at[0][0]
        p2 = at[1][0]

        delta = p2 - p1

        norm = math.sqrt(delta.dot(delta))
        if norm > 1e-14:

            u = delta / norm
            f = force[0][0]

            block = (np.identity(3) - np.outer(u, u)) * (f / norm)
            gs[:3, :3] = block
            gs[3:, 3:] = block

            gs[:3, 3:] = -block
            gs[3:, :3] = -block


        else:
            gs[:, :] = 0
```

See `Compliant/examples/easy.py` for a complete example. For now the code lives in `Compliant` for other people to try out, but could migrate to `SofaPython` if needed. 

This PR also ditches a previous but less elegant/powerful attempt at the same problem.


matthieu-nesme:
[ci-build]

PR #62, opened 2016-11-22
Display 3D text

matthieu-nesme:
Adding "Visual3DText" to display a camera-oriented (billboard) 3D text.

It uses glText::draw, that could be improved a bit.


fredroy:
Yep, drawText could be enhanced, especially by changing the texture containing the letters.
Or directly generating the bitmaps by using libfreetype (http://learnopengl.com/#!In-Practice/Text-Rendering)

PR #74, opened 2016-11-29
WARNING: fix missing r in SofaSparseSolver

Ngautier:
Just a little fix in SparseLDLImplInvertData, r of sparse is missing. 
Fix needed for Cuda plugins update in sofa-dev.


guparan:
The same error appears in SparseLUSolver...

guparan:
Let's fix all the "sparse" misspellings at once. I'm gonna add some commit here.

PR #75, opened 2016-11-29
Flexible strain smoothing

BenjaminGilles:
[Flexible] add engines and examples to smooth strains, for faster convergence + minor cleaning on quadrature engines


matthieu-nesme:
error

PR #110, opened 2016-12-20
[VolumetricRendering] Explicit incompatibility with SOFA_NO_OPENGL

hugtalbot:
If the flag SOFA_NO_OPENGL is active, the plugin VolumetricRendering
cannot be compiled. Add the proper CMake warning and error message.


fredroy:
Good job !

PR #122, opened 2017-01-09
[SofaKernel] Fix FileRepository should not be optional

thomas-lemaire:
FileRepository is used by too many components to be optional.

As a consequence boost filesystem and locale are made required
This commit fixes undefined symbol error occurring at link time when the filesystem or locale boost lib are not installed

### DIFF
--- a/SofaKernel/SofaFramework/CMakeLists.txt
+++ b/SofaKernel/SofaFramework/CMakeLists.txt
@@ -146,13 +146,14 @@ endif()
 set(SOFA_HAVE_PNG ${PNG_FOUND})
 
 ## Boost
+# optional boost libraries
 find_package(Boost COMPONENTS thread system date_time filesystem locale QUIET)
-
 set(SOFA_HAVE_BOOST_SYSTEM ${Boost_SYSTEM_FOUND})
 set(SOFA_HAVE_BOOST_THREAD ${Boost_THREAD_FOUND})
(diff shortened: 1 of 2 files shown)


guparan:
Thanks for this nice complement to #99 !

damienmarchal:
[ci-build]

damienmarchal:
This PR: 
  - [ ] succeed on all compilations setup on the CI.
  - [x] is now 1 week old.

thomas-lemaire:
On the dash 
http://www.sofa-framework.org/dash/index.php?limit=10&pg=1&user=&branch=pr%2Ffix_filerepository&platform= 

I do not see an information to be able to fix any compilation pb, I might not be looking in the right place ? Can you point me the correct dash url ? 

thomas 

----- Mail original -----

damienmarchal:
@thomas-lemaire You are right, the dashboard and the CI were malfunctionning together resulting in bad reports. But @guparan and @bcarrez spend time to fix it and it should work better now. 

I restart your build to refresh the status...
[ci-build]

EDIT3: now the PR build summary is ok. Correctly indicating that the vs2013 failed. In general a failure is not caused by the PR but more often by a bug in the build system...so there is still a need to investgate that. I'm really sorry thomas  for the bothering.

damienmarchal:
[ci-build]

damienmarchal:
@bcarrez Can I merge this one ? The Failure seems related to our broken vs2013.

PR #129, opened 2017-01-11
[CI] Add VS-2013 support

bcarrez:
This PR adds support for Visual Studio 13 in the ci scripts. VS2013 builds (on ci) won't work without this.
As soon as the builds of every non-VS2013 configurations are ok on this branch (see https://www.sofa-framework.org/dash/index.php?branch=origin%2Fci-vs2013 ) , I believe it is save to merge.

### DIFF
--- a/scripts/ci/compile.sh
+++ b/scripts/ci/compile.sh
@@ -6,7 +6,7 @@
 ## Significant environnement variables:
 # - CI_MAKE_OPTIONS       # additional arguments to pass to make
 # - CI_ARCH               # x86|amd64  (32-bit or 64-bit build - Windows-specific)
-# - CI_COMPILER           # important for Visual Studio (VS-2012 or VS-2015)
+# - CI_COMPILER           # important for Visual Studio (VS-2012, VS-2013 or VS-2015)
 
 # Exit on error
 set -o errexit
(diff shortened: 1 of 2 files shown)


guparan (commented):
I'm ok with your code but it would be a good opportunity to factorize these two blocks don't you think?
I propose 
```bash
if [[ "$(uname)" != "Darwin" && "$(uname)" != "Linux" ]]; then
    # Call vcvarsall.bat first to setup environment
    if [ "$CI_COMPILER" = "VS-2015" ]; then
        vcvarsall="call \"%VS140COMNTOOLS%..\\..\\VC\vcvarsall.bat\" $CI_ARCH"
    elif [ "$CI_COMPILER" = "VS-2013" ]; then
        vcvarsall="call \"%VS120COMNTOOLS%..\\..\\VC\vcvarsall.bat\" $CI_ARCH"
    else
        vcvarsall="call \"%VS110COMNTOOLS%..\\..\\VC\vcvarsall.bat\" $CI_ARCH"
    fi
    echo "Calling $COMSPEC /c \"$vcvarsall & nmake $CI_MAKE_OPTIONS\""
    $COMSPEC /c "$vcvarsall & nmake $CI_MAKE_OPTIONS"
else
```

PR #138, opened 2017-01-17
[image] Fixes #136 : Use the cmake install DIRECTORY instead of FILES…

### LINKED ISSUE #136: Image plugin installation does not preserve the file's hierarchy
This issue was first discovered by @rosaliepl : https://www.sofa-framework.org/community/forum/topic/error-with-cimg-plugin-with-dependancy-to-skeleton-h-does-not-compile/#post-8397

When installing SOFA's image plugin (ex.: make install), the image plugin installation does not preserve the file's hierarchy which will failed the compilation of any external project that includes the SOFACImg.h file.

`/.../sofa/build/install/include/CImg/CImg.h:9341:21: fatal error: plugins/skeleton.h: No such file or directory`

If we look in the installation folder (sofa/build/install), all the required headers are in include/CImg instead of include/CImg/plugins which will then failed the compilation.

To reproduce the bug

1. Compile and install SOFA with the image plugin activated
2. Create an external project with the image plugin as a requirement, and add the following line in any of the source file: `#include <CImg/SOFACImg.h>`

jnbrunet:
Fixes #136

### DIFF
--- a/applications/plugins/image/extlibs/CImg/CMakeLists.txt
+++ b/applications/plugins/image/extlibs/CImg/CMakeLists.txt
@@ -2,27 +2,14 @@ cmake_minimum_required(VERSION 3.1)
 
 project(CImg)
 
-install(FILES 
-    CImg.h
-    SOFACImg.h
-    plugins/add_fileformat.h
-    plugins/chlpca.h
-    plugins/draw_gradient.h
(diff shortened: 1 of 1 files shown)


thomas-lemaire:
I thought this cmake file was provided by the CImg library, but this is not the case, good !

PR #147, opened 2017-01-25
[Compliant] minor modifs/fix

matthieu-nesme:
This PR  contains some minor fixes and adds two new components named: UniformStiff & DiagonalStiffness

Checklist: 
  - [x] succeed on each compilation setup on the CI.
  - [x] does not generates new test failure. 
  - [x] introduces new component with tests & documentation. 


damienmarchal:
I have not the competence nore time to validate it on the mechanical aspect so I suggest to let 1 week to anyone to give feedback, and after one week we merge it :)

damienmarchal:
@matthieu-nesme, @maxime-tournier 

Thanks for your PR. The cleaning looks ok to me but it seems there is two new components (UniformStiffness, DiagonalStiffness) added in this PR. 

Maybe you should change the title of the PR to "Minor Fix and Add UniformStiff & DiagonalStiffness Components" 

I don't see any test on the two added components. 

You changed the .py file to use the two added component. As they seems very similar to the existing one maybe this deserve a bit of documentation (doxygen) or example on when to use one component instead of the other otherwise users will be puzzled in deciding which one to choose.

damienmarchal:
[ci-build]

damienmarchal:
[ci-build]

damienmarchal:
Sorry to be so long to merge this PR but the CI is very slow and buggy which really breaks the workflow. 

@matthieu-nesme, @maxime-tournier can you please give me your thinking about adding tests and documentations on the components introduced in this PR. 

EDIT: took the checklist to add it in the PR description.

matthieu-nesme:
@damienmarchal Even if these new components are trivial, you are right and I did push some tests for them.

damienmarchal:
[ci-build]

damienmarchal:
[ci-build] I restart because of the segfault on linux machine (I'm quite sure they are not connected with PR...but just to be sure before merging).

PR #157, opened 2017-01-27
Add instantiation of constraint corrections with  Vec2f

courtecuisse:
Add instantiation of constraint corrections with Vec2f


guparan:
[ci-build]

PR #192, opened 2017-03-03
[SofaHelper] drawFakeSphere: fix ambient component in shaders

fredroy:
Small patch to restore ambient component in fake sphere rendering.

Was:
manyspheres_00000002

Now:
manyspheres_00000001

This PR: 
- [x] has been reviewed 


matthieu-nesme:
wahoo, a description with pictures, that's a first!

PR #194, opened 2017-03-06
FIX: sofa installation. 

epernod:
PR #139 moved licence.txt into license.LGPL.txt. This changed has not been propagated in the cpack config.

I didn't create a branch as only one line is modified. Hope it doesn't break the CI...

This PR: 


epernod:
btw, maybe one VM of each OS should do a make install to test the cpack config.

guparan:
Thank you Erik!
Indeed, we are not installing enough in our CI. Continuous delivery could be a nice task to start in the future (after our Jenkins 2 migration). 😉

PR #204, opened 2017-03-15
[simulation] adding listener mechanism to SceneLoader

matthieu-nesme:
in order to be informed when a scene is loaded.

WARNING: each SceneLoader's loading function must call "notifyLoadingScene".

This PR: 


damienmarchal (commented):
It was much easier than what I expected. I think it is ok.
Can you merge this branch into the one with the reload so that we can use it and see how it fit ?

PR #262, opened 2017-05-04
[SofaDenseSolver] Fix compilation with SOFA_FLOATING_POINT_TYPE as float

FabienPean:
issue sofa-framework/sofa#2 Fix compilation when option SOFA_FLOATING_POINT_TYPE in CMake selected as float

The problem arising when trying to compile with float was due to missing declaration from newmat external library.
The library has been modified for sofa to use a #define SOFA_FLOAT in include.h
However the SOFA_FLOAT was never defined for the extlib since the current procedure is to set CMake variable SOFA_FLOAT and SOFA_DOUBLE to 1. This was done anyway AFTER processing the cmake of the extlibs.

Therefore, this pr moves the option for floating point type in sofa in the root CMakeLists and adds the #define for the newmat library to actually create the float type functions.

This PR: 


damienmarchal:
[ci-build]

guparan:
Dashboard details are here: https://www.sofa-framework.org/dash/index.php?branch=pr%2Ffix%2Fissue%232

guparan:
Hi @FabienPean, 
Any idea why RigidLinearDeformationMappings_test is failing on Windows ?
See [windows7_VS-2015_amd64_origin_options tests report](https://ci.inria.fr/sofa-ci/job/windows7_VS-2015_amd64_pr_2_options/266/testReport/).

EDIT: This is certainly nothing related to your PR. It is also failing for sofa-framework/sofa#261.

EDIT 2: The problem is known, see issue sofa-framework/Flexible#4.

SofaKernel/extlibs/newmat/CMakeLists.txt
```diff
@@ -42,6 +42,9 @@ add_library(${PROJECT_NAME} STATIC ${HEADER_FILES} ${SOURCE_FILES})
 target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>")
 target_include_directories(${PROJECT_NAME} PUBLIC "$<INSTALL_INTERFACE:include>")
 target_compile_definitions(${PROJECT_NAME} PRIVATE "use_namespace")
+if(SOFA_FLOAT)
+    target_compile_definitions(${PROJECT_NAME} PRIVATE "SOFA_FLOAT")
+endif()
```
matthieu-nesme: I do not understand why this is necessary?
Is a dependency to a sofa component missing?

matthieu-nesme: oh it is an extlib....
using a sofa macro in an extlib????

FabienPean: I initially tried the method advocated on the newmat library FAQ [see 2.3 Customizing](http://robertnz.net/nm11.htm). It was not working, because the newmat/include.h file was modified by sofa developers at some point, who also added the sofa macro inside, see line 31. In the end, SOFA_FLOAT in include.h is only a proxy to other define USING_DOUBLE and USING_FLOAT, so it could probably be removed.

matthieu-nesme: hum, looks like something has to be cleaned more deeply in the sofa cmake files to integrate newmat the right way.

FabienPean:
I removed the sofa define from newmat/include.h and corrected the CMakeFiles. Now it activates float only when "float" is explicitly selected in root CMakeFiles

damienmarchal:
[ci-build]

FabienPean:
Ok I did not expect this problem. I understand now better what Matthieu meant.

When I removed SOFA_DOUBLE from newmat, the target_compile_definitions are only visible within newmat project with the PRIVATE attribute, which was causing problem when the newmat.h was used included in SofaDenseSolver, not having the compile definitions set. It works if the compile definition of newmat are PUBLIC, but I do not know the side effects it may cause.

FabienPean:
[ci-build]

damienmarchal:
Thanks Fabian for the quick update of the PR...
[ci-build]

guparan:
Did you try compiling with `SOFA_FLOATING_POINT_TYPE = float`? 
It is still failing on my side (Windows + VS-2015).

The Dashboard is OK but compiling with `SOFA_FLOATING_POINT_TYPE = both`.

FabienPean:
I did not manage to compile it fully. SofaDenseSolver is now ok but not the remaining of Sofa. I started to fix some errors, however it seems that this feature was not actively maintained recently. 
With latest commit it should compile but it gives weird runtime visualization. (Windows VS2015 also)
image
image
image

If the feature is a must-have, maybe the CI should test with float-only and double-only instead of "both" ?

guparan:
Thank you for your fixes @FabienPean !

Two remarks about consistency:
- `Sofa_test` default template is `SReal`. I would go for `Sofa_tests<>` inheritance. Unless there is a specific reason for inheriting from `Sofa_tests<SReal>` ?
- `#ifndef SOFA_FLOAT` block usually comes before `#ifndef SOFA_DOUBLE` block.

I can push these tiny corrections if you're ok 😉

FabienPean:
Hi Guillaume, sure feel free ! :)

damienmarchal:
The 'leaking' snake looks great :)

guparan:
Here they are !

damienmarchal:
It is unclear to me if it is ready. Isn't the screenshot from @FabienPean showing something is broken ?

guparan:
At least the compilation is fixed so I'd say this PR does its job. I had the same visualization glitches on my side though. Maybe a quick review to ensure we did not create the problem - @fredroy ?

SofaKernel/extlibs/newmat/CMakeLists.txt
```diff
@@ -41,7 +41,12 @@ set(SOURCE_FILES
 add_library(${PROJECT_NAME} STATIC ${HEADER_FILES} ${SOURCE_FILES})
 target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>")
 target_include_directories(${PROJECT_NAME} PUBLIC "$<INSTALL_INTERFACE:include>")
-target_compile_definitions(${PROJECT_NAME} PRIVATE "use_namespace")
+target_compile_definitions(${PROJECT_NAME} PUBLIC "use_namespace")
+if(NOT DEFINED SOFA_WITH_DOUBLE)
```
matthieu-nesme, line 45: To be fully clean, it would be great to add this code where the newmat dir in included in cmake.
The idea is that nothing would be required when updating newmat.

modules/SofaGeneralEngine/DilateEngine.cpp
```diff
@@ -38,18 +38,18 @@ SOFA_DECL_CLASS(DilateEngine)
...
 int DilateEngineClass = core::RegisterObject("Dilates a given mesh by moving vertices along their normal.")
 // TriangleOctree implemented with double only
-//#ifdef SOFA_WITH_FLOAT
-//.add< DilateEngine<Vec3fTypes>>(true)
-//#endif //SOFA_FLOAT
+#ifdef SOFA_WITH_FLOAT
+.add< DilateEngine<Vec3fTypes>>(true)
+#endif //SOFA_FLOAT
 #ifdef SOFA_WITH_DOUBLE
 .add< DilateEngine<Vec3dTypes>>(true) // default template
```
matthieu-nesme: warning: two default templates when compiling with both float and double

SofaKernel/extlibs/newmat/CMakeLists.txt
FabienPean: Some points to keep in mind regarding newmat:
* It has not been updated for 9 years or so now.
* It does not offer CMakeLists officially.
* Configuration is meant to be done inside newmat/include.h

Thus the CMakeLists available has been created by Sofa developers, and newmat internals have been modified by Sofa developers too. So if there is an update, the include.h will have to be updated, and the CMakeLists file will remain the same.

Therefore is it still required to have it above in the build tree ? (It would then be located in .\sofa\src\SofaKernel\extlibs\CMakeLists.txt )

matthieu-nesme: ok got it, thanks for your remarks.
Let's merge it then!

damienmarchal:
Rebuild from a recent master... [ci-build]

FabienPean:
My bad it seems I was quite asleep when I tried to fix the DilateEngine.cpp

damienmarchal:
[ci-build]

damienmarchal:
So it seems there is some troubles here in compiling this PR. Anyone can check ?

FabienPean:
The problem comes when using 'both' it seems. For one part of the build failure, TriangleOctree / TriangleOctreeRoot must be instantiated with both precision instead of currently using the default type SReal only.

fredroy:
I built it on VS2015, and got the same behavior as on the pictures.
It appears to be a bug with SparseGridRamification (in caduceus.scn)
If you replace it with SparseGridTopology, everything appears fine (and behaves fine)

fredroy:
As for the glitches with the caduceus, I built it on VS2015, and got the same behavior.
It appears to be a bug with SparseGridRamification.
If you replace it with SparseGridTopology, everything appears fine (and behaves fine)

damienmarchal:
Many thanks @FabienPean for this PR as it seems quite complex to fix everything. 
[ci-build]

FabienPean:
Ok so, hopefully the end of this PR. I reverted some commit and double/both should be back to normal. In the case of float only the DilateEngine is not available and its test deactivated. So basically it should compile fine, but some components would be limited.
I quickly tried to fix it but I encountered some difficulties. I would prefer open a new issue regarding float compliance of remaining components.

To sum up, it should be back to the original goal: fix compilation, not necessarily the components with type limitation

damienmarchal:
The CI disagree with you :)

FabienPean:
I am on Windows and did not try on Linux, so my hopes still hold as long as that yellow dot does not become a cross for the windows VS build :D

The error on the linux build seem trivial fortunately. However it stopped at the first error... Would it be possible for the build system not to stop at this first error and try compiling the remaining ? Since I cannot test on every platform/os that would be valuable to have all potential error at once.

matthieu-nesme:
Regarding the visualization problems when compiling with float only: keep in mind that double precision can be necessary in specific situations, so be careful when replacing some hard-coded `double` with `SReal`. Some `double` must remain `double` and I guess some `SReal` should become `double` also in the actual code.
On platforms where double precision does not exist, tricky numerical solutions have to be employed.

guparan:
I rebased your branch to clean it a bit (removing all merges and reverts).
Let's see how it builds...
https://www.sofa-framework.org/dash/index.php?branch=pr%2Ffix%2Fissue%232

guparan:
[ci-build]

guparan:
Are we good to go with this PR? 
@matthieu-nesme, do you see a specific line where a `SReal` should remain a `double`?

matthieu-nesme:
There is no easy answer to that, numerical limits are really complex.
But obviously, some double are mandatory somewhere in the components that map the snake model ^^

damienmarchal:
[ci-build]

PR #291, opened 2017-06-06
Fix CMakeLists bug on Sofa.ini and installedSofa.ini creation

epernod:
Same as PR #284 but without the unwanted merge commits

Small fix related to issue #114 .

Have tested the compilation without the flag SOFA_INSTALL_RESOURCES_FILES (which is set by default), Sofa.ini and installedSofa.ini, which are necessary for build and install, are well created.

This PR: 


guparan:
Thanks for the cleaning 😄 
[ci-build]

guparan (approved):

PR #296, opened 2017-06-09
FIX compilation for LeapMotion plugin due to moved files

hugtalbot:
VisualTransform moved from SofaBaseVisual to SofaGeneralVisual

ChangeLog:
- FIX: compilation LeapMotion plugin (issue since v16.12)

This PR: 


guparan:
Well, that is a fully labeled PR! 
LGTM (ignoring VS-2015 failure due to side effect).

PR #299, opened 2017-06-10
[MultiThreading] FIX: examples installation

Innokentiy-Alaytsev:
Example scene files are in "scenes" directory and not in "examples" directory,
as expected by CMake install command.

This PR: 


Innokentiy-Alaytsev:
Hello!

I'm not sure about what is better: to fix CMakeLists.txt or to rename the directory with example scenes. Can someone clarify this for me?

guparan:
Hi @WilkGardariki,
Thanks for this fix. Let's start the [ci-build]. 
About your question, I guess the simpler is the better. There are other plugins with a scenes folder (CGal, Geomagic).
I am also interested in a _examples vs scenes_ discussion. I personally prefer _scenes_ as I find it more accurate.

matthieu-nesme:
I would rather rename every "scenes" dir in "examples" to follow plugin guidelines.

Innokentiy-Alaytsev:
Hello!

@guparan , what do you think about matthieu-nesme's suggestion? Renaming a directory is not a big deal, I can do it instead of fixing CMakeLists.txt.

guparan:
I'm ok with the renaming but please make sure to add a WARNING in your commit message as it may break some algos for MultiThreading users.

Innokentiy-Alaytsev:
Hello!

Sorry for taking so much time for this - I've taken fancy on my primary work :)

About pull request: I've renamed example scene directory and fixed documentation. I hope no one mind that I've removed the previous version of the commit.

guparan:
[ci-build]

Issue #308, opened 2017-06-25
SparseGrid : translation imposed


hugtalbot:
A bug has been reported by a user (Cristóbal) in the forum : https://www.sofa-framework.org/community/forum/topic/vertices-from-obj-are-moved/

The SparseGrid seems to shift the mesh. This might be due to "max" Data is related to the BBox.
Whatever, strange behaviour.

epernod:
Note: To test the bug: Demo/CollisionMultiple.scn related to #270

hugtalbot:
Hi @epernod 

Indeed if you look at the demo scene _examples/Demos/collisionMultiple.scn_ you see:
`<RegularGrid nx="6" ny="5" nz="2" xmin="-7.5" xmax="7.5" ymin="-6" ymax="6" zmin="-1.75" zmax="1.75" />
`
But in fact, when you run the scene and look at the position field in the associated _MechanicalObject_ in the z direction (for instance), z ∈ [27.25 ; 30.75] instead of [-1.75 ; 1.75] !
weird !

epernod:
Yes because the mechanicalObject is translating the topology:
`<MechanicalObject dx="0" dy="20" dz="29" rz="64" />`
z= 29 - 1.75 = 27.25

The problem is in fact that the position in the regularGrid are not transformed... which is weird as they should be updated by the data graph.

hugtalbot:
yes sorry you're right! my bad!
But the initial point was about the SparseGrid. It seems to shift positions for no reasons : 
```
<SparseGridTopology name="sparseGridTopology3"  fileTopology="mesh/cube.obj"  n="2 2 2" />
<MechanicalObject template="Vec3d" name="mObject1"  position="0 0 0"  velocity="0 0 0"  force="0 0 0"  externalForce="0 0 0"  restScale="1"  showObject="0"  showObjectScale="1"  showIndices="0"  showIndicesScale="0.002" />
<FixedConstraint template="Vec3d" name="fixedConstraint3"  indices="0 1 4 5" />
```
You will see the cube (obj file with coordinates between -1 and 1), shifted : coordinates are between -1.02 and 1.02

epernod:
I just checked, this is the same for other topologies. For example if you take the scene: Demos/TriangleSurfaceCutting.scn and add change the mechanicalObject like this:
`<MechanicalObject translation="10 0 0"/>`

in MeshGmshLoader: x[0] = [0; 10; 0]  // logical
[...]

PR #324, opened 2017-07-05
[examples] FIX scenes with bad RegularGrid position

epernod:
This PR is linked to the issue #308 .
Since the modification of MechanicalObject reinit method in PR #270 ([this commit](https://github.com/sofa-framework/sofa/pull/270/commits/b6750f06ea0c5082f8c88d5aaea6bc65d8840429) ). The mechanicalObject doesn't change the value of P0. Thus, it is not necessary anymore to shift the OglModel and the Surf mechanicalObj.

Several scenes were impacted but without ground truth it is hard to say if the scene is totally fixed.

This PR: 


guparan:
Hi @epernod, thank you for your fix.
I will backport it in 17.06 asap :-)

PR #368, opened 2017-08-21
[SofaPython/core] forward sys.argv to python scripts

maxime-tournier:
Changelog:
- added a mechanism to obtain extra args from command-line (everything following `--argv`)
- forwarded extra args to python in `SceneLoaderPy`

Example:
```python
# test.py
def createScene(node):
    import sys
    print(sys.argv)
```

```sh
runSofa -a test.py --argv --spam --bacon eggs
# ['test', '--spam', '--bacon', 'eggs']
```
___________________________________________________

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/helper/ArgumentParser.cpp
+++ b/SofaKernel/framework/sofa/helper/ArgumentParser.cpp
@@ -34,6 +34,8 @@ namespace helper
 
 typedef std::istringstream istrstream;
 
+ArgumentParser::extra_type ArgumentParser::extra;
+
 ArgumentBase::ArgumentBase(char s, string l, string h, bool m)
     : shortName(s)
     , longName(l)
@@ -86,9 +88,7 @@ ArgumentParser::~ArgumentParser()
 */
 void ArgumentParser::operator () ( int argc, char** argv )
 {
-    std::list<std::string> str;
-    for (int i=1; i<argc; ++i)
-        str.push_back(std::string(argv[i]));
+    std::list<std::string> str(argv + 1, argv + argc);
     (*this)(str);
 }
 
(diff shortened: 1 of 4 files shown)


SofaKernel/framework/sofa/helper/ArgumentParser.cpp
```diff
@@ -105,16 +108,20 @@ void ArgumentParser::operator () ( std::list<std::string> str )
...
-
+            if( globalHelp.size()>0 ) std::cout<< globalHelp <<std::endl;
+            
             std::cout << "(short name, long name, description, default value)\n-h,\t--help: this help" << std::endl;
             std::cout << std::boolalpha;
             for( ArgVec::const_iterator a=commands.begin(), aend=commands.end(); a!=aend; ++a )
                 (*a)->print();
             std::cout << std::noboolalpha;
+
+            std::cout << "--argv [...]\t" << "forward extra args to the python interpreter" << std::endl;
```
guparan, line 119: This help message makes me think `--argv` is useful only with python. Isn't it something more generic?

guparan:
**Warning**: there are [two failing unit tests](https://ci.inria.fr/sofa-ci/job/ubuntu_gcc-5.4_options/712/testReport/).

SofaKernel/framework/sofa/helper/ArgumentParser.cpp
maxime-tournier: The PR started from discussion in #356, so I sticked to that. I don't quite see where else it could be useful but if needed, extra arguments can be accessed via the static function `ArgumentParser::extra_args()`

maxime-tournier:
Note: empty `ArgumentParser::extra_args()` will not set `sys.argv`, as opposed to set it some empty list (this was the reason why the test was failing: it would reset it to empty on script loading after being set initially from the test binary).

I'm not quite happy with this kind of side-effect, but this matches the previous behaviour so `¯\_(ツ)_/¯`

maxime-tournier:
build is good, tests are good, gentle bump :-)

PR #369, opened 2017-08-21
[runSofa] Make names of scene object parameters and links selectable

sergeiNikolaev:
This fix allows to select and copy parameters names in Sofa GUI interface, which is very useful when setting default values for elements in scene files. Instead of setting parameters names as titles of widgets I created QLabel objects and set them as the first row of the GridLayout manager.

I know that current sofa GUI will be replaced with Qt quick GUI soon, but I guess this modification will still be useful for those who use this version of GUI. I also would to make this change for Qt quick GUI, but all parameters names in QML are covered with "Mouse Area" objects, which don't allow to handle mouse events for text objects under them.

This PR: 

### DIFF
--- a/applications/sofa/gui/qt/QDisplayDataWidget.cpp
+++ b/applications/sofa/gui/qt/QDisplayDataWidget.cpp
@@ -168,7 +168,6 @@ QDisplayDataWidget::QDisplayDataWidget(QWidget* parent,
         QLabel* propertyName = new QLabel(data_->getName().c_str(), this);
         propertyName->setAlignment(Qt::AlignLeft);
         propertyName->setTextInteractionFlags(Qt::TextSelectableByMouse);
-        propertyName->setContentsMargins(10,10,10,10);
         gridLayout_->setVerticalSpacing(0);
         gridLayout_->addWidget(propertyName, 0, 1);
         gridLayout_->setAlignment(propertyName, Qt::AlignVCenter);
--- a/applications/sofa/gui/qt/QDisplayDataWidget.cpp
+++ b/applications/sofa/gui/qt/QDisplayDataWidget.cpp
@@ -173,7 +173,7 @@ QDisplayDataWidget::QDisplayDataWidget(QWidget* parent,
         gridLayout_->setAlignment(propertyName, Qt::AlignVCenter);
         datawidget_->setContentsMargins(0, 0, 0, 0);
         gridLayout_->setContentsMargins(10,10,10,10);
-        gridLayout_->addWidget(datawidget_, 1, 1);
+        gridLayout_->addWidget(datawidget_, 0, 2);
 
         //setColumns(numWidgets_); //datawidget_->numColumnWidget());
     }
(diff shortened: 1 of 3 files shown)


damienmarchal:
Hi @kriptian88,

Thank you very much for your contribution and welcome in the sofa development community. 
I will test it asap.

guparan:
You forgot to [ci-build] Damien ;-)

guparan:
Just tested it:

Before |  After 1 | After 2 
:-------:|:---------:|:--------:|
before | after 1 | after 2

damienmarchal:
The feature is useful but the visual result is problematic. Isn't it possible to implement similar behavior without touching the rendering ?

sergeiNikolaev:
Hello.
I have removed the margins around QLabel that present parameters names to make the size of components smaller.
Anyway, answering to your questions I could mark these issues.
1. Unfortunately, I am not an expert in Qt. But I guess some rather tricky manipulations with layouts should be done to make the rendering absolutely the same.
2. On different systems the rendering is different. I have a notebook with DebianStretch and desktop with Ubuntu 16.04 and I can say that Sofa GUI differs. The changes I've made, unfortunately, affect GUI differently. So, I guess it depends also on how QtWidgets are rendered on different platforms.

Thus, if rendering structure is critical, may be it is better not to touch it at all.


P.S. I have changed my nickname to make it more reasonable.

guparan:
Thanks @sergeiNikolaev, I just updated my first comment with a new screenshot 😉

damienmarchal:
Hi @sergeiNikolaev 

Respecting the existing gui rendering is not crucial in itself..but we should try to respect standard guidelines for GUIs. On that aspect it is more common to have QLabel & checkbox/textfield on the same line otherwise this look a bit weird and consume a lot of screen space for nothing. 

EDIT: But if this is too much work I think we can live with that as runSofa should die :)

sergeiNikolaev:
Hello @damienmarchal.
Ok, I've got you. No, this is not a lot of work.
I wanted to say that I do not know how to replace Widget title with QLabel and remain rendering system the same. In any other case, I guess, there should be a discussion about positions of these QObjects.

As for me, these QLabels present the names of the parameters and they had been placed instead of titles. In this case it is more natural to see the title a bit upper and then the value of parameters. But ok, I can put them on the same line.

damienmarchal:
Well, runSofa was not really following guidelines on that :)

sergeiNikolaev:
Hello. 
Sorry for not responding quickly. But after I put QLabel in the same line there are still some problems with elements and grey color areas alignment. I need more time to fix that.
guipic

sergeiNikolaev:
I've talked with Bruno Marques, and he said that the version of Sofa GUI without these changes also has problems with color areas alignment. Taking this into account I guess it's better to create a separate issue about this alignment problem. Thus, this branch is ready for another review.

guparan:
[ci-build]

guparan:
SOFA-dev meeting report: @damienmarchal will merge this PR with #373 since their contents are related. Doing so, he will try to get rid of this box problem simply by removing all boxes (separating Datas with lines). To be continued...

guparan:
Postponed because of lack of time. Do not hesitate to re-open.

PR #379, opened 2017-08-30
[CI] FIX scene-test segfault on timeout (Windows)

guparan:
Unix timeout command is now accessible in Windows VMs (using "Git for Windows" Bash).

This PR: 


guparan:
[ci-build] [with-scene-tests]

PR #384, opened 2017-08-31
[SofaBoundaryCondition] FIX Removed incorrect warning from LinearForceField

vmagno:
The LinearForceField component does not contribute anything to the matrix when solving a system. Accordingly, the addKToMatrix function does nothing. However, since it is not implemented, it displays a warning when running a simulation. This PR simply adds an empty implementation of that function, which removes the warning displayed by the definition in the base class.

This PR: 


guparan:
Hi @vmagno, thank you for your PR!
This should be merged fast.
Could you add a simple description please?
[ci-build]

PR #386, opened 2017-09-05
[All] FIX removed compilation warnings

vmagno:
There was an unused variable in a frequently included header, and a missing override in another one.

This PR: 


guparan:
Thank you for these cleans ;-)
[ci-build]

PR #399, opened 2017-09-14
[SofaKernel] FIX bug in toEulerVector

raffaellatrivisonne:

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/helper/Quater.inl
+++ b/SofaKernel/framework/sofa/helper/Quater.inl
@@ -488,8 +488,8 @@ void Quater<Real>::quatToAxis(defaulttype::Vec<3,Real> & axis, Real &angle) cons
 
     Real sin_half_theta; // note that sin(theta/2) == norm of the imaginary part for unit quaternion
 
-    // to avoid numerical instabilities of acos for theta < 5°
-    if(q[3]>0.999) // theta < 5° -> q[3] = cos(theta/2) > 0.999
+    // to avoid numerical instabilities of acos for theta < 5Â°
+    if(q[3]>0.999) // theta < 5Â° -> q[3] = cos(theta/2) > 0.999
     {
         sin_half_theta = sqrt(q[0] * q[0] + q[1] * q[1] + q[2] * q[2]);
         angle = (Real)(2.0 * asin(sin_half_theta));
@@ -523,8 +523,8 @@ defaulttype::Vec<3,Real> Quater<Real>::quatToRotationVector() const
 
     Real sin_half_theta; // note that sin(theta/2) == norm of the imaginary part for unit quaternion
 
-    // to avoid numerical instabilities of acos for theta < 5°
-    if(q[3]>0.999) // theta < 5° -> q[3] = cos(theta/2) > 0.999
+    // to avoid numerical instabilities of acos for theta < 5Â°
+    if(q[3]>0.999) // theta < 5Â° -> q[3] = cos(theta/2) > 0.999
     {
         sin_half_theta = sqrt(q[0] * q[0] + q[1] * q[1] + q[2] * q[2]);
         angle = (Real)(2.0 * asin(sin_half_theta));
@@ -559,7 +559,7 @@ defaulttype::Vec<3,Real> Quater<Real>::toEulerVector() const
 
     Quater<Real> q = *this;
         q.normalize();
-        double y = std::max(-1.0,std::min(1.0,2*(q[3]*q[1] - q[2]*q[0])));
+        Real y = std::max(Real(-1.0),std::min(Real(1.0),Real(2.)*(q[3]*q[1] - q[2]*q[0])));
 
         defaulttype::Vec<3,Real> vEuler;
         vEuler[0] = atan2(2*(q[3]*q[0] + q[1]*q[2]) , (1-2*(q[0]*q[0] + q[1]*q[1])));   //roll
--- a/SofaKernel/framework/sofa/helper/Quater.inl
+++ b/SofaKernel/framework/sofa/helper/Quater.inl
@@ -555,11 +555,15 @@ defaulttype::Vec<3,Real> Quater<Real>::toEulerVector() const
 ///    Pitch: rotation about the Y-axis
 ///    Yaw: rotation about the Z-axis
 
+
+
     Quater<Real> q = *this;
         q.normalize();
+        double y = std::max(-1.0,std::min(1.0,2*(q[3]*q[1] - q[2]*q[0])));
+
         defaulttype::Vec<3,Real> vEuler;
         vEuler[0] = atan2(2*(q[3]*q[0] + q[1]*q[2]) , (1-2*(q[0]*q[0] + q[1]*q[1])));   //roll
-        vEuler[1] = asin(2*(q[3]*q[1] - q[2]*q[0]));                                    //pitch
+        vEuler[1] = asin(y);                                     //pitch
         vEuler[2] = atan2(2*(q[3]*q[2] + q[0]*q[1]) , (1-2*(q[1]*q[1] + q[2]*q[2])));   //yaw
         return vEuler;
 }
--- a/SofaKernel/framework/sofa/helper/Quater.inl
+++ b/SofaKernel/framework/sofa/helper/Quater.inl
@@ -547,25 +547,24 @@ defaulttype::Vec<3,Real> Quater<Real>::quatToRotationVector() const
 }
 
 
+/// Compute the Euler angles:
+/// Roll: rotation about the X-axis
+/// Pitch: rotation about the Y-axis
+/// Yaw: rotation about the Z-axis
 template<class Real>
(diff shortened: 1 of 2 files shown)


guparan:
Thanks for your PR @raffaellatrivisonne :-)
Could you add a small description explaining the error and your fix please?
[ci-build]

damienmarchal:
[ci-build]

guparan:
@raffaellatrivisonne could you fix the build failure please?

damienmarchal:
[ci-build]

raffaellatrivisonne:
The function asin is defined in [-1,1].
The fix prevents NAN when the argument is slightly >1 due to numerical errors (1,000000000000001).
Hope it's clear enough.

guparan:
From SOFA-dev meeting: 
Hi @raffaellatrivisonne, is there a chance that the argument `y` is rounded to 1 or -1 due to anything else than a very small numerical error? If yes, we should add an epsilon to ensure the error is acceptable.

untereiner:
I think I get your point but it adds an another parameter that changes nothing in practice. Here the point is to cut everything outside the bounds. With your suggestions, who will test the sensibility of this epsilon parameter ? It is not acceptable to have a number outside these bounds since asin is not defined at all

guparan:
`asin` might be undefined, it is still better to have it crashing instead of working were it should not - again, if `y` contains a "big" error (greater than 10e-15 upon @raffaellatrivisonne comment). 
The question is: could `y` contain such a "big" error? If no, let's just assume that and round this way. If yes or don't know, let's add a tiny security with something like
```cpp
const double epsilon = 1e-15;
Real y = Real(2.)*(q[3]*q[1] - q[2]*q[0]);
if( std::abs( double(y) ) - 1.0 > epsilon )
{
    msg_error("Quat") << "Unexpectedly out of bounds argument for asin: " << y << msgendl;
}
```
or
```cpp
const double epsilon = 1e-15;
Real y = Real(2.)*(q[3]*q[1] - q[2]*q[0]);
if( std::abs( double(y) ) - 1.0 > epsilon )
{
    Real force_round = std::max( Real(-1.0), std::min(Real(1.0), y) );
    msg_warning("Quat") << "Unexpectedly out of bounds argument for asin: " << y
                        << "Force rounding to " << force_round << msgendl;
    y = force_round;
}
```

untereiner:
Why 1e-15 ? why not 1e-11 ? I guess Raphaella's number was just an example. I completely disagree with this idea. Introducing a new random chosen very locally defined epsilon is really a bad idea. But if it a command from the consortium guy's to close this pr I have to bow

guparan:
There is no command and I am far to pretend knowing enough how things are done in SOFA to claim my proposal is what has to be done. It is just a proposal discussed during last SOFA-dev meeting.

untereiner:
Well, the message sounded like "you have to add an epsilon value" 🙄 . I let @raffaellatrivisonne do her job here for now on... good luck

raffaellatrivisonne:
Hi everybody, 
sorry for being late with my answer.
@guparan, as soon as the argument of asin is greater than 1, you have an invalid operation.
It doesn't matter how "big" is your error, it can be huge or infinitesimal.
In this case, it is due to numerical errors, that's why I said 10e-15.
If it is the proper way, I can set an epsilon, but then how ?
As @untereiner  says, why 1e-15 instead of 1e-11 ?
How do you set a parameter on numerical errors ?

fjourdes:
Just passing by, I do not want to raise a flame war but actuall asin method has some documentation,
espcially when it comes to domain error:
http://www.cplusplus.com/reference/cmath/asin/
Just quoting; 

> If a domain error occurs:
> - And math_errhandling has MATH_ERRNO set: the global variable errno is set to EDOM.
> - And math_errhandling has MATH_ERREXCEPT set: FE_INVALID is raised.

Then looking at 
http://www.cplusplus.com/reference/cmath/math_errhandling/
The default behavior for math_errhandling is MATH_ERRNO, so as the doc suggest you can just check for the errno (thread-local) global variable value, and if it is set to EDOM after asin is called, then you can throw whatever error message you want.

This is probably not relevant here, since I presume the checks are there because in theory when a quaternion is normalized the value of 
`Real(2.)*(q[3]*q[1] - q[2]*q[0])` 
should always belong to the range [-1;1]. The only reason it might not be is for some numerical drifting issues (?)
Provided this assumption is correct you are indeed totally allowed to clamp the values there.

guparan:
Thanks for the explanations @raffaellatrivisonne and @fjourdes, let's merge then ! :-)

damienmarchal:
[ci-build]

damienmarchal:
Do someone knows why do tests are failing ? (In QuaterTest)

raffaellatrivisonne:
I'm taking a look.

raffaellatrivisonne:
I'm trying to debug to see where it fails, but it will take a little bit more than expected.
Meanwhile I just want to highlight that coming back to the un-fixed version (the one without my commit) the test doesn't fail YET the **toEulerVector** is doing an invalid operation (the bug that my commit is supposed to fix).

screenshot from 2017-11-10 12-13-39

guparan:
So if I understood well we have 2 problems here:
- `q1` and `p1` should not be `nan`
- `EXPECT_EQ(p0,p1);` should not pass with `nan` values for `p1`

raffaellatrivisonne:
The inital problem is the vector (that in my screenshot I call "q0 euler") which has its second component which is NaN.
It comes from the unfixed **toEulerVector**  "q0 euler"=  q0.Quater<double>::toEulerVector()
So, it doesn't  surprises me that q1 and p1 are NaN, As they derive from:
        Quater<double> q1 = Quater<double>::createQuaterFromEuler(q0.Quater<double>::toEulerVector());
   
        sofa::defaulttype::Vec<3,double> p1 = q1.Quater<double>::rotate(p);

But definetively EXPECT_EQ(p0,p1) should not pass with nan values for p1

fjourdes:
maybe I am stating something already well known, but with c++11 there are some built in functions that can help to test floating point arithmetic. 
So with the current implementation adding something to the `EulerAngle` test like
```lang=cpp
for(std::size_t i=0; i<q0.size(); ++i) // same goes for q1
{
    ASSERT_FALSE(std::is_nan(q0[i]));
}
```

raffaellatrivisonne:
Hi everybody, I'm trying to work on this PR, but I'm quite busy with my PhD in this moment and I don't think I will be able to finish it within a short delay.
As suggested by Francois, I added the error message in Quater_test.cpp.
Now is failing, as with the old-code (without my commit) NaN values may appear. 
In Quater.inl (function toEulerVector) I went back to the old code commenting the modifications I made with my commit. This way, if someone else takes the hands on this PR, he will better know what to do.

guparan:
Hey, I just cleaned this oooooold PR.
Should be working fine :tada:

SofaKernel/framework/sofa/helper/Quater.inl
```diff
@@ -547,21 +547,24 @@ defaulttype::Vec<3,Real> Quater<Real>::quatToRotationVector() const
...
-        return vEuler;
+    q.normalize();
+
+    // Cancel numerical drifting by clamping on [-1 ; 1]
+    Real y = std::max(Real(-1.0), std::min(Real(1.0), Real(2.)*(q[3]*q[1] - q[2]*q[0])));
+
+    defaulttype::Vec<3,Real> vEuler;
+    vEuler[0] = atan2(2*(q[3]*q[0] + q[1]*q[2]) , (1-2*(q[0]*q[0] + q[1]*q[1])));   //roll
+    vEuler[1] = asin(y); // pitch
+    vEuler[2] = atan2(2*(q[3]*q[2] + q[0]*q[1]) , (1-2*(q[1]*q[1] + q[2]*q[2])));   //yaw
```
epernod, line 566: as it is done for asin, shouldn't we add a test on atan2?
I'm not expert in this, are all values possible for atan2? quick search gave me that : 

> Systems supporting symbolic mathematics normally return an undefined value for atan2(0, 0)

raffaellatrivisonne: If I'm not wrong, the scenario atan2(0,0) cannot occur given the arguments definition.

The equation is: vEuler[0] = atan2(2*(q[3]*q[2] + q[0]*q[1]) , (1-2*(q[1]*q[1] + q[2]*q[2])))
A generic quaternion has at least a non-zero value, which means that
even if q[1] = 0 and q[2]=0, you will have atan2(0,1) which exists and it's defined.

The second argument could be equal to zero, if and only if (q[1]*q[1] + q[2]*q[2])=1/2;
In this case,  q[1] and q[2] would be non-zero (since they solve the above equation) but also q[3] and q[0] would be not zero since the quaternion must be unitary.
So, in this case, you would  have atan2(something non zero, 0).

I'm not an expert, so please correct me if I'm wrong.

epernod: ok for me

PR #421, opened 2017-09-27
[SofaFramework] CMake:Fix and clean boost, when using Sofa as an external lib

fredroy:
Fix regression (appeared when boost was supposed to be removed from Sofa).
-> Plugins _not able to find Boost automatically_ failed to compile (because it needs boost to compile obviously)
 
The problem was coming on Windows mainly, as boost are often installed in the system directories on Unix system.

And few changes were added to make boost's CMake part more consistent. 

This PR: 


fredroy:
It would be nice to cherry-pick it into the 17.06 branch when merged as well.

PR #439, opened 2017-10-03
Add stop in add_mbktomatrixVisitor

olivier-goury:
Adds a stop when the visitor encounters a mapping

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/simulation/MechanicalMatrixVisitor.h
+++ b/SofaKernel/framework/sofa/simulation/MechanicalMatrixVisitor.h
@@ -214,9 +214,10 @@ public:
         return RESULT_CONTINUE;
     }
 
-    virtual bool stopAtMechanicalMapping(simulation::Node* /*node*/, core::BaseMapping* map)
+    virtual bool stopAtMechanicalMapping(simulation::Node* node, core::BaseMapping* map)
     {
-        return !map->f_mapMatrices.getValue();
+        SOFA_UNUSED(node);
+        return !map->areMatricesMapped();
     }
 };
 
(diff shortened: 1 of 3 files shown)


SofaKernel/framework/sofa/simulation/MechanicalMatrixVisitor.h
```diff
@@ -213,6 +213,11 @@ class SOFA_SIMULATION_CORE_API MechanicalAddMBK_ToMatrixVisitor : public Mechani
 
         return RESULT_CONTINUE;
     }
+
+    virtual bool stopAtMechanicalMapping(simulation::Node* /*node*/, core::BaseMapping* map)
+    {
```
EulalieCoevoet: Hi, 
Just a tiny comment, you should use "SOFA_UNUSED(node);"

guparan:
Hi @olivier-goury, is this a fix? Could you add a bit of context please? Thanks :-)

SofaKernel/framework/sofa/simulation/MechanicalMatrixVisitor.h
```diff
@@ -213,6 +213,11 @@ class SOFA_SIMULATION_CORE_API MechanicalAddMBK_ToMatrixVisitor : public Mechani
...
+        return !map->f_mapMatrices.getValue();
```
guparan: Should be replaced by `return !map->areMatricesMapped();` ?

guparan:
ping @olivier-goury :-)

olivier-goury:
Hello @guparan, sorry for the delay! 
Yes that is a fix: this mechanical visitor is building the matrices M, B and K of the mechanical system.  If there is a mapping with the value mapMatrices set to 0, then this means the mapping does not wish to map those matrices. This implies the visitor should stop when it encounters such a mapping.

olivier-goury:
Without adding this stop, the default behaviour (define in the parent ) checks if forces are mapped, which is irrelevant for that visitor.

guparan:
[ci-build] [with-scene-tests]

PR #441, opened 2017-10-04
RigidMapping: fixed setRepartition backward compatibility

### LINKED ISSUE #435: SofaRigid/RigidMapping.inl: possible bug
While trying to clear the build from compiler-warnings I found these pieces of code in in the metioned files:

```C++
unsigned int idx = 0;
for( size_t i=0 ; i<size ; )
         for( size_t j=0; j<value && i<size ; ++j, ++i )
            rigidIndexPerPoint[i] = idx;
         ++idx;
```
(code is at line 280 in the file)
The compiler warning is indicating that the `++idx;` statement is not effected by the outer or inner for-loop. As in this case `rigidIndexPerPoint[i] = idx;` is always assigning "0" to the `rigidIndexPerPoint[i]` I would purpose either to change change the mentioned statement to `rigidIndexPerPoint[i] = idx++;` or make `idx` a constant variable.

matthieu-nesme:
fixes #435 RigidMapping

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaRigid/RigidMapping.inl
+++ b/SofaKernel/modules/SofaRigid/RigidMapping.inl
@@ -268,7 +268,7 @@ void RigidMapping<TIn, TOut>::clear(int reserve)
 template <class TIn, class TOut>
 void RigidMapping<TIn, TOut>::setRepartition(unsigned int value)
 {
-    serr<<helper::logging::Message::Deprecated<<"setRepartition function. Fill rigidIndexPerPoint instead."<<sendl;
+    msg_deprecated()<<"setRepartition function. Fill rigidIndexPerPoint instead.";
 
     helper::vector<unsigned int>& rigidIndexPerPoint = *this->rigidIndexPerPoint.beginWriteOnly();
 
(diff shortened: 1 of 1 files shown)


SofaKernel/modules/SofaRigid/RigidMapping.inl
```diff
@@ -268,7 +268,7 @@ void RigidMapping<TIn, TOut>::clear(int reserve)
 template <class TIn, class TOut>
 void RigidMapping<TIn, TOut>::setRepartition(unsigned int value)
 {
-    serr<<"setRepartition: deprecated function"<<sendl;
+    serr<<helper::logging::Message::Deprecated<<"setRepartition function. Fill rigidIndexPerPoint instead."<<sendl;
```
hugtalbot: Should it not use the "msg_deprecated" API ?

hugtalbot: (thank you Matthieu by the way for the PR!)

hugtalbot (approved):

PR #444, opened 2017-10-04
[SofaKernel] FIX don't inline exported functions

gKevin:
We had some errors while compiling sofa 17.06 on macos with Apple LLVM 9.0.
This was caused by two oddly inlined functions.

So here's a PR removing the inline declaration.
We hope it'll be useful.

Cheers.
______________________________________________________

This PR: 
- [ ] builds with SUCCESS for all platforms on the CI.
- [ ] does not generate new warnings.
- [ ] does not generate new unit test failures.
- [ ] does not generate new scene test failures.
- [ ] does not break API compatibility.
- [ ] is more than 1 week old (or has fast-merge label).

**Reviewers will merge only if all these checks are true.**

PR #449, opened 2017-10-06
[SofaKernel] FIX don't inline exported functions

gKevin:
This PR is basically the same as #444 but this time it is against master.

This PR: 


guparan:
Thanks for re-creating your PR.
[ci-build]

PR #458, opened 2017-10-09
[SofaPython] FIX crash in python script when visualizing advanced timer output

vmagno:
That script is used to parse the output of the AdvancedTimer in a Python scene to plot it. One of the function bindings has changed name, but this change was not reflected in the script.

This PR: 


guparan:
Thank for the fix @vmagno. I directly tag it as ready.

PR #461, opened 2017-10-10
[libQGLViewer] FIX missing headers

gKevin:
We ran into some more compile errors on windows using MSVC 2015.

libQGLViewer wouldn't compile without those missing headers.

Here's a quick fix but you might want to update libQGLViewer to the latest version (2.7.0).

This PR: 


guparan:
Thanks @gKevin. I set this PR as fast-merge.
[ci-build]

PR #469, opened 2017-10-12
[SofaBaseLinearSolver] Add comments in the CGLinearSolver

hugtalbot:
Based on the PR https://github.com/sofa-framework/sofa/pull/365, this work add comments 
Agreed that the default value of warmStart option should be false (see Maxime's comment in PR#335).

However, the associated comments in the code would be very useful for beginners starting and understanding the conjugate gradient in SOFA.

This PR: 


SofaKernel/modules/SofaBaseLinearSolver/CGLinearSolver.inl
```diff
@@ -175,32 +184,39 @@ void CGLinearSolver<TMatrix,TVector>::solve(Matrix& M, Vector& x, Vector& b)
...
 
+
         if( verbose )
         {
             sout<<"p : "<<p<<sendl;
         }
 
-        // matrix-vector product
+        /// Compute the matrix-vector product : M p
         q = M*p;
```
vmagno, line 205: The comments indeed help understanding the solver. I had a bit of trouble with that line when I realized it also triggers the visitor(s) that will gather the necessary data for that particular multiplication. It seems a bit much for a simple "=" and hides a lot of what goes on in Sofa. (This comment is not to fix it here, just to maybe start a discussion about it)

guparan (approved):

PR #470, opened 2017-10-12
[Build] FIX Sofa installation failure (tries to install non-existing files)

vmagno:
A directory that was recently removed is still mentioned in the cmake build files, so installation fails (compilation is fine).

This PR: 


jnbrunet (approved):

jnbrunet:
The files were removed by commit https://github.com/sofa-framework/sofa/commit/d48101075620cf6f37000f13ef5279e7a930d926

PR #472, opened 2017-10-13
[SofaKernel] Color deserialize

thomas-lemaire:
serialisation work on color datatype

This PR: 


thomas-lemaire:
yes, @hugtalbot you are right, but I thought the #332 was already solved indeed...

an other "color topic" I undergo currently is to store a RGBAColor on 4 bytes (unsigned char) instead of 4 float - let me known about
- your interest in such a feature (do I push a PR ?)
- if you plan to merge the previous PR which are required for this PR

PR #487, opened 2017-10-20
[SofaHelper SofaGraphComponent] issofa_plugin: Plugin management 

fjourdes:
I don't believe there was a pull request already attached to issofa_plugins branch ? My apologies if that 
was the case. 

# Purpose
Some basic plugin functionalities were silently changed for no good reasons other the past years, namely
- SOFA_LIBSUFFIX  which should be set as "_d" during a debug build is broken under msvc since c89da68 which as the commit name says is indeed monstruous. It was "fixed" later by adding directly inside the PluginManager code a condition to add the "_d" suffix by checking the compiler. 
- The default behavior of the global `PluginRepository`, was changed by c5a19fe for absolutely no good reasons.
It breaks a very basic functionaliy ( you can always load a library that is in your working directory ), also this change had some **interesting** side effects when basically you write a unit test which relies on
a scene file that uses a `RequiredPlugin`... that shortcoming was somewhat hidden by having all tests relying on `SofaGTest`, whose sole reason of being is to actually amend c5a19fe by populating the PluginRepository with the current working dir.

Some additional functionalities are added to `RequiredPlugin` by this PR, namely the ability to specify other suffixes ( or no suffix at all ) to prepend when attemtping to load a plugin with a name hint. 

# CHANGELOG 
FIX
[SofaHelper]
- Restore the `SOFA_LIBSUFFIX` behavior with msvc
- `PluginRepository` is initialized to contain the current working directory

CHANGE
[SofaHelper] 
- `PluginManager` : keep only a single version of the `loadPlugin` method. Also the method returns true
even if the plugin is already loaded. Returning false is kept only for real failures, whereby the library could 
not be loaded. If it is important to check wether or not a plugin is already loaded, the result of the `pluginIsLoaded` method can be inspected beforehand. 

[SofaGraphComponent] 
- `RequiredPlugin` component gives the ability to override the default suffix which is "_d" in debug mode.  
It also gives some more advanced options to extend the plugin search.

# Remarks

I do not know which version of the GUI is still in use. However I believe that the widget that can add/remove plugins to the `PluginManager` should be removed since there is no valid reason to 
keep it.

I am really not sure about the current state of SofaGTestMain. It is hidden somewhere in the folder hierarchy (inside tools directory), and its existence seems to imply that you cannot write a unit test inside sofa without relying on it to "initialise Sofa" which is absolutely not the case. See PR #471
Instead having a library which depends on gtest and SofaCore which 
- factorize some initialisation methods when they are actually required in the test `SetUp()` method
- provide some useful abstractions when you want to test some important concepts of a component, like what SofaTest wanted to do, but without the dependency bloat. A potential non exhaustive list of abstactions such a library could provide 
  - Mapping 
     - linearisation, 
     - applyJ applyJT equivalence 
  - ForceField
    - linearisation
    - addDForce addKToMatrix equivalence 
    - addKToMatrix must give a SPD matrix (well actually in Sofa convention I think internal forcefields actually give a SND matrix but the idea remains...

Nothing of this testing API would require the initialisation of a simulation / node.

   
This PR: 


fredroy:
I totally agree on the fact that the PluginManager widget in the Qt GUI is irrelevant now (useless to load a plugin manually AFTER loading the scene using element from it).

SofaKernel/framework/sofa/helper/system/PluginManager.cpp
```diff
@@ -289,12 +296,14 @@ std::string PluginManager::findPlugin(const std::string& pluginName, bool ignore
             {
                 const std::string& filename = *j;
                 const std::string downcaseFilename = Utils::downcaseString(filename);
-                if (downcaseFilename == downcaseLibName)
+                if (downcaseFilename == downcaseLibName) {
                     return dir + "/" + filename;
+                    std::cout<<"dir : "<<dir + "/" + filename<<std::endl;
```
fredroy: message API (msg_error/warning/etc) should be used here

SofaKernel/framework/sofa/helper/system/PluginManager.h
```diff
@@ -136,15 +136,41 @@ class SOFA_HELPER_API PluginManager
...
+	/// Loads a plugin library in process memory. 
+	/// @param pluginName The filename without extension of the plugin to load
+	/// @param suffix An optional suffix to apply to the filename. Defaults to "_d" with debug builds, empty otherwise.
+	/// @param ignoreCase Specify if the plugin search should be case insensitive (activated by default). 
+	///                   Not used if the plugin string passed as a parameter is a full path
+	/// @param errlog An optional stream for error logging.
+	bool loadPluginByName(const std::string& pluginName, const std::string& suffix = getDefaultSuffix(), bool ignoreCase = true, std::ostream* errlog= nullptr);
+    
+	/// Unloads a plugin from process memory.
+	bool unloadPlugin(const std::string& path, std::ostream* errlog= nullptr);
```
fredroy: Nice comments !

modules/SofaGraphComponent/RequiredPlugin.cpp
```diff
@@ -42,48 +42,93 @@ int RequiredPluginClass = core::RegisterObject("Load the required plugins")
...
+        if ((d_requireAll.getValue() || (d_requireOne.getValue() && loaded.empty())))
+        {
+            serr << errmsg.str();
+            serr << "Required plugin"<<(failed.size()>1?"s":"")<<" failed to load: " << failed << sendl;
+        }
+        else
+        {
+            sout << errmsg.str();
+            sout << "Optional/alternate plugin"<<(failed.size()>1?"s":"")<<" failed to load: " << failed << sendl;
+        }
```
fredroy: same as before, about the message API

fredroy (changes requested):

SofaKernel/framework/sofa/helper/system/PluginManager.cpp
fjourdes: This likely result from the merging. The std::cout statement will in fact never been called, since it is after a return statement. The line should be removed entirely.

fredroy (approved):

modules/SofaGraphComponent/RequiredPlugin.cpp
fjourdes: This one comes from our local version. We do not use the message API at InSimo

fredroy:
[ci-build]

PR #500, opened 2017-10-26
[SofaValidation] move code to set default folder for monitor to init function

sergeiNikolaev:
This fix allows to set the default filename for the Monitor class. The default name assignment is removed in 'init()' function since in constructor it doesn't work. 

This PR: 


guparan:
[ci-build]

PR #510, opened 2017-11-06
[CGALPlugin] ADD compatible CGAL versions in Readme

guparan:
Very basic PR just to let you know that the CGAL plugin is not compatible with latest CGAL.
As is, it compiles with CGAL <= 4.9.1

CI build is unnecessary for this PR.

This PR: 

### DIFF
--- a/applications/plugins/CGALPlugin/CGALPlugin.txt
+++ b/applications/plugins/CGALPlugin/CGALPlugin.txt
@@ -17,3 +17,4 @@ LICENCE :
 
 OTHER COMMENTS :
 Use CGAL components in SOFA.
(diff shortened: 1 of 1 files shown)


bcarrez (commented):
I believe there is no better way to deal with this for now... I looked into fixing compilation with CGAL 4.11 and it is not trivial as I was hoping.

guparan:
Thanks @bcarrez for taking the time to check :-)
@vmagno didn't you fix the plugin to work with latest CGAL? That's what @jnbrunet told us.

guparan:
Replaced by #515

PR #519, opened 2017-11-17
[SofaPython] FIX error in script for plotting advancedTimer output

vmagno:
The script would sometimes plot completely wrong info because some numbers were treated as strings.

This PR: 


guparan:
LGTM, I set this to ready ;-)

PR #520, opened 2017-11-17
[SofaKernel] FIX Remove override warnings

vmagno:
This removes some override warnings that were (re)introduced recently.

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/helper/io/ImageDDS.h
+++ b/SofaKernel/framework/sofa/helper/io/ImageDDS.h
@@ -49,8 +49,8 @@ public:
         load(filename);
     }
 
-    bool load(const std::string &filename);
-    bool save(const std::string &filename, int compression_level = -1);
+    bool load(std::string filename) override;
+    bool save(std::string filename, int compression_level = -1) override;
 };
 
 } // namespace io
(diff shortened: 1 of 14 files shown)


SofaKernel/framework/sofa/helper/io/ImageDDS.h
```diff
@@ -49,8 +49,8 @@ class SOFA_HELPER_API ImageDDS : public Image
         load(filename);
     }
 
-    bool load(const std::string &filename);
-    bool save(const std::string &filename, int compression_level = -1);
+    bool load(std::string filename) override;
+    bool save(std::string filename, int compression_level = -1) override;
```
guparan, line 53: Do you remove the `const &` intentionally here?

vmagno: I did. It's to make it consistent with the function signature from the base class, otherwise we get a warning with clang (I also removed the &).

It might be a better idea to modify the one from the base class. However, in the implementation of that one, the filename is passed to a function that modifies it, so we would need to look into that and change the code there. I think that would be the better option, we should not need to modify the name of a file we want to open...

PR #555, opened 2018-01-04
[CI] EDIT: make scene-tests output file smaller

guparan:
by filtering duplicates warnings/errors/crashes

This should fix JVM heap problems on small VMs when parsing scene-testing/warnings.txt, scene-testing/errors.txt and scene-testing/crashes.txt

Please don't forget the real problem is that we have 60k+ scene-test warnings...

This PR: 

### DIFF
--- a/scripts/ci/scene-tests.sh
+++ b/scripts/ci/scene-tests.sh
@@ -321,7 +321,9 @@ extract-warnings() {
             sed -ne "/^\[WARNING\] [^]]*/s:\([^]]*\):$scene\: \1:p \
                 " "$output_dir/$scene/output.txt"
         fi
-    done < "$output_dir/all-tested-scenes.txt" > "$output_dir/warnings.txt"
+    done < "$output_dir/all-tested-scenes.txt" > "$output_dir/warnings.tmp"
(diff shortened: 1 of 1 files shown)


guparan:
Please someone merge this :-)

epernod:
no need to wait dashboard?

damienmarchal:
Interesting. Isn't it the fault of the scene checker activated by default ? 
It will report a huge amount of things on old scenes that was not reported as error before. Maybe we should have an option to disable it (or when in batch mode). 
  
EDIT: about the failure....it is because this is compiled on the new _pr3 build VM. Where, I assume, some library is missing.

epernod:
Is it error/warning like this you are speaking about? @damienmarchal 

sans titre

damienmarchal:
@epernod is it a question to me or Guillaume ?

guparan:
This PR won't solve any error nor warning. It will just prevent the warning parser (on Jenkins) from crashing.
About the SceneChecker raising lots of warnings: what kind of warnings does it raise? Is it normal that it is triggered for our CI batch tests?

PR #572, opened 2018-01-23
Sofa python without executable

jnbrunet:
Hey guys,

Sorry for the delay. This is a draft PR so that people can start reviewing/testing this feature if they want. There is still some things missing before merging.

### Needs to be addressed before merging

- MacOS and Windows compatibility: python import command only tries to load .so extension, so the cmake of the plugin should generate both SofaPython.dylib (.dll) and SofaPython.so for it to work.
- Handling the case of python script loading a python scene (the current GIL handling implementation will prevent the main python script to get back the interpreter after a python scene runs)

### What this PR brings

Right now, in order to use the SofaPython plugin, you need to create a python scene and run it through a C++ executable (such as runSofa).

What I'm proposing here is to simply load the Sofa plugin inside a python script, that's it. A python script could then look like : 

```#!/usr/bin/python2.7
from libSofaPython import Sofa
import argparse

# Parse the arguments
parser = argparse.ArgumentParser()
parser.add_argument("-n", "--number_of_steps",   type=int,   help="Number of steps until the final forces are reach", default=10)
o = parser.parse_args()

sim = Sofa.createSimulation("DAG")
Sofa.setSimulation(sim)
root = Sofa.createNode("root")

# ...
# some code that creates the scene tree
# ...

root.init()

for i in range(number_of_steps):
   root.simulationStep(dt)
   updateConvergencePlot(root)

analysis = complexCalculsOnMatrices(root)
saveResults(root, analysis)
```

You could even make your own runSofa.py that launches the runSofa GUI (a more complete python runSofa script can be found in an example file provided by this PR) :

```#!/usr/bin/python2.7

from libSofaPython import Sofa

sim = Sofa.createSimulation("DAG")
Sofa.setSimulation(sim)

root = Sofa.loadScene("examples/Demos/caduceus.scn")
root.init()

sofa_prefix = "/home/jnbrunet/sources/sofa/build"
Sofa.GUIManager.setSofaPrefix(sofa_prefix)
Sofa.GUIManager.setConfigDirectoryPath(sofa_prefix + "/config")
Sofa.GUIManager.setScreenshotDirectoryPath(sofa_prefix + "/screenshots")

Sofa.GUIManager.init("qglviewer")
Sofa.GUIManager.createGUI()
Sofa.GUIManager.setDimension(800, 600)
Sofa.GUIManager.setScene(root)

Sofa.GUIManager.MainLoop(root)

Sofa.GUIManager.closeGUI()
```
### What else this PR brings

**Plugin manager**
One of the problem arising from this feature is the duplicated loading of the python plugin: when issuing the `import Sofa` command in the python script the interpreter loads the library (SofaPython.so). If the script explicitly calls the `Sofa.loadPlugin("SofaPython")` command, Sofa's PluginManager will load again the same library. Example of a faulty script:
```#!/usr/bin/python2.7
from libSofaPython import Sofa # Loads SofaPython.so in the memory
Sofa.loadPlugin("SofaPython")  # Loads again SofaPython.so in the memory
```
I think this brings up a wider question:

_What should happen if the same plugin/library is loaded twice or more in the memory?_

Now, most operating systems will prevent loading a library twice in the memory. On OS supporting the dlopen command, and on Windows loadLibrary command, the library will be loaded only once, but a counter is incremented each time we try to load it. This counter is decremented when we unload the plugin, and the library will be completely cleaned from the memory ONLY when this counter reach 0.

On the other hand, Sofa's plugin manager WILL call the `initExternalModule` function implemented on the library each time we load it. This can produce weird behaviors as most plugin won't handled the multiple initialization calls.

_How to reproduce the problem?_

In most cases, people will use the pluginManager to load the plugin. Using this twice or more with the same plugin path will already be handled by the pluginManager which looks into its loaded plugins map to make sure we don't load a same library file more than once (using its absolute path).

In other cases, a library could already be loaded into the memory by another procedure external to Sofa (such as the "import" command from python interpreter, a self made dlopen command, a dynamic linking from the compiler, etc.). Using the pluginManager will than failed to prevent the multiple links since its map doesn't contain the plugin yet (since the manager was not used to load it in the first place).

To sum up:

Case 1
```
 > PluginManager::loadPlugin("/path/to/myplugin.so")
 Plugin loaded and added to the PluginManager plugins map
 > PluginManager::loadPlugin("/path/to/myplugin.so")
 Plugin won't be loaded and a warning is printed to the user telling him that the plugin is already loaded
```
Case 2
```
 > dlopen("/path/to/myplugin.so")
 Plugin loaded but not added to the PluginManager
 > PluginManager::loadPlugin("/path/to/myplugin.so")
 Plugin will be loaded a second time by the PluginManager, this should be prevented
```

Case 3
```
 > PluginManager::loadPlugin("/path/to/myplugin.so")
 Plugin loaded and added to the PluginManager plugins map
 > PluginManager::loadPlugin("/DIFFERENT/path/to/myplugin.so")
 Plugin will be loaded a second time by the PluginManager, this should be prevented
```

_How did I fixed this_
This is one way of doing it. If you have other ideas/suggestions, please tell me.
Since in most of the plugins we have, we define a `getModuleName` and `getModuleVersion`, we should use them to compared loaded plugin instead of plugin paths. Also, we should be able to notify the PluginManager of a plugin that we loaded without him noticing (such as calling dlopen). Here is what I did:

- `PluginManager::loadPlugin` now iterates through its list of loaded plugins and check that the the plugin we want to load does not contains the same `initExternalModule` function's pointer or plugin's name as another already loaded plugin
- A plublic `PluginManager::addPlugin` that adds a plugin to the register without loading it (it must be loaded in the memory beforehand) is now available.

**Supplementary python bindings**
Since me and @vmagno used this PR extensively during the past 2 months for our simulations, we added a couple of python bindings useful to us.

I think this wraps up a lot of things, let me know what you guys think.

This PR: 


jnbrunet:
I'm closing this since the new awesome [SofaPython3 plugin](https://github.com/SofaDefrost/plugin.SofaPython3 ) does everything this PR was doing, but better.

PR #573, opened 2018-01-23
[SofaGraphComponent] FIX Fix suggested by SceneChecker contains an error

vmagno:
The scene checker detects when we use a plugin but forgot the "required plugin" tag and tells us how to add it in our scene, but without closing the tag. This PR fixes the bug in the suggestion and corrects some typos in error messages in the same file.

This PR: 


guparan:
No need to wait the CI for this one as it contains only output corrections. Thanks @vmagno :-)

PR #598, opened 2018-02-22
[SofaGUI] Move GlutGUI to projects and remove all glut references in SofaFramework

fredroy:
Glut is a deprecated library and in my opinion, should be not used anymore. (not compatible with newer version of OpenGL, etc)
It can be replaced with GLFW to create quick graphical applications.

This first PR is moving the SofaGuiGlut to the applications directory, in case any people are still using it.
Actually, the code was not compatible with the current SofaGui API and was not compiling.
The code is now compilable and launchable anyway.

Any reference to Glut are removed from the SofaFramework (header, CMake) and from runSofa.

Lastly, a replacement using gflw will be pushed in the near future, for people wishing to have a simple GUI without the heavy-duty Qt Framework.

This PR: 


applications/projects/CMakeLists.txt
```diff
@@ -16,7 +16,16 @@ endif()
 
 sofa_add_application(GenerateRigid GenerateRigid)
 sofa_add_application(meshconv meshconv OFF)
-sofa_add_application(SofaPhysicsAPI SofaPhysicsAPI)
+
+## Glut GUI
+find_package(GLUT)
+if(GLUT_FOUND)
+	sofa_add_application(SofaPhysicsAPI SofaPhysicsAPI)
```
guparan: Would be better to add every app disabled and let them fail when running `find_package(GLUT REQUIRED)` if enabled, no?  
I don't think this CMakeLists should be smartly checking `GLUT_FOUND`. It's not its dependency.

fredroy: I was wondering the same, but I was lazy to add the check in all the projects needing glut.
But I guess you are right so I will add it in those projects directly

PR #610, opened 2018-03-14
[runSofa] Add functionality of command for infinite iterations solving issue #552

m4n4n:
This solves the issue : [runsofa] -a -n conflict for batch #552
In case to run infinite iterations, the command to be run in command line is: 
 bin/runSofa -g batch -n -1

This PR: 

### DIFF
--- a/applications/projects/runSofa/Main.cpp
+++ b/applications/projects/runSofa/Main.cpp
@@ -215,7 +215,7 @@ int main(int argc, char** argv)
     .option(&gui,'g',"gui",gui_help.c_str())
     .option(&plugins,'l',"load","load given plugins")
     .option(&nbMSSASamples, 'm', "msaa", "number of samples for MSAA (Multi Sampling Anti Aliasing ; value < 2 means disabled")
-    .option(&nbIterations,'n',"nb_iterations","(only batch) Number of iterations of the simulation")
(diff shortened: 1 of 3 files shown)


tgaugry:
Any reason to do this instead of adding a flag `--infinite-run` or similar ? `-1` is a bit counter-intuitive.

m4n4n:
I wanted to pass on a flag after -n in a format such as -n infinite.
However, the general syntax constituted of a number after -n, such as: -g batch -n 20 (for 20 iterations).
Therefore the variable storing it was declared as an integer one and a lot of compilation errors came up when I tried to change to string variable and later on validate it for a digit or string.

tgaugry:
I kinda see what you went through. The argument parser isn't that intuitive TBO.

I recently had to do a similar task, so if you still want the "infinite" flag, you can check this [line](https://github.com/tgaugry/sofa/commit/326f4f4c8fea61783e40225cb4abac648e569922#diff-5904f9a1380b53f6de2fdc644b6b1a48R112)  and the `parseSkipOption` function in the same file. 
Maybe this would solve your problem ?

guparan:
Hi @m4n4n 
Did you have the time to rework this branch to base it on SOFA master as we discussed on Gitter?

m4n4n:
I was busy studying about Pulse this weekend. Working on it currently. Will come up with a PR in the next couple of days.

guparan:
See #613 for updated version.

PR #611, opened 2018-03-14
[CI] Ignore some scenes in scene tests

guparan:
Now we have a tool to list the deprecated components, we can build with `SOFA_WITH_DEPRECATED_COMPONENTS=OFF` + `APPLICATION_GETDEPRECATEDCOMPONENTS=ON` and ignore scenes using deprecated components during scene tests.

I also added a function to ignore scenes using non-existent (because disabled) plugins.

This PR: 


guparan:
[ci-build][with-scene-tests]

PR #614, opened 2018-03-21
[SofaPython] FIX build for MacOS >10.13.0

hugtalbot:
Issue detected by @quentinfrancois0

This comes from a change in the Python configuration since the last High Sierra.
The line #include <Python/Python.h> involved a link "Python" to the python include, but this is no longer valid in High Sierra. This was causing a fail in the compilation (include not found)

This PR: 

### DIFF
--- a/applications/plugins/SofaPython/PythonCommon.h
+++ b/applications/plugins/SofaPython/PythonCommon.h
@@ -23,11 +23,9 @@
 
 #endif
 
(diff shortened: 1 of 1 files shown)


bcarrez (commented):
This fix works, if you have MacOS > 10.13 (I did it myself on my own machine, indeed); but I'm not sure it will still work with OSX < 10.12 ? We should check this before the merge.

PR #617, opened 2018-03-27
[SofaHelper] More robust method to test end of string

alxbilger:
The case where plugin is smaller than dotExt is not handled by std::equal. See how it works: http://en.cppreference.com/w/cpp/algorithm/equal. An assert is called.
I had the case when dotExt = ".dll" and plugin = "PSL".

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/helper/system/PluginManager.cpp
+++ b/SofaKernel/framework/sofa/helper/system/PluginManager.cpp
@@ -204,7 +204,7 @@ bool PluginManager::loadPlugin(const std::string& plugin, const std::string& suf
 {
     // If 'plugin' ends with ".so", ".dll" or ".dylib", this is a path
     const std::string dotExt = "." + DynamicLibrary::extension;
-    if (std::equal(dotExt.rbegin(), dotExt.rend(), plugin.rbegin()))
+    if (plugin.size() > dotExt.size() && std::equal(dotExt.rbegin(), dotExt.rend(), plugin.rbegin()))
     {
         return loadPluginByPath(plugin,  errlog);
     }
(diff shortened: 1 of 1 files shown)


guparan:
Hey @alxbilger, thank you for your PR!
Does this mean that PSL.dll will not be loaded?
[ci-build]

alxbilger:
In debug, an exception is thrown and the program crashes. We can say that PSL.dll is not loaded. Try this code in debug:

```c++
#include <iostream>
#include <algorithm>

int main(int argc, char *argv[]) 
{
    std::string plugin("PSL");
    std::string dotExt(".dll");

    const bool b = std::equal(dotExt.rbegin(), dotExt.rend(), plugin.rbegin());

    if (b) std::cout << "equal\n";
    else std::cout << "not equal\n";
}
```

guparan:
But we want PSL.dll to be loaded as a path and not as a name.
I think we should replace the whole test by a simple `if (plugin.find(dotExt) != std::string::npos)`

alxbilger:
Agreed. It's simpler.
To be clearer: plugin was equal to "PSL" and not "PSL.dll". In this case, we want to to load as a name, but the test based on std::equal crashes (in debug only).

guparan:
Ok I didn't understand that PSL is loaded as a name in your use case sorry ^^

alxbilger:
@guparan Your proposition does not work in the unlikely case of a plugin name containing ".dll" in its name. The initial contract was, from the comments, to find the extension in the end of the string.

guparan:
You are right I answered too fast. Actually your fix is exactly what is needed.
Could you propagate it to other `if (std::equal(dotExt.rbegin(), dotExt.rend(), plugin.rbegin()))` calls (there are others)?

guparan:
Great thank you :+1: 
[ci-build][with-scene-tests]

guparan (approved):

PR #621, opened 2018-04-03
[SofaHelper] PluginManager now checks for file existence instead of library extension match.

jnbrunet:
This PR adds a `isFile` function to the FileSystem component. The PluginManager will then use this new function to test whether or not the plugin path is an existent file.

@guparan Concerning your question last week (what happens if the name of the plugin is passed without a path), the previous behavior was to check if the file was a valid library file (.so, etc.), and then gather the full path from the search directories. The new behavior will first check that the file is a valid file (which won't be if only the name is passed), and then try to find it in the search directories. This should not break any old behaviors.

This PR: 

### DIFF
--- a/SofaKernel/framework/framework_test/helper/system/FileSystem_test.cpp
+++ b/SofaKernel/framework/framework_test/helper/system/FileSystem_test.cpp
@@ -24,6 +24,7 @@
 #include <gtest/gtest.h>
 #include <exception>
 #include <algorithm>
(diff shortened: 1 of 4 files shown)


guparan (approved):

guparan:
As suspected, the two missing unit tests were not yours. Two plugins were ignored in PR builds on Windows, implying 2 unit tests.
It's fixed :+1:

PR #626, opened 2018-04-09
[MultiThreading] FIX: add createSubelements param in MeshGmshLoader

guparan:
This fix was already applied to examples/Components/forcefield/TriangularForceFieldComparison.scn
See 28c8b68f9d6ed4be725baf305a0957097581fb13

Sorry but I can't explain why this fix was needed in the first place.  
If someone could, that would be great :+1: 

This PR: 


guparan:
[ci-build][with-scene-tests]

guparan:
[ci-build][with-scene-tests]

epernod (approved):

epernod:
createSubelements="true": Divide all n-D elements into their (n-1)-D boundary elements (e.g. tetrahedra to triangles). 
Meaning: for a triangulation, create all edges.

In this case the mesh file had only edges of the border but the FF were using all edges of the mesh.

epernod:
This PR fix the crash.

But some work is needed on the meshloader to avoid this behavior.

PR #627, opened 2018-04-09
 [SofaKernel] FIX Bug in BoxROI that is not properly initialized

VannesteFelix:
This PR fix an issue with the BoxROI component.

There was a test in the constructor of BoxROI that created an unwanted behavior.
This test was moved in reinit()

This PR: 


guparan:
[ci-build][with-scene-tests]

VannesteFelix:
the unwanted behavior is describe in the commit of @damienmarchal : 
"
[SofaKernel] FIX Bug in BoxROI that is not properly initialized.

The code is automatically adding an alignedbox if nothing is provided in
the constructor. This kind of test make no sense in the constructor as
the "isSet" is sure to be set to false. The test should go in init or reinit.

This commit fix that.
"

PR #628, opened 2018-04-09
[examples] FIX missing RequiredPlugin

guparan:
One good point about ignoring scenes with unmet `RequiredPlugin` dependencies is that we can see what scenes need some `RequiredPlugin`: they are the only left to crash trying to create unknown components.

This PR: 


guparan:
[ci-build][with-scene-tests]

PR #630, opened 2018-04-10
[All] FIX VS2017 build (Windows)

guparan:
This PR suggested by @alxbilger fixes build on Windows with VS2017.

Error was `C2057: expected constant expression`

This PR: 


epernod (commented):
I pushed a different fix:  explicit template instantiations before the templated class is used.
This seems to do the trick and avoid to redefine new vector template definition.

epernod:
[ci-build][with-scene-tests]

PR #638, opened 2018-04-16
[SofaKernel] Header include cleanup

epernod:
Just removing unnecessary headers include inside header in sofa::core
No forward class or intelligent refactoring.

any suggestion/remark is welcomed. 

This PR: 


epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

damienmarchal:
[ci-build][with-scene-tests]

modules/SofaGeneralLoader/SphereLoader.h
```diff
@@ -21,11 +21,8 @@
 ******************************************************************************/
 #ifndef SOFA_COMPONENT_LOADER_SPHERELOADER_H
 #define SOFA_COMPONENT_LOADER_SPHERELOADER_H
-#include "config.h"
```
damienmarchal: I'm not sure about removing this one....because no one will define SOFA_TARGET anymore. 

While it is supposed to be used in the ObjectFactory:
```cpp
/// The name of the library or executable containing the binary code for this component
    virtual const char* getTarget()
    {
#ifdef SOFA_TARGET
        return sofa_tostring(SOFA_TARGET);
#else
        return "";
#endif
    }
```


EDIT: Actually all the component in SofaGeneralLoader should have somewhere the target set to SOFA_TARGET "SofaGeneralLoader" (which I'm sure is what is happening). 

NB: I'm really not found of silently ignoring the undefined SOFA_TARGET.

modules/SofaGeneralLoader/GeneralLoader.h
```diff
@@ -1,33 +0,0 @@
-/******************************************************************************
```
damienmarchal: Hello Erik, 

I'm not sure this one should be removed, on the contrary its use should be generalized. Additionally I think it should also define the SOFA_TARGET to SofaGeneralLoader.

Please confirm @guparan

modules/SofaGeneralLoader/GeneralLoader.cpp
```diff
@@ -1,35 +0,0 @@
-/******************************************************************************
```
damienmarchal: Why did you removed also code ? 

From PR description I would have expected only changing lines with #include<XXX>

modules/SofaGeneralLoader/GeneralLoader.h
epernod: Ok I sent a message to @guparan already regarding this.
Right now all macros are defined in sofa-build\include\SofaMisc\config.h etc..
This class is not included. Let me know which one should be kept. I'll revert the commit if needed.

modules/SofaGeneralLoader/SphereLoader.h
epernod: ok I can revert that. I think it has been removed because this class include already another loader header that include config.h
If the logic is: every class include it's config.h, I'm ok with it.

modules/SofaGeneralLoader/GeneralLoader.cpp
epernod: yes, I saw all includes could be removed because this class is not anymore in the cmakeLists.
Another PR regarding deprecated files will be better indeed because I found other similar cases.

modules/SofaGeneralLoader/SphereLoader.h
damienmarchal: Maybe we should replace this code by:
```cpp
#ifndef SOFA_TARGET
        #error....the SOFA_TARGET is not defined. 
#else: 
        return sofa_tostring(SOFA_TARGET);
#endif 
```
(Not in this PR of course)

epernod:
[ci-build][with-scene-tests]

modules/SofaGeneralLoader/SphereLoader.h
epernod: I will put those headers back and we will see later where those target should be set.

modules/SofaGeneralLoader/GeneralLoader.h
epernod: I will report our discussion in an issue and remove that change from the PR to have only changes on header includes.

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

damienmarchal:
READY and recently compiled over the issofa PR ...I merge it.

SofaKernel/framework/sofa/core/behavior/BaseMechanicalState.h
```diff
@@ -22,17 +22,14 @@
 #ifndef SOFA_CORE_BEHAVIOR_BASEMECHANICALSTATE_H
 #define SOFA_CORE_BEHAVIOR_BASEMECHANICALSTATE_H
 
-#include <sofa/config/build_option_experimental_features.h>
```
damienmarchal, line 25: I merged without noticing this one.
It shouldnt be removed. 
Because this one add some macro that are tested with
#ifdef XXXXX 

So removing it "compile" but the resulting code is not working anymore. 
....

damienmarchal: A very bad things is that even in condition like #if MACRO == 1
When the macro is not defined then its value is == 0 so removing the include line is silent.
https://gcc.gnu.org/onlinedocs/cpp/If.html

A workaround I can propose it to do:
```cpp
#define MYTEST() 0 
....
#if MYTEST() 
        std::cout << "COUCOU A" << std::endl ;
#else
        std::cout << "COUCOU B" << std::endl ;
#endif 
```

So that we get error when MYTEST is not defined (the bad things is that the error is not very clear with some compilers):
with gcc:
toto.cpp:8:11: error: missing binary operator before token "(" 
 #if MYTEST()

with clang:
toto.cpp:8:5: error: function-like macro 'MYTEST' is not defined
#if MYTEST() 
    ^

PR #651, opened 2018-05-04
[CMake] Fix SHARE_DIR and EXAMPLES_DIR

marques-bruno:
This PR makes it easier & more reliable for plugins / external projects to find paths to the "examples/" and the "share/" folders in both the build directory and install directory.

The current system was filling a file (sofa.ini) for this single purpose, that GUIManager was then parsing to retrieve those paths. if another project (tests for instance) would want to retrieve those paths, they would have to first locate this sofa.ini file (which can be very tricky if not part of the SOFA project tree), parse it and add it to the DataRepository.

Now, preprocessor defines holds the paths and you just have to add it to the DataRepository as such:
sofa::helper::system::DataRepository.addFirstPath(SOFA_SHARE_DIR)
sofa::helper::system::DataRepository.addFirstPath(SOFA_EXAMPLES_DIR)

In both cases, the paths are *absolute* which is great (and what I needed) when building an external project, but generating distributable binaries with absolute paths will not work out well...
I realise this PR is not mergeable until this issue is resolved, but don't know how to change that.
With relative paths, finding the install directory from external projects is impossible, and with absolute paths, redistribution of binaries is not feasible. 

Looking fwd to your inputs!

### DIFF
--- a/SofaKernel/SofaFramework/CMakeLists.txt
+++ b/SofaKernel/SofaFramework/CMakeLists.txt
@@ -202,9 +202,9 @@ set(SHARE_DIR \"${CMAKE_SOURCE_DIR}/share\")
 set(EXAMPLES_DIR \"${CMAKE_SOURCE_DIR}/examples\")
 configure_file("config.h.in" "${CMAKE_BINARY_DIR}/include/sofa/config.h")
 
-## Paths to share/ and examples/, this one contains relative paths to the installed resources directory.
-set(SHARE_DIR \"../../share/sofa\")
-set(EXAMPLES_DIR \"../../share/sofa/examples\")
+## Paths to share/ and examples/, this one contains absolute paths to the installed resources directory.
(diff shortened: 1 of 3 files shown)


marques-bruno:
Another possibility would be to create 2 macros, SOFA_SHARE_DIR_ABS and SOFA_SHARE_DIR_REL and add them both to the DataRepository. Then no matter if one path is incorrect, it should be fine... kind of a hack, but that's more or less what's done with RPATHs when the linker searches for libraries, to my understanding..?
Again, looking forward to your opinion!

marques-bruno:
Hum... After thinking it through a little longer, I realize that this PR is absolutely non-functional in practice. When deploying the "installed" config.h, the preprocessor macro is indeed replaced with the correct path in the config.h file, but, the preprocessor / compiler is not ran again for the deployment step, of course.. which means that in the binaries, SHARE_DIR is still equal to the build step value. Hence the use of a .ini file... So I'm closing this dumb PR...

Issue #42, opened 2016-10-19
Duplicate code between MeshXX and MeshXXLoader


epernod:
Is there a real reason to have duplicate code between 
MeshSTL and  MeshSTLLoader

idem for
SOFA_LINK_CLASS(MeshOBJ)
SOFA_LINK_CLASS(MeshTrian)
SOFA_LINK_CLASS(MeshSTL)

What is the final decision?

Edit: BTW how do we add a label to github issue? I wanted to put this one under "discussion"

JeremieA:
`sofa::helper::io::Mesh*` has been deprecated since many years, replace by `MeshLoader` components, it would be indeed good now to clean it up ;)
The biggest chunk of work required is to remove all the `"filename"` Data still around in quite a few components (BaseMeshTopology, MechanicalObject, DiagonalMass, UniformMass, VisualModelImpl, SpringForceField, RigidMapping, RigidRigidMapping), which would be easy except that we need clear warnings if an XML is loaded still using this, as those components are used everywhere. A quick search in Sofa examples directory finds 500 instances of OglModel with a filename or fileMesh attribute...

epernod:
Up!
Can we remove the sofa::helper::io::Mesh* ?
And also do we go to a solution using external mesh tool, like Assimp?

JeremieA:
👍 for removing sofa::helper::io::Mesh*

There is however an issue that the support for loading texture coordinates was removed from MeshObjLoader, so the OglModel version on the master can only have textures when using these deprecated loaders. In our latest integration branch we reverted this removal of texcoords support and fixed it to have the same level of support using MeshObjLoader+OglModel as when loading the file directly in OglModel (most notably handling of "seams" where a single vertex has multiple texture coordinates). The relevant commits are:
- https://github.com/InSimo/sofa/commit/915ee46368efe140b242759a0110682f71200dc1
- https://github.com/InSimo/sofa/commit/08493b9ab5c8ea376ba9a7f9a2cea3e81f513511
- https://github.com/InSimo/sofa/commit/e6bec10562ab797963027003c2282cbc24a53bb9
- https://github.com/InSimo/sofa/commit/1925b381b7d8cb939d250ab24e92fb28df918c0b

To be merged to the master the first commit need to be done in a cleaner way than a full revert, which removed later refactorings like apparently the introduction of a File class.

epernod:
Ok thanks for the information. I will check that.

epernod:
[...]

PR #661, opened 2018-05-21
[SofaKernel][SofaLoader] Merge duplicated loader codes

epernod:
Start work on merging the duplicated code of the mesh loader components versus helper::io::mesh loaders. See issue #42 

Basically, all code to load a mesh file directly in the topology without using a meshLoader component. For example: `<GridTopology fileMesh="my_mesh.msh" /> or <OglModel filename="my_mesh.obj" />`
go through the class: helper::io::MeshTopologyLoader which had directly some code to load several formats or was using a helper::io::meshXX to read XX file format.

All this code is duplicated with the meshLoader components and with some differences (otherwise it would not be fun...)

For the moment I only updated the MeshTopologyLoader and the helper::io::mesh classes to check the tests on all plateforms. 
Changes are:
- obj -> nothing (use MeshObj class)
- trian -> nothing (use MeshTrian class)
- vtk -> remove code from 2009 and use MeshVTK class from 2015. Still different from meshVTKLoader.
- stl -> nothing (use MeshSTL class)
- gmsh -> create a MeshGmsh class and move the code in it.
- xsp -> create MeshXsp class and move the code in it.
- cgal -> remove the methods. Assume it is deprecated.
- unknown format -> the "else" method when format not found return false now and is set as deprecated.

- Update the helper::io::mesh class to have topology buffers in order to be similar as baseMeshLoader for future merge.

This PR: 


epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

PR #671, opened 2018-05-29
[SofaKernel] Improve Displayflags

hugtalbot:
This PR allows to add aliases in C++ in DisplayFlags.
Moreover, DisplayFlags are now case insensitive but a warning is given.
The DisplayFlagsDataWidget is now fixed to correspond to the FlagTreeItem names.

This PR: 


epernod:
[ci-build][with-scene-tests]

PR #680, opened 2018-06-06
[SofaConstraint] FIX: Moving semicolon under preprocessor define

marques-bruno:
Quick fix of compile error, quite self-explanatory I guess

This PR: 


tgaugry:
Seems pretty fine to me !
Do we know if there are other places this happens as well ?

guparan (approved):

PR #687, opened 2018-06-13
[applications] FIX Install files that were ignored by the `make install` command

vmagno:
This removes warnings that appear when running runSofa from a directory where it was installed from source. This should not affect compilation in any way (but not sure how it would affect packaging Sofa).

This PR: 


guparan:
Thanks @vmagno ! This works completes mine from #635. I will directly merge it in my PR and close this one ;)

guparan:
It's done, I let you close your PR if you agree :wink:

guparan:
ping @vmagno

PR #692, opened 2018-06-19
[CMake] use ccache if available

untereiner:
This PR makes so that the user can activate an option to look for and use CCACHE for compilation.
If the option is activated, then CMake looks for ccache. If ccache is found, it is used for compilation.

This PR: 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -236,7 +236,7 @@ if(SOFA_USE_CCACHE)
         set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK "${CCACHE_PROGRAM}")
     else()
         message(WARNING "ccache not found, disabling option")
-        set(SOFA_USE_CCACHE OFF)
+        set(SOFA_USE_CCACHE OFF CACHE bool FORCE)
     endif()
 endif()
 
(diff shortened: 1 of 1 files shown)


guparan:
Hi @untereiner, thanks for your PR. Actually, this feature should be contained in an option because we may not want to use ccache even if we have it installed.
Could you make that change?

CMakeLists.txt
```diff
@@ -225,6 +225,21 @@ if(SOFA_BUILD_TESTS)
...
+if(SOFA_USE_CCACHE)
+    find_program(CCACHE_PROGRAM ccache)
+    if(CCACHE_PROGRAM)
+        # Support Unix Makefiles and Ninja
+        #TODO replace by <LANG>_COMPILER_LAUNCHER when min cmake version > 3.4
+        set_property(GLOBAL PROPERTY RULE_LAUNCH_COMPILE "${CCACHE_PROGRAM}")
+        set_property(GLOBAL PROPERTY RULE_LAUNCH_LINK "${CCACHE_PROGRAM}")
+    else()
+        message(WARNING "ccache not found, disabling option")
+        set(SOFA_USE_CCACHE OFF)
```
guparan: Not sure this will uncheck the option, you might use FORCE here.

untereiner: I added a commit. Not shure its what you wanted..

PR #710, opened 2018-07-10
[SofaMiscMapping] Fix rigid barycentric mapping

IPeterlik:
FIX: a bug which appeared when the simulation is restarted without restarting SOFA. There is a map-vector (mapOrient) which was not cleared properly (unlike his friend "map") which was. 

This PR: 


epernod:
[ci-build][with-scene-tests]

untereiner:
do not forget to delete the branch

PR #726, opened 2018-07-24
[Topology Test][Mass] Fix some recent compilation warnings

epernod:
- use some size_t instead of int in the topology_test
- Change the update method from the mass to return the bool update as result.
- some small warning fix in other classes.

This PR: 


epernod:
[ci-build][with-scene-tests]

hugtalbot (approved):

PR #729, opened 2018-07-27
[All] replace some int/unit by size_t

epernod:
Only did some base classes that were generating multiple warnings.
And also the topology containers.

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/core/objectmodel/Link.h
+++ b/SofaKernel/framework/sofa/core/objectmodel/Link.h
@@ -555,7 +555,8 @@ public:
             }
 
             // Remove the objects from the container that are not in the new list
-            for (size_t i = 0; i != container.size(); i++)
+            unsigned int csize = (unsigned int)container.size();
+            for (unsigned int i = 0; i != csize; i++)
             {
                 DestPtr dest(container[i]);
                 bool destFound = false;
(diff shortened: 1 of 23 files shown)


epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

SofaKernel/framework/sofa/core/objectmodel/Link.h
```diff
@@ -555,7 +555,8 @@ class TLink : public BaseLink
             }
 
             // Remove the objects from the container that are not in the new list
-            for (size_t i = 0; i != container.size(); i++)
+            unsigned int csize = (unsigned int)container.size();
+            for (unsigned int i = 0; i != csize; i++)
```
guparan: What about `for (size_t i = 0, n = container.size() ; i < n ; i++)` ?

epernod: Yes I wanted to discuss that. Because it is a problem we have everywhere in Sofa.
if `i` is a `site_t`, then in the next line:
`DestPtr dest(container[i]);`
`container `should be of `<size_t>` which has a big impact. Nearly a "replace all" uint into size_t in objectModel::Link, BaseLink etc... 

You find the same problem in topology mapping where the containers `Map<unsigned int, unsigned int>` are holding the id of buffers which have max equal to `size_t`.

epernod: 🦈 🦈 🦈 
(Jaws music theme )

PR #742, opened 2018-08-01
[SofaPython][PSDE] Psde derive io

marques-bruno:
Depends on #740 
Part 2 of #629 This PR allows PSDEs to derive the type of their I/O datafields according the what they are linked to (position="@../whatever")

This PR: 


marques-bruno:
Small explanation about this PR:

While automatically deriving a PSDE input's type is not so hard when using links (you just need to parse the link's value, follow the path to the linked component's datafield, check its type and that's it), it is a bit trickier when it comes to outputs:
- During the creation of a PSDE's output field, we don't know yet what it's gonna be linked against, so deriving the type is not possible. We have to wait for the creation of the component that will take the PSDE's output as a link to derive it. This means that the steps to derive an output's datatype are the following:

- in Python, in the `parse()` method, call addNewOuput, and give it a datafield name, but no type and an input value field `addNewInput("MyName", value="")`
- in C++, a new datafield is created, but its datatype is void*. To know that this field has been created from a PSDE output, and is not yet properly connected, an alias is added on the data `addAlias("psde_output");`
- The next component is created in the scene, a c++ component this time, that knows the type of its inputs. 
- In its parse method, the link is found to the PSDE's output, but its type is incompatible (void* vs vector<Vec3d> for instance). 
- We check if the linked datafield has an alias "psde_output" and if so, remove that datafield and recreate it with the same field name, but this time the good datatype
- We set the newly created field as the parent of the parsed datafield in the component.

Done. 

To do so it was necessary to:
- Make creation of Data possible through a `getNewInstance()` method in Base
- Make BaseData removable from Sofa components in Base `removeData(BaseData*)`
- add in Binding_Base a method called from `addNewIO()` to find a data's type from its value when a link is provided (`deriveTypeFromParentValue()`)  
- Modify Base::parseField() to handle connecting to a parent value that has incompatible type when necessary
- Modify quite a bit the error handling process in addNewData() and parse() (which I hope to have done properly)

These explanations should help in understanding the code

bcarrez:
[ci-build]

applications/plugins/SofaPython/Binding_Base.cpp
```diff
@@ -115,55 +261,81 @@ BaseData* helper_addNewData(PyObject *args, PyObject * kw, Base * obj) {
...
+            bd = new Data<void*>();
+            bd->setName(dataName);
+            obj->addAlias(bd, "psde_output");
+        }
+        else if (std::string(dataName) != "type")
+        {
+            msg_warning(obj) << "No type provided for Data" << dataName << " with value " << val << ", creating void* data";
+            return bd;
+        }
+        else return new Data<void*>();
```
jnbrunet: Maybe just verify that those new pointers are deleted somewhere

guparan:
[ci-build][with-scene-tests]

applications/plugins/SofaPython/Binding_Base.cpp
```diff
@@ -75,6 +191,35 @@ static char* getStringCopy(char *c)
...
+
+    // if data is a link
+    if (value.length() > 0 && value[0] == '@')
+    {
+        std::string componentPath = value.substr(1, value.find('.') - 1);
+        std::string parentDataName = value.substr(value.find('.') + 1);
+
+        if (!o->getContext())
+        {
+	    msg_warning("SofaPython") << "No context created. Cannot find data link to derive input type";
```
tgaugry: A . is missing at the end of the line

applications/plugins/SofaPython/Binding_Base.cpp
```diff
@@ -115,55 +261,84 @@ BaseData* helper_addNewData(PyObject *args, PyObject * kw, Base * obj) {
...
+            }
 
-        tmp = PyDict_GetItemString(kw,"helptxt");
-        if (tmp!=nullptr){
-            dataHelp = getStringCopy(PyString_AsString(tmp));
-        }
+            tmp = PyDict_GetItemString(kw,"helptxt");
+            if (tmp!=nullptr){
+                dataHelp = getStringCopy(PyString_AsString(tmp));
+            }
```
tgaugry, line 282: maybe use a lambda ?

applications/plugins/SofaPython/Binding_Base.cpp
```diff
@@ -75,6 +191,35 @@ static char* getStringCopy(char *c)
...
+
+        if (!o->getContext())
+        {
+	    msg_warning("SofaPython") << "No context created. Cannot find data link to derive input type";
+            return nullptr;
+        }
+        BaseObject* component;
+        component = o->getContext()->get<BaseObject>(componentPath);
+        if (!component)
+	    msg_warning("SofaPython") << "no object with path " << componentPath << " in scene graph.";
```
tgaugry: `no`  -> `No`

jnbrunet (approved):

tgaugry (approved):

PR #743, opened 2018-08-01
[SofaPython][PSDE] Create object psde

marques-bruno:
Depends on #740 and #742 
Part 3 of #629 This PR allows you to create PSDE component in your scene the same way you do with Cpp engines, by calling createObject() and  passing your Input/output datafields directly there.

This PR: 


marques-bruno:
A bit of explanation about this PR:
When a attribute is passed to createObject, it is added to the BaseObjectDescription of the component. This means that, in cpp, the parse() method, that takes a BaseObjectDescription in parameter, has this information. The problem is, that in the PythonScriptDataEngine, the parse method calls the python version of the parse() method, which doesn't take this BaseObjectDescription in parameter (maybe that should change btw, since its mainly a simple key/value dictionnary after all...)
The datafields could be added directly from the C++, but since most of the error handling in the creation of derived I/Os in #742 is performed in the bindings, it would mean recoding a lot of redundant code.
Since I didn't want to touch the parse() method, I created an additional method, that can (or not) be used in PSDEs, named "datalinks". This method is called right after the parse() binding, and takes a filtered version of the BaseObjectDescription as an argument (all datafields already created in the parse() method are removed from the dictionnary). The function then calls addNewInput() on each of the leftover fields.

I'm certain there is a cleaner way of doing that.
- I was thinking maybe calling directly the python  addNewInput() method from the cpp parse() method, to avoid the creation of this extra module function, but I am not sure I know how to do that.  
- Another idea would be to pass the BaseObjectDescription in the parse() method of the PSDE (as it should be done IMHO), and let the user decide whether or not he wants to loop over those fields to create inputs.. more error prone in my opinion.

But you guys have good input on that ;)

marques-bruno:
[ci-build]

sescaida:
I like this feature, but I feel a little bit divided about the fact that it introduces yet another nuance to the creation of components in Sofa. If a novice user reads the python code and sees the createObject()-call he can get confused: Whereas in other components (written in C++) this syntax allows the user to understand that this component has a Data with a certain name, here it is an instruction to add a Data to the component (the PSDE).

marques-bruno:
@damienmarchal @tgaugry @jnbrunet, what do you think of @sescaida 's comment?

I agree that it makes it a bit tricky to look into the inner mechanisms of the creation of datafields from the createObject method.
I tried to make it as clear / explicit / fail-safe as possible though: 
- datafields passed through the createObject() function MUST be inputs fields referenced in the `args` dictionnary present in the Python class and there is a nice warning showing up when you try to pass an argument that doesn't exist in the args `dictionary` to createObject
- The user needs to explicitely use the decorator function in the SofaPython module on the parse() method, so he kinda has to know what the decorator does
- The syntax is as close as possible to the syntaxt of normal sofa::core::DataEngines created in python:
     ```node.createObject('TransformEngine', input_position='@component.data', translation='0 0 0')```
    would translate into
    ```node.createObject('PythonScriptDataEngine', input_position='@component.data', translation='0 0 0', filename=__file__, classname='PyTransformEngine')```

I personnally believe it's a must-have feature, because, let's say you have a python scene in which you have to create 5 times the same PSDE component, a custom TransformEngine for instance:
- Without this PR, you would have to create 5 separated classes, with the exact same implementation, just a different value in the addNewInput() in parse(). This is very redundant. 
- Passing the inputs values through createObject allows you to create a component only once, and use it as many times as you want in your scene afterwards.

sescaida:
@marques-bruno you're right! You had already mentioned that this would improve the reusability of PSDE-Python-Code written by the user. In that case I am in favor of this feature. About the intuitiveness: do you (and the others) think it would make sense to have an additional argument in createObject() that explicitly has to be set to use this feature of "inline" PSDE I/O creation?

marques-bruno:
@sescaida you mean a (boolean) flag passed to createObject that enables/disables the parsing of the  kwargs passed to createObject?

It's a god idea, I think. Another way of guaranteeing that the user knows what the call really does could be to leave the `createObject` method exactly as it currently is, and create a new method, maybe `createEngine`, that implements the what this PR does? 
Advantage of this would be that the Object type (`PythonScriptDataEngine`) could entirely disappear in favor of the PSDE's class name, currently passed through the `className` attribute. We could also make filename an optional attribute, taking `__file__` as a default argument, making the whole call even more resembling Sofa's traditional object creation, without affecting the `createObject()` method

The following calls would then do the same thing:

```
## Old call style:
node.createObject('PythonScriptDataEngine', input_position='@component.data', translation='0 0 0', className='MyPSDEEngine', filename=__file__)
## New syntax:
node.createEngine('MyPSDEEngine', input_position='@component.data', translation='0 0 0')
```

What do you say?

sescaida:
@marques-bruno:

>  you mean a (boolean) flag passed to createObject that enables/disables the parsing of the kwargs passed to createObject?

Exactly! It would be FALSE by default and wouldn't need to be specified if "inline" I/O-creation is not used.

About introducing a new function to create the PSDEs I personally think it is nice for this purpose, but I would probably leave it out for the sake of consistency regarding how all other components are created.

marques-bruno:
> It would be FALSE by default and wouldn't need to be specified if "inline" I/O-creation is not used

Actually, I didn't think of it immediately when you said it, but this is already what the current implementation does (kind of, the implementation is different but the concept is the same):

if the decorator `@sp.ParseInputs` is not set on the parse method, the passing of arguments through the createObject() is disabled, and the input argument will be ignored.
setting @sp.ParseInputs is kind of the equivalent to a boolean flag in the class allowing "inline" definitions. The only difference is that it's made at the declaration step and not during the instantiation of the class. the name (ParseInputs) is maybe not very clear. it could be called differently. "fwd_inline_args"  or "allow_inlines"..?

marques-bruno:
@sescaida, I just discovered a syntax, that exists since, well ... ever, and makes this whole PR completely useless....:

```py
import Sofa
import os

class PyEng(Sofa.PythonScriptDataEngine):
    def __init__(self, ctx, name, positions):
        self.name = name
        self.addNewInput('positions', datatype="vector<Vec3d>", value=positions, help="test")
        self.addNewOutput('average', datatype="Vec3d", help="average")
    
    def parse(self, arg):
        pass
    
    def init(self):
	pass
    
    def update(self):
        value = [0, 0, 0]
        for pos in self.positions:
            value[0] += pos[0]
            value[1] += pos[1]
            value[2] += pos[2]
        value[0] /= len(self.positions)
        value[1] /= len(self.positions)
        value[2] /= len(self.positions)
        self.cleanDirty()
        self.average = [[-value[0],-value[1],-value[2]]]
        pass


def createScene(rootNode):
    loader = rootNode.createObject("MeshVTKLoader", name="loader", filename='mesh/liver.vtk')
    psde = PyEng(rootNode, name="PSDE", positions=loader.findData('position').getLinkPath())
    rootNode.createObject('TransformEngine', name='TE', input_position="@loader.position", translation=psde.findData('average').getLinkPath())
    rootNode.createObject('OglModel', name="visu", position="@TE.output_position", triangles="@loader.triangles",
                          color="1 0 0 1")
```

With this syntax, I can create my engines definition once, and create as many as I want, on the fly, with any parameter I want for my input fields. It's actually even better, because:
- I don't have to create inputs with the fields I pass to the constructor. It could be simple data (addNewData vs addNewInput)
- Construction of datafields is more explicit, thus less error-prone
- No need to go search for a decorator function somewhere in SofaPython, or to declare a dictionary of default arguments, everything is performed in the constructor

So, yeah, this PR is now virtually useless.
There's one thing I don't like with this syntax though, it's that the creation of the object is performed at the same time as it is assigned to the scene graph, which, in my opinion is not always desirable.
Something closer to this would be neat:
```py
rootNode.addObject(PyEng(name='PSDE', positions=loader.findData('position').getLinkPath())
```

I'll work on that in the following days :)

sescaida:
@marques-bruno: Inheriting from the PythonScriptDataEngine seems to be a good solution! I didn't really know this was possible, but like you describe, covers most of the intended use-case. I'll keep it in mind in case I might need it myself. Ohh, and maybe it makes for a good example for a more advanced usage of the PSDE that could be added to the codebase?

PR #751, opened 2018-08-07
[SofaPython][PSDE] Update on demand as designed initially

marques-bruno:
Dependency on #760 

This PR mimics the behavior of the DataEngine, by binding the cleanDirty method in python, and not calling cleanDirty before calling update()
Furthermore, update() is not called from OnBeginAnimationStep events anymore, but only through the natural DDGNode update mechanism.

This PR: 

### DIFF
--- a/applications/plugins/SofaPython/Binding_PythonScriptDataEngine.cpp
+++ b/applications/plugins/SofaPython/Binding_PythonScriptDataEngine.cpp
@@ -68,6 +68,14 @@ static PyObject * PythonScriptDataEngine_parse(PyObject * self, PyObject * /*arg
     Py_RETURN_NONE;
 }
 
+
+static PyObject * PythonScriptDataEngine_cleanDirty(PyObject * self, PyObject */* args*/)
+{
+    ScriptDataEngine* _this = sofa::py::unwrap<ScriptDataEngine>(self);
+    _this->cleanDirty();
+    Py_RETURN_NONE;
+}
+
 struct error { };
 
 template<class T>
@@ -109,7 +117,8 @@ static PyObject * PythonScriptDataEngine_new(PyTypeObject * cls, PyObject * args
 SP_CLASS_METHODS_BEGIN(PythonScriptDataEngine)
 SP_CLASS_METHOD(PythonScriptDataEngine,parse)
 SP_CLASS_METHOD(PythonScriptDataEngine,init)
-SP_CLASS_METHOD(PythonScriptDataEngine,update)
+SP_CLASS_METHOD_DOC(PythonScriptDataEngine, update, "method called everytime a component tries to access one of this engine's outputs while at least one of its inputs is dirty")
+SP_CLASS_METHOD_DOC(PythonScriptDataEngine, cleanDirty, "A method to clean the engine's inputs. To call before trying to access the outputs")
 SP_CLASS_METHODS_END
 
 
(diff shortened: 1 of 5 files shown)


marques-bruno:
I'm personally still puzzled by this cleanDirty method.. Sometimes, not calling it has little to no effect at all, sometimes, it puts my engines in cyclic recursions between their I/O's getValues...
Also, I would expect "false" as a result of calling "isDirty()" on an I/O after a call to cleanDirty(), but this doesn't seem to be the case.

sescaida:
@marques-bruno: can you provide examples for the behavior you describe? (maybe add them to the PR) I don't have as much experience with the update-on-demand feature. Pertaining to the discussion in PR#629 (should we use on-demand updating or events) I would also appreciate the opinion of the others. The pros of on-demand updating is consistency with the DataEngines, the pros of the event-based approach is simplicity. @marques-bruno has been raising the issue of code maintainability regarding the update-on-demand feature of DataEngines for a while and this PR seems a good opportunity to decide what the developments in the future could be. For me the questions are: Refactor the implementation? Document the behavior in a more understandable way? Simplify/Change behavior alltogether?

marques-bruno:
@sescaida Thanks for the review
The reason why I didn't add examples for this PR is because the current examples themselves, without any modification, demonstrate that it works as-is, thus showing that the call to update() in the handleEvent method was redundant with the default DDGNode's update mechanism.

Basically, if your scene is well created (links are properly made between datafields & data are correctly accessed by components lower in the scene graph), everything should work just as well as before.

The only difference I would say, is that the user has the hand over the cleanDirty() method, which allows him to fix use cases where a read acces is performed on an input, after setting an output's value has been set, in update().

as for whether it is a good idea to keep that system, I must say, that I agree with you, as we discussed before: Bringing the complexity of the DDGNode's mechanism in Python might be something we want to avoid.
Although, I think that the good approach to fight this issue would rather be to solve the problem on the C++ side (simplifying the Data-Update mechanism) than to diverge from it by simplifying the Python process.
I think @damienmarchal woul be the best placed to give us a feedback on this question, maybe? (When he'll be back from his holidays of course!)

sescaida:
@marques-bruno: I mean an example for the following behavior you mentioned:
>  I'm personally still puzzled by this cleanDirty method.. Sometimes, not calling it has little to no effect at all, sometimes, it puts my engines in cyclic recursions between their I/O's getValues...

marques-bruno:
@damienmarchal @epernod, Taking @sescaida's comment into account, I tried to create an example showing the undesirable behavior related to the cleanDirty() call, unsuccessfully.
The comments for the sofa::core::DataEngine class in Sofa are very clear about the order in which the update() method should be implemented: first call update / getValue / updateIfDirty on your *inputs*, then call cleanDirty, then call setValue on your outputs.

I have had issues in the past related to this specific cleanDirty() method, but  for some reason I can't reproduce it. 
Does any of you understood the problem enough to create a small example causing the issue? I'll keep digging on my side.

marques-bruno:
@sescaida, you can find [here](https://github.com/marques-bruno/snippets/blob/master/testDataEngine.cpp) a test code for Sofa showing what happens when cleanDirty is called at the wrong moment. After testing a little bit the engines, I noticed that the only case where it wouldn't work (to have all the data updates & cleanDirty() before calling the update() method would be in the very unlikely case where the update() method would fetch an earlier engine in the dependency graph through getContext, and set its output value. this is very unlikely, which means that, apart for some performance reasons, I don't see why it would be important to bind the cleanDirty() method in python. Since python is much slower anyway, I think this PR does not make much sense anymore. although, I still insist on removing the call to update() from handleEvent, since it is redundant with its normal on-demand calling mechanism. :) I'll update the PR soon!

marques-bruno:
[ci-build][with-scene-tests]

PR #752, opened 2018-08-10
[Gui][HeadlessRecorder]Remove avcodec dependency in HeadlessRecorder.h

tgaugry:
Forward declare a class, so that avcodecs headers are only required when compiling the cpp.
That way, this header can be included even if FFmpeg headers are not in context (which lead to compilation failure).

--> Can this change be also backported to 18.06 branch ?

This PR: 


guparan:
Ok for backport :+1:

PR #755, opened 2018-08-26
[All] Fix some compilation warnings

epernod:
- Fix some size_t to uint int conversion warnings
- Fix some double to float warnings

- When looping on topology buffer, use the typedef corresponding to id instead of using uint or int. 

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/core/objectmodel/BaseLink.cpp
+++ b/SofaKernel/framework/sofa/core/objectmodel/BaseLink.cpp
@@ -60,9 +60,9 @@ BaseLink::~BaseLink()
 /// Print the value of the associated variable
 void BaseLink::printValue( std::ostream& o ) const
 {
-    std::size_t size = getSize();
+    unsigned int size = (unsigned int)getSize();
     bool first = true;
-    for (std::size_t i = 0; i<size; ++i)
+    for (unsigned int i = 0; i<size; ++i)
     {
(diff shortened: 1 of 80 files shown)


SofaKernel/framework/sofa/core/objectmodel/BaseLink.cpp
```diff
@@ -60,9 +60,9 @@ BaseLink::~BaseLink()
 /// Print the value of the associated variable
 void BaseLink::printValue( std::ostream& o ) const
 {
-    std::size_t size = getSize();
+    unsigned int size = (unsigned int)getSize();
```
hugtalbot, line 63: You are finally no using the size_t
Is there a reason to prefer the UI ?

epernod:
[ci-build][with-scene-tests]

SofaKernel/framework/sofa/core/objectmodel/BaseLink.cpp
epernod: in this case in the next line:
`std::string path = getLinkedPath(i);`
method `getLinkedPath` definition should have a size_t as argument...

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

epernod:
[ci-build][with-scene-tests]

PR #764, opened 2018-09-04
[SofaKernel] Move ScriptEvent class from SofaPython to core/objectModel

epernod:
Like the port salut, everything is in the title. 

This PR: 

### DIFF
--- a/applications/plugins/SofaPython/PythonScriptEvent.h
+++ b/applications/plugins/SofaPython/PythonScriptEvent.h
@@ -25,7 +25,7 @@
 #include "PythonCommon.h"
 
 #include <SofaPython/config.h>
-#include "ScriptEvent.h"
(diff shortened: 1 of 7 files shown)


guparan:
Ok but why ?

epernod:
just because it is useful in the rest of Sofa in cpp without having sofaPython. 
If you have a better place in /modules it is ok for me. I looked but I saw all events where stored in objectModel

applications/plugins/SofaPython/PythonScriptEvent.h
```diff
@@ -25,7 +25,7 @@
 #include "PythonCommon.h"
 
 #include <SofaPython/config.h>
```
marques-bruno: Are those still necessary?

marques-bruno (approved):
RAS, clean PR

PR #766, opened 2018-09-05
[CMake] FIX: cyclic recursion 

marques-bruno:
in sofa_install_libraries:

cyclic recursions when globbing for external libraries due to symlinks resolution.
I have done this fix for about a month and a half now, without noticing any side effects, but I can't tell if there are any.

This PR: 


guparan:
Just had a look and it seems not messing with installation nor packaging :+1:

guparan (approved):

PR #813, opened 2018-11-08
[SofaHelper] Change drawTriangle and drawQuad with internal functions

hugtalbot:
FIX issue #757 

This PR: 


guparan:
@hugtalbot Could we have a brief explanation on why this is needed (difference between draw and internalDraw)? Thanks :smile: 

EDIT: Got it: draw calls glBegin and glEnd, internalDraw does not.

jnbrunet (approved):

PR #821, opened 2018-11-09
[examples] Fix scenes having issue with CollisionGroup

hugtalbot:
Chains are behaving normally now.
DefaultCollisionGroup seems to bring a different behavior by solving contact (with penalties) when the all objects are considered as one system

This PR: 


guparan:
@hugtalbot  Does that mean we should not use CollisionGroup anymore? Is the component broken?

guparan:
[ci-build][with-scene-tests]

PR #878, opened 2019-01-02
[SofaBaseTopology] Change triangles orientation in tetrahedron

epernod:
Right now for a tetrahedron: [P0, P1, P2, P3] (see picture below)
The 4 triangles saved in the tetrahedron are:
- T0: [P1, P3, P2]
- T1: [P2, P3, P0]
- T2: [P3, P1, P0]
- T3: [P0, P1, P2]

This means the 4 triangles are clockwised oriented and thus their normals are going inside the tetrahedron. If there is a special reason for that I couldn't find it in the doc. 

image

As Gmsh nice ascii picture (from gmsh full doc) and the 2nd picture suggest. I changed to have counter-clockwise orientation so triangles on borders are by default well oriented to have normals going out.

Then, T0 being the 2D plan [u,v], T1 sharing vector u and then T2 and T3 to close the tetrahedron
So new triangles are:
- T0: [P0, P2, P1]
- T1: [P0, P1, P3]
- T2: [P1, P2, P3]
- T3: [P0, P3, P2]

image

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaBaseTopology/MeshTopology.cpp
+++ b/SofaKernel/modules/SofaBaseTopology/MeshTopology.cpp
@@ -255,26 +255,22 @@ void MeshTopology::TriangleUpdate::doUpdate()
     Triangle tr;
     unsigned int triangleIndex;
     unsigned int v[3],val;
+    const unsigned int trianglesInTetrahedronArray[4][3]= {{0,2,1}, {0,1,3}, {1,2,3}, {0,3,2}};
     /// create the m_edge array at the same time than it fills the m_trianglesInTetrahedron array
     for (unsigned int i = 0; i < tetrahedra.size(); ++i)
     {
         const Tetra &t=topology->seqTetrahedra.getValue()[i];
-        for (unsigned int j=0; j<4; ++j)
+        for (TriangleID j=0; j<4; ++j)
         {
-            if (j%2)
-            {
-                v[0]=t[(j+1)%4]; v[1]=t[(j+2)%4]; v[2]=t[(j+3)%4];
-            }
-            else
-            {
-                v[0]=t[(j+1)%4]; v[2]=t[(j+2)%4]; v[1]=t[(j+3)%4];
-            }
-            //		std::sort(v,v+2);
+            for (PointID k=0; k<3; ++k)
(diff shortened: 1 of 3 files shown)


epernod:
small change but might be breaking so I put it as independant PR.

epernod:
tetra2triangletopologicalmapping_00000003

`examples/Components/Topology/Tetra2TriangleTopologicalMapping.scn` normals display with the fix.

epernod:
[ci-build][with-all-tests]

jnbrunet:
Oups sorry @hugtalbot , I didn't see the notification that you ping me in this thread.

I don't know why the triangles were oriented this way. For the breaking part, if someone was using them for BC or collisions, it will definitely break.

The next thing I can see is, you are not only changing the orientation of the triangle, but also the ordering of these triangles inside the tetrahedron. If I do `TetrahedronSetTopologyContainer::getLocalTrianglesInTetrahedron (const TriangleID i)`, I no longer have the same triangle as before. That might also break something, although I can't find anything with a quick search in Sofa using it.

~~Finally, these indices are hardcoded in the cpp, and not accessible from the outside. So if someone is using them, it means he copy/pasted them in his source file. They will no longer match the one of the tetra container. This might have weird effects. It could be a good idea to make them available either directly from the header or inside a get function so that in the future, anyone using them will be in sync with the tetra container.~~ They have `getLocalTrianglesInTetrahedron` for this, I'm a little bit tired !

I guess it is ok to go forward with this, but have this PR in mind if you got "my pressure FF doesn't work anymore, what are you guys doing????" on the forum :-)

jnbrunet:
Just a small side note: `Tetra2TriangleTopologicalMapping` seems to be inverting the triangles from the container, so you might want to remove the default swap since it is no longer required

epernod:
thx @jnbrunet for your answer.
Indeed there is a missmatch with the Tetra2TriangleTopologicalMapping but I removed that in another PR. In fact the mapping was inverting the triangles because their orientation in the tetra-container was not logical to have a well oriented surface. 

Let see if it breaks something. At least people know on whom to shoot.

epernod:
now that I have fixed the tests on tetrahedron yes.

Issue #810, opened 2018-11-07
check which component is using : handleTopologyChange() function or TopologicalData


hugtalbot:
Help on the Topology work, in finding the work affected by handleTopologicalChange()

Forcefields handling topology change:
- TriangularFEMForceField
- TriangularBendingSprings
- RegularGridSpringForceField
- QuadularBendingSprings

Forcefields not handling:
- EdgePressureForceField
- QuadBendingSprings

List of components with ```handleTopologyChange```  method

1. void BarycentricMapping<TIn, TOut>::handleTopologyChange ( core::topology::Topology* t )
2. void BarycentricMapperHexahedronSetTopology<In,Out>::handleTopologyChange(core::topology::Topology* t)
3. void IdentityMapping<TIn, TOut>::handleTopologyChange()
4. ~~void UniformMass<DataTypes, MassType>::handleTopologyChange()~~ PR #2377 
5. ~~void VisualModelImpl::handleTopologyChange()~~ PR #2300 
6. void SpringForceField<DataTypes>::handleTopologyChange(core::topology::Topology *topo)
7. void TLineModel<DataTypes>::handleTopologyChange()
8. ~~void PointLocalMinDistanceFilter::handleTopologyChange()~~ _Component removed_
9. ~~void TriangleLocalMinDistanceFilter::handleTopologyChange()~~ _Component removed_
10. ~~void TTriangleModel<DataTypes>::handleTopologyChange()~~ PR #903
11. inline void HexahedronFEMForceField<DataTypes>::handleTopologyChange()
12. inline void TetrahedronFEMForceField<DataTypes>::handleTopologyChange()
13. ~~OglTetrahedralModel< gpu::cuda::CudaVectorTypes<TCoord,TDeriv,TReal> >~~ _Component removed_
14. CudaVisualModel
15. ~~TetrahedronModel::handleTopologyChange()~~ PR #1354
16. void PositionBasedDynamicsConstraint<DataTypes>::handleTopologyChange()
17. ~~void UncoupledConstraintCorrection< DataTypes >::handleTopologyChange()~~ PR #1115
17. void BarycentricMapperHexahedronSetTopology<defaulttype::Vec3Types, defaulttype::Rigid3Types>::handleTopologyChange(core::topology::Topology* t)
18. void DeformableOnRigidFrameMapping<TIn, TInRoot, TOut>::handleTopologyChange(core::topology::Topology* t)
[...]

PR #880, opened 2019-01-04
[SofaKernel] Clean & Fix TopologyChangeVisitor and StateChangeVisitor behavior

epernod:
This should fix all the scene crashing when performing topological change with some mechanicalMapping in the scene.

Like for example identityMapping between the mechanicalObject and the OglModel.

This is a temporary fix. Those 2 visitors will be removed as soon as all the component listed in #810 will be updated to use TopologicalData.

This PR: 


epernod:
[ci-build][with-all-tests]

PR #897, opened 2019-01-16
[CMAKE] Add check to prevent the inclusion of non-existant file in cmake 3.13

tgaugry:
Cmake 3.13 does not seem to allow the inclusion of `CPackNSIS` anymore.
This add a simple check on cmake version to prevent cmake configure failure.

This PR: 


guparan:
I definitely should rework that part as it is causing too much trouble...

PR #903, opened 2019-01-22
[SofaMeshCollision] Fix TriangleModel to handle topology changes

epernod:
Fix crash when removing triangles in scene using TriangleModel component.

Clean and comment the class while undusting it (sorry hard to read the diff)

- remove serr/sout
- change all protected members prefix
- rewritte init and updateFromTopology methods
- Remove the handleTopologyChange method. As this class use directly the topology buffers, no need neither for TopologyData. Just check the topology revision and update the component if topology has changed.
- Update RemovingTrianglesProcess and AddingTrianglesProcess tests scenes to test the TriangleModel

This PR: 


epernod:
[ci-build][with-all-tests]

epernod:
[ci-build][with-all-tests]

PR #906, opened 2019-01-25
[MultiThreading] TaskAllocator Interface

fspadoni:
Add TaskAllocator interface to customize the task dynamic allocations and its standard implementation using operator new.

The PR also include a change of return type of Task::run() function from bool to enum MemoryAlloc to let the scheduler know if the task must be deallocated or not just after it has been executed.

This PR: 


fspadoni:
[ci-build] [with-scene-tests] [with-regression-tests]

PR #933, opened 2019-02-13
 [SofaBaseVisual] Add default texcoord in VisualModel

epernod:
When the Data texcoord is not set using a Mesh file or using a Data link from another component no mechanism is present in VisualModel to compute default UV texcoord leading to incompatible buffer size between vertices and texcoords.

This PR add a method to compute default texcoords using a sphere projection. Not perfect but at least create texcoords for volumique mesh with surface mapping.

Here are some results examples:

liver-tetra2triangle_00000001
liver-tetra2triangle_00000002

Mesh sphere projection used:
liver-tetra2triangle_00000003

This PR: 


epernod:
Armadillo with better Abs
tetra2triangletopologicalmapping_00000002
tetra2triangletopologicalmapping_00000003

epernod:
Support topological changes:
tetra2triangletopologicalmapping_00000004

epernod:
[ci-build][with-all-tests]

epernod:
[ci-build][with-all-tests]

PR #937, opened 2019-02-17
Add possibilities to draw lines on surfaces in DrawTool

hdeling:
Add 2 functions void enablePolygonOffset(float factor, float units); and l void disablePolygonOffset();
to handle the depth offset to draw lines on surfaces.

This PR: 


guparan:
Thanks for the PR @hdeling !
Can you check it out @fredroy ?

epernod (approved):

epernod:
This is the result expected I suppose:
image

PR #958, opened 2019-03-07
[Cgal plugin] Fix windows cmake dll path and add a scene example

epernod:
- Not sure this is the good way to do it with cmake findCGal but otherwise the dll are not linked (if not on env_path)

- Add a quick scene test for cgal without image plugin.

This PR: 


guparan:
So I tried to play a bit with CGALPlugin and realized we can use CGAL in header-only mode. So, why should we copy some dlls?
I made a branch here: https://github.com/guparan/sofa/tree/clean_cgalplugin_cmake

EDIT: here is a patch to try my changes in this branch: [cgal-plugin_fix.patch.txt](https://github.com/sofa-framework/sofa/files/2983964/cgal-plugin_fix.patch.txt)

guparan:
Did you have some time to test my patch @epernod ?

epernod:
ok, it works if I don't build cgal, so in header only (only execute cgal-setup.exe, no visual build after this step)
Otherwise it fails at compilation: error: cannot open input file  'CGal.lib' when building plugin.

guparan:
Header only usage should work even if CGAL has been built before. I will try to force header_only, there is a CMake variable for that in CGAL.

epernod:
@guparan do you remember what is the status of CGal plugin config on windows?

PR #959, opened 2019-03-07
[CMake] FIX build/install plugins directory

vmagno:
There was an error in the name of the (origin) directory to install. Not sure about the destination though. This fix has no impact on building Sofa, only on its installation.

This PR: 

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,7 +321,7 @@ if(SOFA_INSTALL_RESOURCES_FILES)
 endif()
 
 file(WRITE "${CMAKE_BINARY_DIR}/plugins/README.txt" "This folder will be automatically scanned by the Plugin Manager.")
-install(DIRECTORY applications/plugins/ DESTINATION plugins COMPONENT resources)
+install(DIRECTORY ${CMAKE_BINARY_DIR}/plugins/ DESTINATION plugins COMPONENT resources)
 
 
 #######################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -321,7 +321,7 @@ if(SOFA_INSTALL_RESOURCES_FILES)
 endif()
 
 file(WRITE "${CMAKE_BINARY_DIR}/plugins/README.txt" "This folder will be automatically scanned by the Plugin Manager.")
-install(DIRECTORY plugins/ DESTINATION plugins COMPONENT resources)
+install(DIRECTORY applications/plugins/ DESTINATION plugins COMPONENT resources)
 
 
 #######################


jnbrunet:
Haaa now it makes sense! We got this bug yesterday.

Thanks Vincent !

guparan:
Hi, this is not a bug but a new feature I introduced in https://github.com/sofa-framework/sofa/pull/913 to provide (in build and install dirs) a folder that shall always be scanned by PluginManager when searching for a plugin (like the bin/ folder is).

The idea was to get an easy way to add/remove plugin binaries in an installed SOFA **without polluting the bin/ folder** which contains stuff I'm not supposed to touch (or scared to touch).

The typical use case is: 
I have SOFA binaries (downloaded from sofa-framework.org or provided by someone) and I have a plugin binaries. How to run the examples of the plugin in runSofa?

Many ways to do that:
1. launch runSofa, open Plugin Manager, load the plugin, open the scene
2. edit the scene to add `<AddPluginRepository path="/path/to/the/plugin"/>` and `<RequiredPlugin name="the_plugin"/>`, launch the scene in runSofa 
3. copy/paste the plugin binaries (library or complete folder) in the bin/ folder of SOFA, add `<RequiredPlugin name="the_plugin"/>` or add the_plugin in bin/plugin_list.conf.default, launch the scene in runSofa 
4. **[NEW]** copy/paste the plugin binaries (library or complete folder) in the plugins/ folder of SOFA, add `<RequiredPlugin name="the_plugin"/>` or add the_plugin in bin/plugin_list.conf.default, launch the scene in runSofa 

(1) is perfect for a new user but becomes heavy quite soon
(2) is more agile but has limitations for plugins that need to be loaded **before** the scene creation (e.g. for Python env init)
(3) solves the limitation of (2) if we use plugin_list.conf but, as I said, putting stuff in bin/ is scary
(4) is what I proposed to separate "plugins that came with SOFA" from "plugins I add myself"


So now about this PR. What you proposed here is to install ALL the sources of applications/plugins into this new plugins folder. ~~Why? 
@jnbrunet You got a bug yesterday, what was it?~~

EDIT: Understood what the bug is, see my review :-)

CMakeLists.txt
```diff
@@ -321,7 +321,7 @@ if(SOFA_INSTALL_RESOURCES_FILES)
 endif()
 
 file(WRITE "${CMAKE_BINARY_DIR}/plugins/README.txt" "This folder will be automatically scanned by the Plugin Manager.")
-install(DIRECTORY plugins/ DESTINATION plugins COMPONENT resources)
+install(DIRECTORY applications/plugins/ DESTINATION plugins COMPONENT resources)
```
guparan: Actually the bug is real, as stated by @jnbrunet and @marques-bruno but the fix should be
`install(DIRECTORY ${CMAKE_BINARY_DIR}/plugins/ DESTINATION plugins COMPONENT resources)`

guparan (changes requested):

guparan (approved):

PR #963, opened 2019-03-11
[SofaPython][SofaGuiQt] REMOVE: public export of target SofaExporter

marques-bruno:
SofaExporter was a linked to SofaPython & SofaGuiQt under the PUBLIC scope while that dependency is internal & should not be exposed publicly

This PR: 


marques-bruno:
FYI I didn't find any other PUBLIC link of SofaExporter in sofa's CMakeLists

PR #968, opened 2019-03-14
[SofaKernel] Remove warnings

fredroy:
Remove warnings, mostly from missing override keyword.
(+ some old style C cast)

And fix a (potential) crash in MechanicalMatrixMapper (uninitialized pointer) 

This PR: 


fredroy:
[ci-build][with-all-tests]

PR #969, opened 2019-03-15
[SofaKernel] FIX: A few fix to compile on Mac OSX Xcode 9 and Linux gcc 7.3.0

fspadoni:

This PR: 


SofaKernel/framework/sofa/simulation/CMakeLists.txt
```diff
@@ -131,7 +131,10 @@ set(SOURCE_FILES
     InitTasks.cpp
 )
 
-
+# is this a compiler/linker version specific problem?
+if(CMAKE_SYSTEM_NAME STREQUAL Linux)
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread" )
```
tgaugry: [target_compile_options](https://cmake.org/cmake/help/v3.4/command/target_compile_options.html)

SofaKernel/framework/sofa/helper/CMakeLists.txt
```diff
@@ -353,6 +353,8 @@ if(SOFA-EXTERNAL_FFMPEG)
 
 endif()
 if(APPLE)
+    find_library(ICONV iconv)
```
guparan, line 355: This replaces the find_library already at line 342?

SofaKernel/framework/sofa/simulation/CMakeLists.txt
fspadoni: -pthread should be passed as linking  options:
target_link_options(${PROJECT_NAME} PUBLIC "-pthread")
This doesn't work because the target_link_options is  available only starting from CMake 3.13

PR #972, opened 2019-03-27
Deactivate Modeler by default, since it is deprecated

tgaugry:

This PR: 

### DIFF
--- a/applications/projects/CMakeLists.txt
+++ b/applications/projects/CMakeLists.txt
@@ -7,9 +7,7 @@ if(NOT Qt5Core_FOUND)
     set(SOFA_HAVE_QT_LIBRARIES 0)
 endif()
 
-if(SOFA_HAVE_QT_LIBRARIES)
-    sofa_add_application(Modeler Modeler OFF)
-endif()
+sofa_add_application(Modeler Modeler OFF)
 
 sofa_add_application(getDeprecatedComponents getDeprecatedComponents OFF)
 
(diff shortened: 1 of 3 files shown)


applications/projects/CMakeLists.txt
```diff
@@ -7,9 +7,7 @@ if(NOT Qt5Core_FOUND)
     set(SOFA_HAVE_QT_LIBRARIES 0)
 endif()
 
-if(SOFA_HAVE_QT_LIBRARIES)
-    sofa_add_application(Modeler Modeler OFF)
-endif()
```
guparan, line 12: Why? I would have kept the condition.

tgaugry: I did too at first, but it's a bit weird to do a check required for activation to then deactivate the thing. If the modeler require qt, then it should check for qt itself.

guparan: > If the modeler require qt, then it should check for qt itself.

I agree with that but you can't just remove the condition then, you should move it to Modeler's CMakeLists (and fail or warn+return if false).

tgaugry: Modeler does `find_package(Qt5 COMPONENTS Core QUIET)`
Maybe just change that to `REQUIRED` ?

guparan: I'm good with that :+1:

PR #985, opened 2019-04-08
[all] Several bug fixes

epernod:
Some bug fix from my working branch. List of fix:
- **[FixedConstraint]** fix bug when several invalid indices need to be removed. Need to parse all first and remove in a second stage.
- **[SphereModel]** FIX: display of collision sphere in wireframe mode.
- **[TetrahedronSetTopologyContainer]** FIX: triangle created with bad orientation in addTetrahedronProcess method.
- **[TriangleModel]** FIX: need to check if topology has changed and update the triangles if needed before computing triangleModel BBox
- **[PointModel]** Fix pointModel crash when removing points. need to be well resized in computeboundingtree method.
- **[SofaBaseTopology]** Add a single static trianglesOrientationInTetrahedronArray in baseMeshTopology.h and remove all the different occurences (with different values) in the cpp files.

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/core/topology/BaseMeshTopology.h
+++ b/SofaKernel/framework/sofa/core/topology/BaseMeshTopology.h
@@ -326,6 +326,9 @@ public:
 
 };
 
+// map of triangles orientation inside a tetrahedron.
+static const unsigned int trianglesOrientationInTetrahedronArray[4][3]= {{1,2,3}, {0,3,2}, {1,3,0},{0,2,1}};
+
 } // namespace topology
 
 } // namespace core
--- a/SofaKernel/modules/SofaBaseCollision/SphereModel.inl
+++ b/SofaKernel/modules/SofaBaseCollision/SphereModel.inl
@@ -143,7 +143,8 @@ void TSphereModel<DataTypes>::draw(const core::visual::VisualParams* vparams)
     //if (!vparams->isSupported(core::visual::API_OpenGL)) return;
     if (vparams->displayFlags().getShowCollisionModels())
     {
-        vparams->drawTool()->setPolygonMode(0,vparams->displayFlags().getShowWireFrame());
+        // Force no wireframe mode to draw sphere collision
+        vparams->drawTool()->setPolygonMode(0,false);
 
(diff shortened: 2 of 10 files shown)


epernod:
[ci-build][with-all-tests]

epernod:
[ci-build][with-all-tests]

SofaKernel/framework/sofa/core/topology/BaseMeshTopology.h
```diff
@@ -326,6 +326,9 @@ class SOFA_CORE_API BaseMeshTopology : public core::topology::Topology
 
 };
 
+// map of triangles orientation inside a tetrahedron.
+static const unsigned int trianglesOrientationInTetrahedronArray[4][3]= {{1,2,3}, {0,3,2}, {1,3,0},{0,2,1}};
```
hugtalbot, line 330: So now the only ordering is the following: {1,2,3}, {0,3,2}, {1,3,0},{0,2,1} ?

hugtalbot (approved):

SofaKernel/framework/sofa/core/topology/BaseMeshTopology.h
epernod: yes it is the original one.
The one I changed last month: ```const unsigned int trianglesInTetrahedronArray[4][3]= {{0,2,1}, {0,1,3}, {1,2,3}, {0,3,2}};```
should have the same behavior but it isn't the case. I don't have time to dig into it. This one works well.

epernod:
yes the change of trianglesInTetrahedron need to be propagated in the tests. I will check that.

SofaKernel/framework/sofa/core/topology/BaseMeshTopology.h
epernod: yes it is the original one.
The one I changed last month: ```const unsigned int trianglesInTetrahedronArray[4][3]= {{0,2,1}, {0,1,3}, {1,2,3}, {0,3,2}};```
should have the same behavior but it isn't the case. I don't have time to dig into it. This one works well.

PR #986, opened 2019-04-10
[SofaGeneralLoader] Compute subElement by default for Gmsh format

### LINKED ISSUE #631: MeshGmshLoader should create all edges
> createSubelements="true": Divide all n-D elements into their (n-1)-D boundary elements (e.g. tetrahedra to triangles).
> Meaning: for a triangulation, create all edges.
> 
> In this case the mesh file had only edges of the border but the FF were using all edges of the mesh.

by @epernod in #626

epernod:
This bug just pop up again in #967  and was referenced in issue #631 
Several Gmsh file have non coherent topology information, for example Triangulation with edges on the border only. This mess up information feeding the topology container.

If this behavior is wanted, use ``` <MeshGmshLoader createSubelements="0" />``` and be sure to link the different topology buffer data to the appropriate components.
For example:
``` 
<MeshGmshLoader name="loader" createSubelements="0" />

<Node name="triangulation" >
   <TriangleSetTopologyContainer name="tri_topo" triangles="@../loader.triangles" />
</Node>

<Node name="border" >
   <TriangleSetTopologyContainer name="edge_topo" edges="@../loader.edges" />
</Node>
```

fixes #631

This PR: 


epernod:
[ci-build][with-all-tests]

PR #1008, opened 2019-04-26
[SofaBaseTopology] Fix Crash when loading a vtk file generated by Gmsh using TetrahedronSetTopologyContainer as container

olivier-goury:
Good afternoon!
This PR proposes a fix for a crash when loading a vtk file that was generated by Gmsh.

`
 myNode.createObject('MeshVTKLoader', name='loader', filename='gmshMesh.vtk')
 myNode.createObject('TetrahedronSetTopologyContainer', src="@loader")
`

The mesh generated by default with Gmsh gives a list of edges and triangles that is incomplete and contains only edges and triangles on the boundary of the mesh. However the list of Tetrahedra is complete. This leads to a crash when using a TetrahedronSetTopologyContainer.
This PR tries to address this problem.

Cheers!
Olivier

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaBaseTopology/TetrahedronSetTopologyContainer.cpp
+++ b/SofaKernel/modules/SofaBaseTopology/TetrahedronSetTopologyContainer.cpp
@@ -327,6 +327,10 @@ void TetrahedronSetTopologyContainer::createTrianglesInTetrahedronArray()
             }
             else
             {
+                msg_error() << "Cannot find triangle " << j
+                    << " [" << t[(j + 1) % 4] << ", " << t[(j + 2) % 4] << ", " << t[(j + 3) % 4] << "]"                     
+                    << " in tetrahedron " << i;
+
                 m_trianglesInTetrahedron.clear();
                 return;
             }
@@ -380,17 +384,47 @@ void TetrahedronSetTopologyContainer::createTetrahedraAroundTriangleArray ()
     // first clear potential previous buffer
     clearTetrahedraAroundTriangle();
 
-    if(!hasTrianglesInTetrahedron())
+    if (!hasTetrahedra()) // this method should only be called when tetrahedra exist
+        createTetrahedronSetArray();
+
+    if (hasTetrahedraAroundTriangle()) // created by upper topology (inside createTetrahedronSetArray)
+        return;
+
+    const size_t numTetra = getNumberOfTetrahedra();
+    if (numTetra == 0)
+    {
+        msg_warning() << "TetrahedraAroundTriangle buffer can't be created as no tetrahedra are present in this topology.";
+        return;
+    }
+
+    if (!hasTriangles()) // this method should only be called when triangles exist
+        createTriangleSetArray();
+    
+    const size_t numTriangles = getNumberOfTriangles();
+    if (numTriangles == 0)
+    {
+        msg_warning() << "TetrahedraAroundTriangle buffer can't be created as no triangles are present in this topology.";
+        return;
+    }
+
+
+    if(!hasTrianglesInTetrahedron()) 
         createTrianglesInTetrahedronArray();
-    m_tetrahedraAroundTriangle.resize( getNumberOfTriangles());
-    if (m_trianglesInTetrahedron.size() !=0){
-        for (size_t i=0; i<getNumberOfTetrahedra(); ++i)
+
+    if (m_trianglesInTetrahedron.empty())
+    {
+        msg_warning() << "TetrahedraAroundTriangle buffer can't be created as trianglesInTetrahedron buffer creation failed.";
+        return;
+    }
+
+    m_tetrahedraAroundTriangle.resize(numTriangles);
+
+    for (size_t i=0; i<numTetra; ++i)
+    {
+        // adding tetrahedron i in the shell of all neighbors triangles
+        for (TriangleID j=0; j<4; ++j)
         {
-            // adding tetrahedron i in the shell of all neighbors triangles
-            for (TriangleID j=0; j<4; ++j)
-            {
-                m_tetrahedraAroundTriangle[ m_trianglesInTetrahedron[i][j] ].push_back( (TetrahedronID)i );
-            }
+            m_tetrahedraAroundTriangle[ m_trianglesInTetrahedron[i][j] ].push_back( (TetrahedronID)i );
         }
     }
 }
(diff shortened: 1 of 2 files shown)


epernod:
Hi,
not sure to understand, are the edges in the VTK correct regarding the tetra? Or do they correspond to the border?

olivier-goury:
> Hi,
> not sure to understand, are the edges in the VTK correct regarding the tetra? Or do they correspond to the border?

There are not consistent with the tetra. Basically, the tetra information contains all the tetras as you would expect, but the edges and triangles correspond only to the boundary of the mesh.

epernod:
Ok and you want to keep them or you want to recompute them? 
It is not easy to read the fixin the PR.

I made some change recently in the container to be sure that all info edges/triangles and tetrahedra are coherent.
If you want to create the border topology you should use TopologicalMapping. Or link data by data to different containers and not using src.

Otherwise to recompute the sub element, you should use a "hidden" option in the VTKLoader: See what has been done for Gmsh in PR #986

olivier-goury:
Hi,
I don't actually want to use them, it's just that it was crashing! I created the mesh in Gmsh but exported it in vtk. Exporting it in the msh format does not lead to a crash indeed.
I can understand that my fix is not super clear!  
Essentially, in the code when it starts using the edges and something goes wrong, I move on to the option like if no edges had been given (For example in the function createEdgesInTetrahedronArray() with the if (hasEdges) ). I hope it clarifies a little bit.

SofaKernel/modules/SofaBaseTopology/TetrahedronSetTopologyContainer.cpp
```diff
@@ -234,12 +199,50 @@ void TetrahedronSetTopologyContainer::createEdgesInTetrahedronArray()
                         foundEdge=true;
                     }
                 }
-
+                if (foundEdge == false)  // The edges and triangles in this mesh are only on the border: this mesh was probably created using Gmsh
```
epernod: and what happen if it enter this case? 
The container will stay without edges?

SofaKernel/modules/SofaBaseTopology/TetrahedronSetTopologyContainer.cpp
```diff
@@ -375,17 +386,16 @@ void TetrahedronSetTopologyContainer::createTetrahedraAroundTriangleArray ()
...
+    if (m_trianglesInTetrahedron.size() !=0){
     for (size_t i=0; i<getNumberOfTetrahedra(); ++i)
     {
         // adding tetrahedron i in the shell of all neighbors triangles
         for (TriangleID j=0; j<4; ++j)
         {
             m_tetrahedraAroundTriangle[ m_trianglesInTetrahedron[i][j] ].push_back( (TetrahedronID)i );
         }
-    }
+    }}
```
guparan: This block still needs some indentation

SofaKernel/modules/SofaBaseTopology/TriangleSetTopologyContainer.cpp
```diff
@@ -292,19 +259,19 @@ void TriangleSetTopologyContainer::createEdgesInTriangleArray()
...
 
         for (size_t edge=0; edge<numEdges; ++edge)  //Todo: check if not better using multimap <PointID ,TriangleID> and for each edge, push each triangle present in both shell
         {
             edgesAroundVertexMap.insert(std::pair<PointID, EdgeID> (m_edge[edge][0], (EdgeID)edge));
             edgesAroundVertexMap.insert(std::pair<PointID, EdgeID> (m_edge[edge][1], (EdgeID)edge));
         }
-
-        for(size_t i=0; i<numTriangles; ++i)
+        size_t i=0;
+        while(i<numTriangles && foundEdge==true)
```
guparan: `for ( size_t i = 0 ; (i < numTriangles) &&  (foundEdge == true) ; ++i )` ?

SofaKernel/modules/SofaBaseTopology/TriangleSetTopologyContainer.cpp
```diff
@@ -292,19 +259,19 @@ void TriangleSetTopologyContainer::createEdgesInTriangleArray()
...
         {
             const Triangle &t = m_triangle[i];
             // adding edge i in the edge shell of both points
-            for(unsigned int j=0; j<3; ++j)
+            int j=0;
+            while(j<3 && foundEdge==true)
```
guparan: `for ( unsigned int j = 0 ; (j < 3) && (foundEdge == true) ; ++j )` ?

SofaKernel/modules/SofaBaseTopology/TetrahedronSetTopologyContainer.cpp
```diff
@@ -207,20 +174,21 @@ void TetrahedronSetTopologyContainer::createEdgesInTetrahedronArray()
...
 
         for (EdgeID edge=0; edge<numEdges; ++edge)  //Todo: check if not better using multimap <PointID ,TriangleID> and for each edge, push each triangle present in both shell
         {
             edgesAroundVertexMap.insert(std::pair<PointID, EdgeID> (m_edge[edge][0],edge));
             edgesAroundVertexMap.insert(std::pair<PointID, EdgeID> (m_edge[edge][1],edge));
         }
-
-        for(size_t i=0; i<numTetra; ++i)
+        size_t i=0;
+        while(i<numTetra && foundEdge == true)
```
guparan: `for ( size_t i = 0 ; (i < numTetra) && (foundEdge == true) ; ++i )` ?

SofaKernel/modules/SofaBaseTopology/TetrahedronSetTopologyContainer.cpp
```diff
@@ -207,20 +174,21 @@ void TetrahedronSetTopologyContainer::createEdgesInTetrahedronArray()
...
         {
             const Tetrahedron &t = m_tetrahedron[i];
             // adding edge i in the edge shell of both points
-            for(EdgeID j=0; j<6; ++j)
+            EdgeID j=0;
+            while(j<6 && foundEdge == true)
```
guparan: `for ( EdgeID j = 0 ; (j < 6) && (foundEdge == true) ; ++j )` ?

guparan (changes requested):

olivier-goury:
Thanks for your comments Guillaume, I agree with your changes!

guparan (approved):

epernod (approved):

epernod:
I'm ok with the change regarding the edges. I can't see some border effect right now... and as all the non regression tests are green I'm 99% confident!
I just added some additional warning and tests inside the tetrahedron methods you fixed to match what has been done inside the TriangleSetTopologyContainer. You could double check if needed.
Thanks, I missed those one.
:+1:

PR #1015, opened 2019-04-29
[image] Fix image_gui plugin loading

hugtalbot:
Coming from [Mei remark on the forum](https://www.sofa-framework.org/community/forum/topic/image-viewer-in-image-plugin/)

- Add the automatic loading of the image_gui library when the plugin image is loaded (if the library is well there)
- Add formating in associated CMakeLists

With the correct loading of image_gui, now the ImagePlaneWidget will work fine when opening a ImageViewer component

This PR: 

### DIFF
--- a/applications/plugins/image/initImage.cpp
+++ b/applications/plugins/image/initImage.cpp
@@ -21,6 +21,7 @@
 ******************************************************************************/
 #include <image/config.h>
 #include <sofa/helper/system/config.h>
+#include <sofa/helper/system/PluginManager.h>
 
 #ifdef SOFA_HAVE_SOFAPYTHON
     #include <SofaPython/PythonFactory.h>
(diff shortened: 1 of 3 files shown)


applications/plugins/image/initImage.cpp
```diff
@@ -21,6 +21,7 @@
 ******************************************************************************/
 #include <image/config.h>
 #include <sofa/helper/system/config.h>
+#include <sofa/helper/system/PluginManager.h>
```
guparan, line 24: Maybe add `using sofa::helper::system::PluginManager;` here to avoid the long namespaces below :wink:

epernod: you are picky, for one line only, it is nearly more work to add the using :wink:

epernod (approved):

PR #1033, opened 2019-05-10
[SofaKernel] Remove boost::locale dependency

fredroy:
boost::locale was only once in FileRepository ; and since we can use std::locale from the STL, this library seems not useful anymore.

This PR: 


guparan (approved):

guparan:
[ci-build][with-all-tests]

epernod:
@guparan vs @fredroy  fight!

guparan:
Let me resolve those conflicts ;)

PR #1056, opened 2019-05-22
[SofaKernel] Clean output data when doUpdate in BoxROI

hugtalbot:
Fix crash on CI with scene _HexahedronForceFieldTopologyChangeHandling.scn_
Further to PR #1031 , the case where topo changes occur and X0 (points) size becomes zero (all points removed) the doUpdte() directly led to `return` without actually updating the output data field as being empty !
Now, all data are by default cleaned at the begin of the doUpdate() and filled only if X0.size is not zero and topology is not empty neither

This PR: 


olivier-goury:
Sounds good to me! Thanks for the fix!

olivier-goury (approved):

fredroy (approved):

fredroy:
[ci-build][with-all-tests]

PR #1064, opened 2019-05-23
[examples] Update the scene StandardTetrahedralFEMForceField.scn

hugtalbot:
Towards a green dashboard.

Scene was broken since the change on StaticSolver
But I am not convinced it was even working before, the simulation seems very unstable.

A python script handling the incremental loading is added.

This PR: 


jnbrunet (approved):
:+1:

PR #1078, opened 2019-06-12
[SofaTests] Fix small bugs in the Multi2Mapping_test

adagolodjo:

This PR: 


ChristianDuriez (approved):
It seems good: the Multi2Mapping should allow for having 2 different DataTypes as input. The previous implementation was considering that only 1 input DataType was possible. 
The bug fix allows now for 2 input DataTypes (In1Deriv / In2Deriv). 
I think the change made by @Younesssss should be approved.

PR #1083, opened 2019-06-14
Fix SetDirectory bug

tgaugry:
Fixes #992
Old version was just creating an object in the ctor...

This PR: 


tgaugry:
I call dibs on shooting the whole class once we get C++17 !

tgaugry:
Take a look at the removed overload. 
```
SetDirectory::SetDirectory(const std::string& filename)
{
    SetDirectory(filename.c_str());
}
``` 
is not valid object creation.
It is both wrong and unneeded.

As for the link with #992, latest message mention failing tests, which are due to the aforementionned bug.

guparan (approved):

PR #1088, opened 2019-06-21
[All] Cleans and fixes for v19.06

guparan:
- Changelog update
- CMake fix for Windows Dependency Pack
- CMake fixes for packaging (SofaMacros, GLU target missing)
- New script + CMake rules to add MacOS post-install custom steps (for packaging)
- CMake fixes for modularized projects (SofaExporter, SofaHaptics, SofaOpenglVisual, SofaPreconditioner, SofaSparseSolver)
- CMake clean tests dispatch in build tree, e.g now SofaConstraint_test build files go to `<build-dir>/SofaGeneral/SofaConstraint/SofaConstraint_test`

These changes also permit to generate binaries for all platforms with our CI. Continuous Deployment is not far!

This PR: 


guparan:
[ci-build][with-all-tests]

package.cmake
```diff
@@ -24,10 +24,10 @@ macro(setSofaFilePath name value)
 endmacro()
 ######################
 
-setSofaString(CMAKE_BUILD_TYPE Release)
+message("---------- Setting some options ---------")
 
 if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
-	setSofaString(CMAKE_CXX_FLAGS "-std=c++11")
+    setSofaString(CMAKE_CXX_FLAGS "-std=c++11")
```
marques-bruno, line 39: I didn't know we were compiling in c++11 when building on clang. Why is that?

package.cmake
```diff
@@ -24,10 +24,10 @@ macro(setSofaFilePath name value)
 endmacro()
 ######################
 
-setSofaString(CMAKE_BUILD_TYPE Release)
```
marques-bruno, line 27: What's the default build type without this?

modules/SofaExporter/CMakeLists.txt
```diff
@@ -44,10 +44,23 @@ list(APPEND SOURCE_FILES
...
+    TARGETS ${PROJECT_NAME}
+    INCLUDE_SOURCE_DIR "src/SofaExporter"
+    INCLUDE_INSTALL_DIR "SofaExporter/SofaExporter"
+    )
+
+# Enable tests if SofaTest found
+# Will pass only in out-of-SOFA builds
+# See modules/tests/CMakeLists.txt if building through SOFA
+find_package(SofaTest QUIET)
+if(SofaTest_FOUND)
```
marques-bruno, line 64: Just a simple question..: Is that the clean way of doing now, when compiling tests in a separated module / plugin?
SOFA_BUILD_TESTS should only be used in SofaKernel/framework ?

marques-bruno (approved):

package.cmake
guparan: Because Clang (old) version on our CI needs it :)

package.cmake
guparan: Default is Release. This was used to force Release when packaging but I prefer letting the user choose what type of package he wants (we never know).

modules/SofaExporter/CMakeLists.txt
guparan: I thinks this is the right way **for "modularized" modules** as long as we have a dependency mess with SofaTest (SofaTest depends on all SOFA modules).
As my comment says, SofaTest will be found only when building out-of-SOFA. 
That's why, for "modularized" modules that are still built in-SOFA but could now also be built out-of-SOFA, we need to have their tests added both in their `CMakeLists.txt` and in `modules/tests/CMakeLists.txt`.

For plugins, we should have something like this
```
option(MYPLUGIN_BUILD_TESTS "Build tests for MyPlugin" ON)
if(MYPLUGIN_BUILD_TESTS)
    add_subdirectory(MyPlugin_test)
endif()
```
without forgetting in MyPlugin_test/CMakeLists.txt: `find_package(SofaTest REQUIRED)`

PR #1089, opened 2019-06-21
FIX: import of SofaExporter: #include <SofaExporter/headerFile.h>

marques-bruno:
I noticed while trying to build SofaPython as out-of-tree that the build could not find <SofaExporter/OBJExporter.h>. It was because SofaExporter exposed its header files directly, instead of through a common folder named SofaExporter.
Thus this PR fixes the INCLUDE_INSTALL_DIR in SofaExporter/CMakeLists.txt

This PR: 


guparan:
Nice call @marques-bruno, however this fix (and others related) is included in my recently opened #1088 :wink: 
I let you close this one, unless you need the fix today?

marques-bruno:
@guparan thanks for the reply!
Actually, ASAP would be great: I need to this setup the CI for the plugin SofaPython3 alongside our other plugins @SofaDefrost.

guparan:
OK I'll update mine ;)

PR #1093, opened 2019-06-29
[SofaKernel] Add option in StiffSpringFF to track list of input springs

epernod:
- [SofaKernel] Add option in stiffspringFF to add a list of springs given 2 lists of indices

To be merged after #1131 
I tested the merge of both it works

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaDeformable/StiffSpringForceField.inl
+++ b/SofaKernel/modules/SofaDeformable/StiffSpringForceField.inl
@@ -58,6 +58,12 @@ StiffSpringForceField<DataTypes>::StiffSpringForceField(MechanicalState* object1
 template<class DataTypes>
 void StiffSpringForceField<DataTypes>::init()
 {
+    this->SpringForceField<DataTypes>::init();
+}
+
+template<class DataTypes>
+void StiffSpringForceField<DataTypes>::bwdInit()
+{
     if (f_indices1.isSet() && f_indices2.isSet() && (f_indices1.getValue().size() == f_indices2.getValue().size()))
     {
         helper::vector<Spring>& _springs = *this->springs.beginEdit();
--- a/SofaKernel/modules/SofaDeformable/StiffSpringForceField.inl
+++ b/SofaKernel/modules/SofaDeformable/StiffSpringForceField.inl
@@ -29,7 +29,7 @@
 #include <sofa/helper/AdvancedTimer.h>
 
 #include <sofa/core/visual/VisualParams.h>
-
+#include <SofaBaseTopology/TopologySubsetData.inl>
 namespace sofa
 {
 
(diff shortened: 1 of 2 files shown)


fredroy:
[ci-build][with-all-tests]

epernod:
- Add:
DataTracker (getValue)
or
DataCallBack (functor) 
instead of bwdInit()

- Change f_ to d_

SofaKernel/modules/SofaDeformable/StiffSpringForceField.inl
```diff
@@ -48,16 +48,72 @@ StiffSpringForceField<DataTypes>::StiffSpringForceField(double ks, double kd)
...
+        createSpringsFromInputs();
+    }
+
     this->SpringForceField<DataTypes>::init();
 }
 
+template<class DataTypes>
+void StiffSpringForceField<DataTypes>::updateSpringsIfChanged()
+{
+    if (!d_indices1.isSet() && !d_indices2.isSet()) // nothing to do in this case
```
hugtalbot, line 75: are you sure about the use of isSet() here ?
if the data is changed in any fancy way, isSet will be true ?

SofaKernel/modules/SofaDeformable/StiffSpringForceField.inl
```diff
@@ -48,16 +48,72 @@ StiffSpringForceField<DataTypes>::StiffSpringForceField(double ks, double kd)
...
+        return;
+
+    // force update
+    d_indices1.updateIfDirty();
+    d_indices2.updateIfDirty();
```
hugtalbot: I don't recall the process triggered by updateIfDirty .. can you light me up ?

epernod:
Hi @hugtalbot ,
we discussed it as well with @marques-bruno and @damienmarchal on gitter, we all have several way do to it. Can't say if one is best.
I want to put a generic mechanism in BaseObject (or another child/parent) that you could call, if you want, at init to link a dataTracker and which would automatically call an update (maybe not exactly the name update to not confuse with engines) virtual method. 

I suggest either we start a discussion issue on this and/or we add that discussion for the next dev meeting agenda.

epernod:
I removed the commit not linked to StiffSpringFF. I will add them in another PR

epernod:
@hugtalbot any update on the Data update callback?

SofaKernel/modules/SofaDeformable/StiffSpringForceField.inl
epernod: isSet here is just used to differentiate to the normal case where you use all points of your meshes and not only a subset identified by the indices

SofaKernel/modules/SofaDeformable/StiffSpringForceField.inl
epernod: this will force the update (if dirty),
means: 
1- it will check if this Data is dirty. This can be the case if a parent Data has been modified. In this case all children tree is set to dirty (but not yet updated)
2- Call the update method (if dirty == true), for a simple Data link it means copy new values from parent Data (starting from the first grand parent not dirty or being The elder).
3- Set this Data to clean (dirty == false)
4- (this I'm not 100% sure, would need to check the code) Increase the counter of your Data.

5- In the line below m_dataTrackerIndices.hasChanged() we check if all Data counter are still the same as the value stored by the dataTracker. if one is different, hasChanged return true.

epernod:
It won't compile here on the CI if you don't merge your PR in this one. Or I missed something.
I will test it on my computer.

epernod:
Here is what I have before taking your changes @hugtalbot 
This demo remove 1000 tetrahedra every second of simulation. 
[Alt text](https://youtu.be/7DfSFXvv2TU)

epernod:
[Alt text](https://youtu.be/Mif2UyuGyjE)

epernod:
[ci-build][with-all-tests]

PR #1095, opened 2019-07-01
[SofaExporter] Move bindings from SofaPython

fredroy:
SofaPython could not compile without having SofaExporter.
To solve that, bindings included in SofaPython are moved into SofaExporter.

Bindings should be included into their respective modules/plugins anyway.

This PR: 

### DIFF
--- a/applications/plugins/SofaPython/CMakeLists.txt
+++ b/applications/plugins/SofaPython/CMakeLists.txt
@@ -170,7 +170,6 @@ endif()
 
 add_library(${PROJECT_NAME} SHARED ${HEADER_FILES} ${SOURCE_FILES} ${PYTHON_FILES})
 
-target_compile_definitions(${PROJECT_NAME} PRIVATE  "${SOFAPYTHON_COMPILER_FLAGS}")
 target_compile_definitions(${PROJECT_NAME} PUBLIC "-DSOFA_HAVE_SOFAPYTHON")
 
 target_include_directories(${PROJECT_NAME}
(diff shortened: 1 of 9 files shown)


fredroy:
[ci-build][with-all-tests]

marques-bruno (approved):

applications/plugins/SofaPython/CMakeLists.txt
```diff
@@ -188,6 +187,7 @@ endif()
 target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${CMAKE_BINARY_DIR}/include>")
 target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/..>")
 target_include_directories(${PROJECT_NAME} PUBLIC "$<INSTALL_INTERFACE:include>")
+set_target_properties(${PROJECT_NAME} PROPERTIES COMPILE_FLAGS "${SOFAPYTHON_COMPILER_FLAGS}")
 set_target_properties(${PROJECT_NAME} PROPERTIES PUBLIC_HEADER "${HEADER_FILES}")
```
marques-bruno: I don't want to be a bummer but... @guparan  didn't we say we should remove all the PUBLIC_HEADER props?

guparan: Yup, they have to go. It could also be done everywhere in a future PR.

guparan (approved):

PR #1096, opened 2019-07-02
[scenes] Change OglModel to use a MeshObjLoader instead of loading the mesh internally.

epernod:
I just drop that here to remove warnings... (every thing is in the title)

This PR: 


epernod:
[ci-build][with-all-tests]

marques-bruno:
Shouldn't we remove the fileMesh field alltogether?

epernod:
by default use handleSeams="1" each time it was to be linked to OglModel
Until we have a good solution.

PR #1098, opened 2019-07-02
[Scenes] Clean some alias warnings

epernod:
Because I can imagine I'm not the only one to copy/past scenes and I don't like warnings!

Removed alias in scenes for: 
- DefaultPipeline, DefaultContactMapper DefaultCollisionGroupManager
- EulerImplicitSolver
- Point/Line/Triangle CollisionModel

This PR: 


epernod:
[ci-build][with-all-tests]

guparan (approved):

PR #1107, opened 2019-07-19
[SofaGraphComponent] Update sceneCheckerAPI and deprecate MatrixMass

hugtalbot:
This PR:
- updates the sceneCheckerAPI by setting the default leve at 19.06 and removing the warning for BoxStiffSpringFF
- set as deprecated MatrixMass for developers
- warns users about the deprecation

This PR: 


epernod:
and BoxStiffSpringFF is not deprecated anymore?

PR #1110, opened 2019-07-25
[Carving plugin] Small fix at init.

epernod:
- Fix: init send error when good collision models are not found instead of randomly picking the first one found in the graph.
- Clean model check in doCarve and fix advancedTimer instruction

This PR: 


hugtalbot (approved):

guparan:
[ci-build][with-scene-tests]

guparan:
The scene CarvingTool.scn throws an error: `[CarvingManager(carvingManager1)] m_surfaceCollisionModels not found. Set tag 'CarvingSurface' to the right collision models.`
Could you have a look @epernod ?

applications/plugins/SofaCarving/CarvingManager.cpp
```diff
@@ -91,11 +87,6 @@ void CarvingManager::init()
         std::vector<core::CollisionModel*> models;
         getContext()->get<core::CollisionModel>(&models, core::objectmodel::Tag("CarvingSurface"), core::objectmodel::BaseContext::SearchRoot);
     
-        // extend the research to model without the tag. 
-        if (models.empty())
```
hugtalbot, line 95: Then there is no check anymore if models is empty or not

applications/plugins/SofaCarving/CarvingManager.cpp
```diff
@@ -76,11 +76,7 @@ void CarvingManager::init()
 {
     // Search for collision model corresponding to the tool.
     if (d_toolModelPath.getValue().empty())
-    {
         m_toolCollisionModel = getContext()->get<core::CollisionModel>(core::objectmodel::Tag("CarvingTool"), core::objectmodel::BaseContext::SearchDown);
```
marques-bruno, line 80: I didn't know that it was possible to retrieve objects in ctx with tags. is it possible to get rid of the getContexts here?

applications/plugins/SofaCarving/CarvingManager.cpp
epernod: No because, if there is no models, this means no collision models with the tag CarvingSurface have been found. The user need to set that tag to insure a correct behavior of the CarvingManager.
He will be informed by: 
> error: [CarvingManager(carvingManager1)] m_surfaceCollisionModels not found. Set tag 'CarvingSurface' to the right collision models.

The old behavior I removed was to pick up the first randomly collision model encountered in the graph.
```
if (models.empty())
getContext()->get<core::CollisionModel>(&models, core::objectmodel::BaseContext::SearchRoot);
```

applications/plugins/SofaCarving/CarvingManager.cpp
epernod: no idea.

epernod:
[ci-build][with-all-tests]

PR #1115, opened 2019-07-27
[SofaConstraint] Fix UncoupledConstraintCorrection topology change handling

epernod:
In UncoupledConstraintCorrection class:
- Use PointData instead of handleTopologyChange method
- Small cleaning
- Remove duplicated constructor

Fix one of the component in #810 

This PR: 


epernod:
[ci-build][with-all-tests]

PR #1118, opened 2019-08-01
[SofaKernel] ADD: static method in events to retrieve the classname

marques-bruno:
This PR changes the API of the Event class slightly so that there's no need to instantiate an Event to get its classname.
The getClassName method that was previously implemented by each of Event's subclasses is moved in the SOFA_EVENT_H macro, and a static GetClassName method becomes the one to override.

This is important as Events do not have a default ctor and it is thus impossible to instantiate a dummy event to check its type without knowing the signature of the subclass's ctor.

This allows me to implement better and cleaner bindings for events in SofaPython3.

This PR: 


marques-bruno:

tgaugry (approved):

PR #1119, opened 2019-08-02
[SofaBoundaryCondition] Fix FixedRotationConstraint when using more than one locked axis

tgaugry:
Current implementation has weird behavior when more than one axis is locked. Current implementation may make the objects rotate 180° when they should not.

Current PR aims to prevent this issue using swing-twist decomposition algorithm.
As a bonus, code is now shorter.

Everything was done inside the class, but the swing twist decomposition could be move to the main quaternion class.

This PR: 

### DIFF
--- a/modules/SofaBoundaryCondition/FixedRotationConstraint.inl
+++ b/modules/SofaBoundaryCondition/FixedRotationConstraint.inl
@@ -120,7 +120,7 @@ void FixedRotationConstraint<DataTypes>::projectPosition(const core::MechanicalP
         sofa::defaulttype::Quat Qp_remaining = Q_prev;
         sofa::defaulttype::Quat to_keep = sofa::defaulttype::Quat::identity();
 
-        auto remove_rotation = [&Q_remaining, &Qp_remaining, &to_keep](const Vec3 axis) {
+        auto remove_rotation = [&](const Vec3 axis) {
             Q_remaining = decompose_ts(Q_remaining, axis).second;
             sofa::defaulttype::Quat twist;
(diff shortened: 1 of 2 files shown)


modules/SofaBoundaryCondition/FixedRotationConstraint.inl
```diff
@@ -91,95 +91,52 @@ template <class DataTypes>
...
+            Vec3 projected = project(vec3_part, twistAxis);
+            sofa::defaulttype::Quat twist(projected[0], projected[1], projected[2], q[3]);
+            // Singularity : A perpendicular angle would give you quaternion (0, 0, 0, 0)
+            if(std::none_of(twist.ptr(), twist.ptr() + 4 * sizeof(double), [](double x) {return x != 0. ;})) {
+                twist = sofa::defaulttype::Quat::identity();
+            }
+            twist.normalize();
+            sofa::defaulttype::Quat swing = q * twist.inverse();
+            swing.normalize();
+            return std::make_pair(twist, swing);
```
marques-bruno, line 115: Since you're using so much c++11/14/17 awesomeness in this PR I'd suggest replacing the std::make_pair(twist, swing) with std::pair(twist, swing) :)

Can't really judge the algo, although it seems like a good refactoring to me so +1

marques-bruno (approved):

modules/SofaBoundaryCondition/FixedRotationConstraint.inl
tgaugry: Actually, I'm c++11 compliant ^^ What you are proposing will only work with c++17 deduction, right ?

PR #1126, opened 2019-08-21
[SofaTopologyMapping] Cleanups of some topological mappings + better initialization

jnbrunet:
Various cleanup and better checks at the init stage to make sure the topological mappings have everything they need.

Only Hexa2Quad, Quad2Triangle and Tetra2Triangle for now.

This PR: 

### DIFF
--- a/modules/SofaTopologyMapping/Hexa2QuadTopologicalMapping.cpp
+++ b/modules/SofaTopologyMapping/Hexa2QuadTopologicalMapping.cpp
@@ -169,7 +169,7 @@ void Hexa2QuadTopologicalMapping::updateTopologicalMappingTopDown()
     auto itBegin=fromModel->beginChange();
     auto itEnd=fromModel->endChange();
 
-    auto & Loc2GlobVec = *(Loc2GlobDataVec.beginEdit());
+    Topology::SetIndices & Loc2GlobVec = *(Loc2GlobDataVec.beginEdit());
 
     while( itBegin != itEnd )
     {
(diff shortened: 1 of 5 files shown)


modules/SofaTopologyMapping/Hexa2QuadTopologicalMapping.cpp
```diff
@@ -101,10 +106,8 @@ void Hexa2QuadTopologicalMapping::init()
         return;
     }
 
-    QuadSetTopologyContainer *to_tstc;
-    toModel->getContext()->get(to_tstc);
```
epernod, line 105: could toModel be different than to_tstc ?
I think you are right, but I don't remember why this get() was here.

modules/SofaTopologyMapping/Quad2TriangleTopologicalMapping.cpp
```diff
@@ -267,14 +272,14 @@ void Quad2TriangleTopologicalMapping::updateTopologicalMappingTopDown()
         }
         case core::topology::QUADSREMOVED:
         {
-            const sofa::helper::vector<unsigned int> &tab = ( static_cast< const QuadsRemoved *>( *itBegin ) )->getArray();
+            const auto & tab = ( static_cast< const QuadsRemoved *>( *itBegin ) )->getArray();
```
epernod: I have nothing against using auto type, but in those cases maybe changing the name of the variable would be good.
Because ```auto tab ``` gives no info of what we are manipulating.  maybe tabIndices or what you think is best.

modules/SofaTopologyMapping/Hexa2QuadTopologicalMapping.cpp
jnbrunet: Yes, toModel is a BaseMeshTopology, so it could be something else than a QuadSetTopologyContainer. I'm not sure if casting it to a QuadSetTopologyContainer is really necessary here... but at least this PR keeps the "toModel must be a QuadSetTopologyContainer" requirement.

modules/SofaTopologyMapping/Quad2TriangleTopologicalMapping.cpp
jnbrunet: I went to the file very quickly changing to auto where my IDE would suggest to do so (and a bit more), I agree that the name isn't very well chosen. I will try if I have some time to change the names... but here the idea of auto is more that it will automatically handle the case where the return value type changes in the future.

jnbrunet:
[ci-build][with-all-tests]

jnbrunet:
Thanks a lot @epernod !!

PR #1127, opened 2019-08-21
[SofaKernel] Add a strict option to the BoxROI to prevent partially inside element to be in the box.

jnbrunet:
Right now, if you put a BoxROI in your scene to get the edges, triangles, etc. contained inside a box, the BoxROI will gets you every elements for which its center point is inside the box. An element partially inside the box can than be flagged as inside.

This PR adds an option named "strict" to the BoxROI which will instead test that every nodes of the elements are contained inside the box. The option is false by default so that the current behavior remains unchanged.

This PR: 


epernod:
hum... weird, I never checked it was the center of Mass that was used in this engine.
I would expect the default behavior to be the one you just added as strict behavior!

jnbrunet:
> hum... weird, I never checked it was the center of Mass that was used in this engine.
> I would expect the default behavior to be the one you just added as strict behavior!

Yeah...and it was the cause of a big headache of mine when running some simulations!

SofaKernel/modules/SofaEngine/BoxROI.h
```diff
@@ -205,10 +206,15 @@ class BoxROI : public DataEngine
...
     bool isEdgeInBoxes(const Edge& e);
+    bool isEdgeInBoxesStrict(const Edge& e);
     bool isTriangleInBoxes(const Triangle& t);
+    bool isTriangleInBoxesStrict(const Triangle& t);
     bool isTetrahedronInBoxes(const Tetra& t);
+    bool isTetrahedronInBoxesStrict(const Tetra& t);
     bool isHexahedronInBoxes(const Hexa& t);
+    bool isHexahedronInBoxesStrict(const Hexa& t);
     bool isQuadInBoxes(const Quad& q);
+    bool isQuadInBoxesStrict(const Quad& q);
```
hugtalbot, line 217: Could not the check of strict option just be done in the existing function to simplify ?

jnbrunet: The compiler won't be able to optimize it if you do this, so you will have a condition to evaluate at each calls, quite slow if the branch predictions fails a lot.

hugtalbot (approved):

SofaKernel/modules/SofaEngine/BoxROI.h
```diff
@@ -147,7 +147,7 @@ class BoxROI : public DataEngine
     Data<bool> d_computeTetrahedra; ///< If true, will compute tetrahedra list and index list inside the ROI. (default = true)
     Data<bool> d_computeHexahedra; ///< If true, will compute hexahedra list and index list inside the ROI. (default = true)
     Data<bool> d_computeQuad; ///< If true, will compute quad list and index list inside the ROI. (default = true)
-    Data<bool> d_strict; ///< If true, an element is inside the box iif all of its nodes are inside. If False, only the center point of the element is checked. (default = false)
```
jnbrunet: That was a "if and only if" statement :-)

hugtalbot: Ah ok : iff ^^
sorry I did not know this shortening

jnbrunet: Me neither haha, it was a typo ;)

epernod:
I would be to have only the strict condition in BoxRoi and remove the old behavior.
It doesn't make sens. Otherwise you jsut need to work with the points and not the topology elements.
Could we discuss that in next meeting?

PR #1129, opened 2019-08-22
[SofaCore][ExecParams] Remove thread specific declaration

fspadoni:
Remove thread specific declaration for ExecParams class.

accessing thread local variables is slow and if the thread-local variable is accessed very frequently, the cost may become an issue. 
https://software.intel.com/en-us/blogs/2011/05/02/the-hidden-performance-cost-of-accessing-thread-local-variables

The ExecParams global pointer is declared thread specific but it is often passed in a function call as argument.
A function executed in a second thread can receive as argument the thread local variable of the first thread and it is allowed to use it.
In this situation there is no need of use the thread local variable.

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/core/ExecParams.cpp
+++ b/SofaKernel/framework/sofa/core/ExecParams.cpp
@@ -41,7 +41,7 @@ ExecParams::ExecParamsThreadStorage::ExecParamsThreadStorage(int tid)
 /// Get the default ExecParams, to be used to provide a default values for method parameters
 ExecParams* ExecParams::defaultInstance()
 {
-    SOFA_THREAD_SPECIFIC_PTR(ExecParams, threadParams);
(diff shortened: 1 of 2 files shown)


fspadoni:
@JeremieA 
Hi Jeremie,  as you might know, I am currently working on multithreaded simulation with SOFA.
In this PR, I propose to remove  the checkValidStorage() test in Node::ExecuteVisitor function because it prevents some parallel execution and also suggest to remove the thread specific declaration for ExecParams variable.
Do you see a reason to keep the ExecParams specific for each thread?
I especially noticed accessing the thread local memory of a thread A from the thread B with these different ExecParams is significantly slow (~20%) .  
Your opinion would be valuable for us. Thanks!

JeremieA:
@fspadoni I answered in #1147 
Best,
Jeremie.

PR #1130, opened 2019-08-28
[SofaViewer] Prevent the GUI to ouput every CTRL actions in the console

hugtalbot:
Significant improvement.
Just removing msg coming in the console in release due to CTRL key pressed.

This PR: 


guparan:
Wow!

PR #1139, opened 2019-09-06
[SofaDefaulttype] FIX too many ExtVec warnings

guparan:
Deprecation warnings were displayed at every `#include <sofa/defaulttype/VecTypes.h>`
They are now emitted only when using the deprecated types ExtVec*

@fredroy Is there a reason you did not use inherited structs in the first place?

Also, could someone explain why putting the `[[deprecated]]` attribute on a `using` alias does not emit a warning only when the alias is used?

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/defaulttype/VecTypes.h
+++ b/SofaKernel/framework/sofa/defaulttype/VecTypes.h
@@ -196,9 +196,9 @@ public:
     }
 };
 
-template <class TCoord, class TDeriv, class TReal> using ExtVectorTypes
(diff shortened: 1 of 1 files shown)


fredroy:
Because it is creating a new type, and we will break instancing templated class like Mapping and so on. If you do MyMapping<Vec3, ExtVec3> it will complain that there is no BaseMapping<Vec3, ExtVec3> and so on. (basically).
I remember asking people in a weekly meeting, and we agreed it was better to do a typedef instead of an inheritance.

guparan:
Indeed, this in not a good solution to our problem. Thanks for the explanation, I close this for now!

PR #1146, opened 2019-09-17
[SofaKernel] fix root's getPathName

marques-bruno:
This PR proposes an alternative representation of the root path's name.

If we consider a node named "root", with a child node named "child", a component named "object" and a data field named "data", here were the previous values returned by their `getPathName` method:

```
root.getPathName(): ""
root.child.getPathName(): "/child"
root.object.getPathName(): "/object"
root.data.getLinkPath(): ".data"
```

I find this naming convention misleading, as an empty string for the root path does not seem right to me, and ".data" for the path of root.data also look wierd:

- When working with link paths, like we do all the time in Sofa, making a link between a component's data and root.value would look like this:
```
root.createObject('MyComponent', value="@.value")
```

- The same way, if a component for some reason would need a take a node as an input field, passing root would look like:

```
node.createObject("MyComponent", nodePath="@") # passing root to nodePath would lead to syntax errors during parsing.
```

Thus I suggest that, as it is done in FTP clients for instance, or filesystem browsers in general, the root node, regardless of its name be represented as a forward slash `/`:

```
root.getPathName(): "/"
root.child.getPathName(): "/child"
root.object.getPathName(): "/object"
root.data.getLinkPath(): "@/.data" # this cas still looks a bit funny, but makes more sense IMHO

root.createObject('MyComponent', value="@/.value")
root.createObject('MyComponent', nodePath="@/")
```

This PR: 

### DIFF
--- a/SofaKernel/framework/sofa/core/objectmodel/BaseNode.cpp
+++ b/SofaKernel/framework/sofa/core/objectmodel/BaseNode.cpp
@@ -81,19 +81,26 @@ void BaseNode::clearObjectContext(BaseObject::SPtr obj)
         obj->l_context.reset();
 }
 
-std::string BaseNode::getPathName() const {
+std::string BaseNode::_getPathName() const {
     std::string str;
     Parents parents = getParents();
     if (!parents.empty())
     {
         // for the full path name, we arbitrarily take the first parent of the list...
         // no smarter choice without breaking the "Node" heritage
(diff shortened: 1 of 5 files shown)


marques-bruno:
[ci-build]

damienmarchal (approved):
I would have prefer to get rid of the first / as 'root' but given that it would break every scene. I think the current PR is the best compromise.

PR #1149, opened 2019-09-19
[SofaBaseMechanics] Use doUpdateInternal API in UniformMass

hugtalbot:
Following the ConstantForceField let's apply the updateInternal API to UniformMass

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaBaseMechanics/UniformMass.inl
+++ b/SofaKernel/modules/SofaBaseMechanics/UniformMass.inl
@@ -142,6 +142,8 @@ void UniformMass<DataTypes, MassType>::init()
 template <class DataTypes, class MassType>
 void UniformMass<DataTypes, MassType>::initDefaultImpl()
 {
+    this->m_componentstate = core::objectmodel::ComponentState::Valid;
+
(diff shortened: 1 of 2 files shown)


SofaKernel/modules/SofaBaseMechanics/UniformMass.inl
```diff
@@ -366,10 +361,10 @@ void UniformMass<DataTypes, MassType>::handleTopologyChange()
     {
         indices.clear();
         for(size_t i=0; i<mstate->getSize(); i++)
-            indices.push_back((int)i);
+            indices.push_back(int(i));
```
epernod: maybe indices should be a vector<unsigned int> instead of int, would avoid all the conversions

epernod, line 364: maybe indices should be a vector instead of int, would avoid all the conversions

hugtalbot: yes but this was not the purpose of the PR

epernod (approved):

PR #1150, opened 2019-09-20
[SofaBaseMechanics] Use doUpdateInternal API in DiagonalMass

hugtalbot:
Following other components let's apply the updateInternal API to DiagonalMass

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaBaseMechanics/DiagonalMass.inl
+++ b/SofaKernel/modules/SofaBaseMechanics/DiagonalMass.inl
@@ -657,6 +657,8 @@ void DiagonalMass<DataTypes, MassType>::init()
 {
     m_componentstate = ComponentState::Valid;
 
+    this->f_listening.setValue(false);
+
     if (!d_fileMass.getValue().empty())
     {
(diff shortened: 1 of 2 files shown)


SofaKernel/modules/SofaBaseMechanics/DiagonalMass.inl
```diff
@@ -809,21 +808,21 @@ void DiagonalMass<DataTypes, MassType>::computeMass()
...
             masses.resize(this->mstate->getSize());
             for(unsigned int i=0; i<masses.size(); ++i)
-              masses[i]=(Real)0;
+              masses[i]=Real(0);
 
             Real md=d_massDensity.getValue();
-            Real mass=(Real)0;
-            Real total_mass=(Real)0;
+            Real mass=Real(0);
+            Real total_mass=Real(0);
```
epernod, line 815: do you really need all those casts ?

hugtalbot: No idea, this change was just made to fix the warning.
Do you think this could really slow down things ?

PR #1151, opened 2019-09-20
[SofaMiscForceField] Use doUpdateInternal API in MeshMatrixMass

hugtalbot:
Finally apply the doUpdateInternal to MeshMatrixMass
This class should definitely be in SofaBaseMechanics ...

This PR: 


epernod (approved):

epernod:
there are 2 more unit tests failing on windows only, could you check if relevant to this PR or not. thx

PR #1157, opened 2019-09-26
[SofaHaptics] Add mutex and option to lock the ForceFeedback computation

epernod:
Add abstract method in ForceFeedBack and override in LCPForceFeedback which will lock/unlock a mutex used in doComputeForce.

This allow to lock the ForceFeedback computation, done in a dedicated working thread, from another component which is changing the mechanical system in the main thread of SOFA.

This PR: 

### DIFF
--- a/modules/SofaHaptics/src/SofaHaptics/LCPForceFeedback.inl
+++ b/modules/SofaHaptics/src/SofaHaptics/LCPForceFeedback.inl
@@ -181,6 +181,12 @@ void LCPForceFeedback<DataTypes>::init()
     }
 }
 
+template <class DataTypes>
+void LCPForceFeedback<DataTypes>::setLock(bool value)
+{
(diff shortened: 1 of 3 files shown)


modules/SofaHaptics/src/SofaHaptics/LCPForceFeedback.inl
```diff
@@ -191,8 +198,11 @@ void LCPForceFeedback<DataTypes>::computeForce(const VecCoord& state,  VecDeriv&
         return;
     }
     updateStats();
+    
+    lockForce.lock(); // check if computation has not been locked using setLock method.
```
hugtalbot, line 202: isn't it used uninitialized ?

epernod: it is a mutex, no need to initialize: https://fr.cppreference.com/w/cpp/thread/mutex

PR #1160, opened 2019-09-26
[SofaBaseTopology] Add new geometric methods in TetrahedronSetGeometryAlgorythms

epernod:
```cpp
/// Method to check if points stored inside the Tetrahedron, given by the tetrahedron id, are in the right order (by checking the cross products between edges).
bool checkNodeSequence(const TetraID tetraId) const;

/// Method to check if the dihedral angles of the tetrahedron have correct values (between 20 and 160 degrees).
bool checkTetrahedronDihedralAngles(const TetraID tetraId) const;

/// Method to check if Tetrahedron is elongated, meaning the longest edge > 10x min edge
bool isTetrahedronElongated(const TetraID tetraId) const;

/// Return false if one of the test method: @sa isTetrahedronElongated, @sa checkTetrahedronDihedralAngles and @sa checkNodeSequence return false for the given Tetrahedron Id.
bool checkTetrahedronValidity(const TetraID tetraId) const;

/// Will call @sa checkTetrahedronValidity for each Tetrahedron of the mesh and store the bad tetrahedron ID in @sa m_badTetraIds
const sofa::helper::vector <TetraID>& computeBadTetrahedron();
```

This PR: 


epernod:
the results are empire star destroyers picture ;)

epernod:

PR #1175, opened 2019-10-08
[CMake] FIX Eigen finding

guparan:
- FIX unkown command sofa_add_subdirectory_external
- FIX EIGEN3_ROOT priority over system Eigen
- CLEAN naming (Eigen -> Eigen3)

This PR: 


jnbrunet:
Within recent version of Eigen installed on most OS, there is already a cmake find_package(Eigen3) available. By create a FindEigen3.cmake you will override this. I've seen a couple of projects dealing with this kind of problem with, for example:

 SofaKernel/cmake/Modules/**FindEigen3Sofa**.cmake:
```
# try to find "FindEigen3.cmake" from system Eigen
find_package(Eigen3) 

if (not Eigen3)
# do your thing here
else()
# use system Eigen3
endif()
```

CMakeLists.txt
`find_package(Eigen3Sofa)`

pybind11 do this for example with find_package(PythonNew) which point to their PythonNew.cmake which trigger a system find_package(Python).

Maybe it could be a solution for us?

Edit: Sorry I'm polluting this fast-merge PR for that, we can discuss this elsewhere ;-D

guparan:
Thanks for the remark @jnbrunet, I will look at this in a new PR :+1:

PR #1176, opened 2019-10-09
[SofaSimulationGraph] Stop DAGNode get parent topology process in BarycentricMapping

adagolodjo:
Stop the search of the Container/BaseTopology when one is found (using break)

This PR: 


adagolodjo:
Merci @hugtalbot, j'avais oublié de mettre "to review".

PR #1186, opened 2019-10-14
[SofaSphFluid] Several clean in the plugin components

epernod:
Only some cleaning in:
- Class define macro
- Data and member prefix
- Move code from header to inl or cpp files

Update several plugin demo scenes.

SpatialGridContainer_00000001
SPHFluidForceField_benchmarks_00000001
SPHFluidSurfaceMapping_00000001

This PR: 


epernod:
@guparan @hugtalbot could you check this plugin is compiled on the CI, I have a doubt

guparan:
Done, let's [ci-build][with-all-tests] again :+1:

PR #1192, opened 2019-10-18
[CMake] FIX unknown compiler option on VS2015

guparan:
I know we are planning to drop VS2015 support in a very near future (I'm actually working on it) but this "unknown option" warning bothers me at each compilation (yes, I'm still under VS2015 myself).

This PR: 

### DIFF
--- a/SofaKernel/cmake/CompilerOptions.cmake
+++ b/SofaKernel/cmake/CompilerOptions.cmake
@@ -36,7 +36,10 @@ endif()
 ## Windows-specific
 if(WIN32)
     add_definitions("-wd4250 -wd4251 -wd4275 -wd4675 -wd4996 -D_USE_MATH_DEFINES")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP /Zc:__cplusplus")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+    if(MSVC_TOOLSET_VERSION GREATER 140) # > VS 2015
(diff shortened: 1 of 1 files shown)


SofaKernel/cmake/CompilerOptions.cmake
```diff
@@ -36,7 +36,10 @@ endif()
 ## Windows-specific
 if(WIN32)
     add_definitions("-wd4250 -wd4251 -wd4275 -wd4675 -wd4996 -D_USE_MATH_DEFINES")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP /Zc:__cplusplus")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /MP")
+    if(MSVC_TOOLSET_VERSION GREATER 140) # > VS 2015
```
epernod, line 40: I'm quite sure I saw that warning as welll, and I'm on 2017

fredroy:
It was supposed to be supported from VS2015: https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus?view=vs-2015 . But according to https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/ , it seems to be enabled only from VS2017 Update 15.7 .

For now, this option is not needed while the C++ standard is still c++11. 
And the moment we enable c++17 will be the same time we will abandon VS2015, so everything should be OK (in theory)

fredroy (approved):

SofaKernel/cmake/CompilerOptions.cmake
fredroy: Which version of VS2017 ?

PR #1198, opened 2019-10-28
[SofaGui] FIX missing find_package in SofaGuiConfig.cmake.in

AlbanOdot:
find_package(SofaGui) was failing because of SofaGuiCommon's dependency on SofaComponentAll

This PR: 


guparan:
Congrats @AlbanOdot for your first PR :tada: 

The fix is OK but I was wondering: does SofaGuiCommon really need SofaComponentAll (a.k.a. all SOFA modules)? I'll take a look at that.

PR #1212, opened 2019-11-05
[SofaBaseMechanics] Make Uniform and DiagonalMass compatible with topo change

hugtalbot:
This (old) PR allows to have topological change with UniformMass and DiagonalMass.

PointAdded and PointRemoved are added.
Add clean tracker function (temporary regarding Data discussion)

This should fix the testIncise.

This PR: 


epernod:
[ci-build][with-all-tests]

SofaKernel/modules/SofaBaseMechanics/DiagonalMass.inl
```diff
@@ -224,15 +313,37 @@ void DiagonalMass<DataTypes,MassType>::DMassPointHandler::ApplyTopologyChange(co
     const sofa::helper::vector< sofa::helper::vector< unsigned int > >& ancestors = e->ancestorsList;
     const sofa::helper::vector< sofa::helper::vector< double > >& coefs = e->coefs;
 
-    applyTriangleCreation(triangleAdded,elems,ancestors,coefs);
+    if(dm->triangleGeo)
```
epernod, line 316: To avoid all those Geometry component check, I would put that test in the init method when the callback are created. Thus if for example you are using the mass on the Triangulation, TetraGeo and hexaGeo would be nullptr and so you don't create the call back for this level of topology.

SofaKernel/modules/SofaBaseMechanics/DiagonalMass.inl
```diff
@@ -597,6 +763,10 @@ bool DiagonalMass<DataTypes, MassType>::checkTopology()
             else
             {
                 msg_info() << "Hexahedral topology found.";
+                edgeGeo = nullptr;
+                triangleGeo = nullptr;
+                quadGeo = nullptr;
+                tetraGeo = nullptr;
```
epernod, line 769: are they not all nullptr by default?

epernod:
I have tested both mass on a triangulation and tetra mesh, seems to work. Let's see the CI.

epernod (approved):

PR #1215, opened 2019-11-14
[SofaSphFluid] Update rendering & other

fredroy:
* Add blending, therefore keeps background and other opaque models (but rendering is still better with a whitish background)
* Add scene with collisions
* Add an example of the execution policy (std::execution::par) introduced by c++17 in the brute-force findNeighbors() in the forcefield (faster than the spatialgrid if nbParticles is approx less than 5K)

Blending should be temporary, and having a multi pass would be better (refraction, etc)

This PR: 

### DIFF
--- a/applications/plugins/SofaSphFluid/examples/SPHParticleSink_obstacle.scn
+++ b/applications/plugins/SofaSphFluid/examples/SPHParticleSink_obstacle.scn
@@ -1,0 +1,76 @@
+<?xml version="1.0" ?>
+<Node dt="0.01" gravity="0 -10 0" bbox="-4 -4 -4 4 4 4">
+    <RequiredPlugin name="SofaOpenglVisual"/> 
+    <RequiredPlugin name='SofaSphFluid' />
+    <VisualStyle displayFlags="hideBehaviorModels hideForceFields hideWireframe" />
+
+    <DefaultPipeline depth="15" verbose="0" draw="0" />
+    <BruteForceDetection name="N2" />
+    <NewProximityIntersection name="Proximity" alarmDistance="1.0" contactDistance="0.5" />
+    <DefaultContactManager name="Response" response="default" /> 
+
+    <Node name="Particles">
+        <EulerSolver symplectic="1" />
+        <MechanicalObject name="MModel" />
+        <ParticleSource name="Source" translation="0 10 0" radius="0.01 0.1 0.01" velocity="0 -10 0" delay="0.02" start="-0.1" stop="4" printLog="0"
+        center="-0.375 0 -0.75 
+            0.0 0.0 -0.75 
+            0.375 0.0 -0.75 
+            -0.75  0.0 -0.375 
+            -0.375 0.0 -0.375 
+            0.0 0.0 -0.375 
+            0.375 0.0 -0.375 
+            0.75 0.0 -0.375 
+            -0.75 0.0 0.0 
+            -0.375 0.0 0.0 
+            0.0 0.0 0.0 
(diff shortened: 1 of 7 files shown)


applications/plugins/SofaSphFluid/examples/SPHParticleSink_obstacle.scn
```diff
@@ -0,0 +1,76 @@
...
+        <SpatialGridContainer cellWidth="0.75" />
+        <SPHFluidForceField radius="0.7" density="25" kernelType="1" viscosityType="2" viscosity="10" pressure="1000" surfaceTension="-1000" printLog="0" />
+
+        <!--
+        <PlaneForceField normal="1 0 0" d="-4" showPlane="1" />        
+        <PlaneForceField normal="0.5 1 0.1" d="-4" showPlane="1" />
+        <PlaneForceField normal="0 0 1" d="-4" showPlane="1" />
+        <PlaneForceField normal="0 0 -1" d="-4" showPlane="1" />
+        
+        <BoxROI name="ROI1" box="10 -20 -20 20 20 20" drawBoxes="1" drawPoints="1" position="@MModel.position"/>-->
```
epernod: remove commented line?

applications/plugins/SofaSphFluid/src/SofaSphFluid/SPHFluidForceField.inl
```diff
@@ -164,13 +170,30 @@ void SPHFluidForceField<DataTypes>::computeNeighbors(const core::MechanicalParam
...
+                Real r2 = (rj- ri).norm2();
+                if (r2 < h2)
+                {
+                    Real r_h = (Real)sqrt(r2/h2);
+                    m_particles[i].neighbors.push_back(std::make_pair(j,r_h));
+                }
+            }
+
+        });
+#endif
```
epernod, line 190: How faster is it?

epernod (approved):

applications/plugins/SofaSphFluid/examples/SPHParticleSink_obstacle.scn
fredroy: I copied-paste from SPHParticleSink.scn :p , so I will remove from both of them then

applications/plugins/SofaSphFluid/src/SofaSphFluid/SPHFluidForceField.inl
fredroy: 25% or so ? But this is not really important as you would want to use the spatial grid anyway I guess.

fredroy:
[ci-build][with-all-tests]

applications/plugins/SofaSphFluid/src/SofaSphFluid/SPHFluidForceField.inl
```diff
@@ -164,13 +170,30 @@ void SPHFluidForceField<DataTypes>::computeNeighbors(const core::MechanicalParam
                 }
             }
         }
+#else
+        std::for_each(std::execution::par, x.begin(), x.end(), [&](const auto& ri)
+        {
+            int i = &ri - &x[0]; // only possible with vector, etc.
```
guparan, line 176: use `std::distance` here?

fredroy: nope, you need iterators for std::distance(), and for_each gives you the value directly ("ri" here is not an iterator)

guparan (approved):

PR #1224, opened 2019-11-25
[SofaCore] Making BaseData::setParent() virtual

marques-bruno:
setParent is called on Datas when parsing BaseObjectDescriptions on BaseObjects.
The default behavior is incompatible with the Data-based API we want to substitute to the current Link system between components.

I thus want to be able to override the default setParent() behavior in my ObjectLink POC (see [here](https://github.com/SofaDefrost/ComponentData_POC/pull/1) for draft implementation)

Please, don't hesitate to comment on the PR on the POC repository's pull request mentioned above

This PR: 


marques-bruno:
not necessary anymore.. see #1226

PR #1226, opened 2019-11-29
[SofaCore] Add DDGLinks

marques-bruno:
Hey!

First PR with a POC of Object links that aren't based on SingleLinks

Here, a new template class (with its base class) is introduced: DDGLink.
Just like BaseDatas, DDGLink inherits DDGNode.
A DDGLink thus benefits from the whole dirty flag propagation thingy, and implements the update() function to clean its dirty flag, and update from its inputs.
Just like Datas, a DDGLink also possesses a counter, that is incremented when the link value changes.

Just like BaseLinks, and SPtrs, to retrieve the object linked with a DDGLink, you need to call the get() method. The set() method will set the link.

When a link is set within a DDGLink, the componentState of the linked object is set as input to that DDGLink. It's up to you to connect your DDGLink to some datas to connect your dependency graph.

Contrary to the current link implementation in SOFA, these links, can be very easily manipulated as their base class (BaseDDGLink), or as DDGNodes even.

DDGLinks are parsed in Base::parseField just like BaseLinks and BaseDatas, which makes them usable with the BaseObjectDescriptions. (they really mirror the behavior of all other field types.)

A very simple test case is present in this PR to show you how it's used.

You can reference your links with their pathnames, just like normal datas or links, by calling the getPathName() method

There's not bi-directionality implemented in this PR, (because I didn't have time in the end...) but it's not rocket science:
- all there is to do is add a vector of Base* in Base with helper methods to add, remove retrieve the list of link owners. Then in DDGLink::set(), call the add() method of your link to register`this`. (maybe in another PR...) 

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
+++ b/SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
@@ -371,7 +371,7 @@ std::vector< BaseData* > Base::findGlobalField( const std::string &name ) const
     std::vector<BaseData*> result;
     //Search in the aliases
     auto range = m_aliasData.equal_range(name);
-    for (mapIterator itAlias=range.first; itAlias!=range.second; ++itAlias)
+    for (auto itAlias=range.first; itAlias!=range.second; ++itAlias)
         result.push_back(itAlias->second);
     return result;
 }
@@ -382,7 +382,7 @@ std::vector< BaseDDGLink* > Base::findGlobalDDGLink( const std::string &name ) c
     std::vector<BaseDDGLink*> result;
     //Search in the aliases
     auto range = m_aliasDDGLink.equal_range(name);
-    for (mapIterator itAlias=range.first; itAlias!=range.second; ++itAlias)
+    for (auto itAlias=range.first; itAlias!=range.second; ++itAlias)
         result.push_back(itAlias->second);
     return result;
 }
@@ -418,7 +418,7 @@ std::vector< BaseLink* > Base::findLinks( const std::string &name ) const
     std::vector<BaseLink*> result;
     //Search in the aliases
     auto range = m_aliasLink.equal_range(name);
-    for (mapIterator itAlias=range.first; itAlias!=range.second; ++itAlias)
+    for (auto itAlias=range.first; itAlias!=range.second; ++itAlias)
         result.push_back(itAlias->second);
     return result;
 }
--- a/SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
+++ b/SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
@@ -339,16 +339,14 @@ void Base::removeTag(Tag t)
 void Base::removeData(BaseData* d)
 {
     m_vecData.erase(std::find(m_vecData.begin(), m_vecData.end(), d));
-    typedef MapData::const_iterator mapIterator;
-    std::pair< mapIterator, mapIterator> range = m_aliasData.equal_range(d->getName());
+    auto range = m_aliasData.equal_range(d->getName());
     m_aliasData.erase(range.first, range.second);
 }
 
 void Base::removeDDGLink(BaseDDGLink* d)
 {
     m_vecDDGLink.erase(std::find(m_vecDDGLink.begin(), m_vecDDGLink.end(), d));
-    typedef MapDDGLink::const_iterator mapIterator;
-    std::pair< mapIterator, mapIterator> range = m_aliasDDGLink.equal_range(d->getName());
+    auto range = m_aliasDDGLink.equal_range(d->getName());
     m_aliasDDGLink.erase(range.first, range.second);
 }
 /// Find a data field given its name.
@@ -358,8 +356,7 @@ BaseData* Base::findData( const std::string &name ) const
     //Search in the aliases
     if(m_aliasData.size())
     {
-        typedef MapData::const_iterator mapIterator;
-        std::pair< mapIterator, mapIterator> range = m_aliasData.equal_range(name);
+        auto range = m_aliasData.equal_range(name);
         if (range.first != range.second)
             return range.first->second;
         else
@@ -373,8 +370,7 @@ std::vector< BaseData* > Base::findGlobalField( const std::string &name ) const
 {
     std::vector<BaseData*> result;
     //Search in the aliases
-    typedef MapData::const_iterator mapIterator;
-    std::pair< mapIterator, mapIterator> range = m_aliasData.equal_range(name);
+    auto range = m_aliasData.equal_range(name);
     for (mapIterator itAlias=range.first; itAlias!=range.second; ++itAlias)
         result.push_back(itAlias->second);
     return result;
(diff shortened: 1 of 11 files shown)


marques-bruno:
[ci-build]

SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/DDGLink.h
```diff
@@ -0,0 +1,126 @@
...
+
+    explicit BaseDDGLink(const InitDDGLink& init);
+
+    virtual ~BaseDDGLink() override;
+
+    void setOwner(Base* owner);
+
+    void set(Base* linkedBase);
+
+    Base* get();
```
jnbrunet, line 74: I would add a `const Base * get() const;` in case I create a const link `DDGLink<const MyComponent>`

SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
```diff
@@ -354,6 +380,18 @@ std::vector< BaseData* > Base::findGlobalField( const std::string &name ) const
     return result;
 }
 
+/// Find fields given a name: several can be found as we look into the alias map
+std::vector< BaseDDGLink* > Base::findGlobalDDGLink( const std::string &name ) const
+{
+    std::vector<BaseDDGLink*> result;
+    //Search in the aliases
+    typedef MapDDGLink::const_iterator mapIterator;
```
jnbrunet: I think this is a good case for using auto : `auto range = m_aliasDDGLink.equal_range(name);`

SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
```diff
@@ -368,6 +406,19 @@ BaseLink* Base::findLink( const std::string &name ) const
         return nullptr;
 }
 
+/// Find a link given its name.
+/// Return nullptr if not found. If more than one link is found (due to aliases), only the first is returned.
+BaseDDGLink* Base::findDDGLink( const std::string &name ) const
+{
+    //Search in the aliases
+    typedef MapDDGLink::const_iterator mapIterator;
```
jnbrunet: Same here

SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
```diff
@@ -325,6 +344,13 @@ void Base::removeData(BaseData* d)
     m_aliasData.erase(range.first, range.second);
 }
 
+void Base::removeDDGLink(BaseDDGLink* d)
+{
+    m_vecDDGLink.erase(std::find(m_vecDDGLink.begin(), m_vecDDGLink.end(), d));
+    typedef MapDDGLink::const_iterator mapIterator;
```
jnbrunet: And here

jnbrunet:
Hey @marques-bruno !

I think this will help us a lot to add robustness to our components.

With this PR, do you see any reason why one would use `BaseLink`s in his component instead of a `BaseDDGLink`s?


Also, maybe instead of have a new `InitDDGLink` function in BaseClass, we could simply overload the `initLink` with
`initLink(BaseDDGLink* owner, std::string name, std::string help, std::string group = "")` ?

Thanks !

marques-bruno:
> Hey @marques-bruno !
> 
> I think this will help us a lot to add robustness to our components.
> 
> With this PR, do you see any reason why one would use `BaseLink`s in his component instead of a `BaseDDGLink`s?
> 
> Also, maybe instead of have a new `InitDDGLink` function in BaseClass, we could simply overload the `initLink` with
> `initLink(BaseDDGLink* owner, std::string name, std::string help, std::string group = "")` ?
> 
> Thanks !

Thanks @jnbrunet for your review, I'll do the changes tomorrow.
Regarding using BaseLink instead of BaseDdgLink, BaseLinks are actually used internally in DDGNodes (for input/outputs). So they have their use. What I want to replace with DDGLink, are the SingleLink entries in sofa components.
We'll discuss it at the sofa meeting if you're there tomorrow ;)

SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/DDGLink.h
marques-bruno: Done. I'm not super happy about the design though: The setter taking a const Base* stores the const_cast'ed pointer internally and prays nobody calls the non-const getter afterwards.

When manipulating DDGLink<T> objects, that's not an issue because the wrong use case is encapsulated in the child class that guarantees it doesn't happen. But nothing prevents someone from manipulating a DDTLink<const T> through its abstract type BaseDDGLink and calling the non-const getter, which would lead to undefined behaviors.

@jnbrunet I'd like a second pair of eyes on that one, if you have some time, in case you would see a way to fix that..?

jnbrunet: Hey @marques-bruno 

I think the idea of having a "const DDGLink" is not really that the DDGLink is declared as `const`, but more that its linked `T` object is. Since the `DDGLink::get` function cannot realistically be marked `const` (because it updates its data counters), you would not have a `const DDGLink<T>` but instead have a `DDGLink<const T>`.

It think this is how I would do it:
```
// The BaseDDGLink does not actually contains a linked object
class BaseDDGLink : public DDGNode
{
public:
// (...)
    void set(Base* linkedBase) = 0; // Pure virtual, must be override by DDGLink<T>
    // !! Remove the const  version of set from here

    const Base* get() = 0; // Pure virtual, must be override by DDGLink<T>
    // !! No non-const getter here !!
    
// (...)
}

// Non-Const version
template <class T>
class DDGLink : public BaseDDGLink
{
  public:
// (...)
    void set(T* linkedBase) final;
    T * get(); // Here the owned of the DDGLink will be able to modify the linked object T
    const T * get() final;
// (...)

private:
    T * m_linkedBase {nullptr};
}

// Const version
template <class T>
class DDGLink<const T> : public BaseDDGLink
{
  public:
// (...)
    void set(T* linkedBase) final;
    void set(const T* linkedBase);
    const T * get() final;
// (...)
private:
    const T * m_linkedBase {nullptr};
}
```

What do you think?

marques-bruno: Hey! So I'm gonna revert the latest changes (the handling of const template types in DDGLinks), because clearly the implementation really sucks and adds more problems than it solves...

@jnbrunet I personally don't have a use case for DDGLink<const T*>, and I'm wondering if enforcing constness here isn't a bit of a mistake. Sure it would enforce and explicit the fact that a component cannot modify another one through its link, but does this issue really occurred already?
I think the same applies for datafields. We *could* imagine a use case where a datafield should be readonly, but maybe we don't need to enforce constness for that. I'm thinking using flags, like we do on data (required, readonly, etc)

jnbrunet: Well, side effects was pretty much one of the starting point of the discussion on data, no?

Maybe I did not understand correctly our discussions, but I though that we talked about enforcing links to be "const" by "default". This will prevent calling a method of the linked component that would changes its internal state without knowing it and instead use the data mechanism which is made for this.

If your link is "const", then calling a method of the linked object that changes its internal state will produce a compilation error, forcing the developer to think if it is the right way of doing this and also help the PR reviewers (eg: why are you changing the link to a non-const?). 

This should also point out some "limits" of the data mechanism (links that will have to be non-const because datas cannot do its job for this case).

marques-bruno:

PR #1227, opened 2019-11-29
[SofaCore] Add ddg callback

marques-bruno:
Depends on #1226 

This PR provides a callback method in Base to easily connect DDGNodes (inputs & outputs) within a component, through the use of DataTrackerEngines.

It is inspired by a POC from @damienmarchal in a branch we have in SofaDefrost/sofa.
The syntax to create a callback is the following:

```cpp
addUpdateCallback("dataTrackerEngineName", {&inputs}, []()->ComponentState{
// Some code in a lambda setting the componentState by return value
}, {&outputs});
```
This method will create a DataTrackerEngine in a map within Base, give it a name, set this as its owner, set the given list of inputs and outputs for the engine, and finally, use the given lambda as a delegate method for the update() function of the engine (delegate, because engine->updateAllInputsIfDirty() is called before the lambda, and engine->cleanDirty() is called after... just like a normal engine.)

The test in SofaSimulationGraph_test/DDGLink_test.cpp has been updated to show how to use it in a simple class with 1 input / 1 output.

Actually, the code doesn't work... I get an ugly exception when my engine tries to call the delegate lambda:
```
unknown file: Failure
C++ exception with description "bad_function_call" thrown in the test body.
```

No idea what this is, if you have suggestions, please don't hesitate to correct me or tell me what I did wrong... ;)
Enjoy the WE =)

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaCore/CMakeLists.txt
+++ b/SofaKernel/modules/SofaCore/CMakeLists.txt
@@ -110,6 +110,7 @@ set(HEADER_FILES
     ${SRC_ROOT}/objectmodel/ContextObject.h
     ${SRC_ROOT}/objectmodel/DDGNode.h
     ${SRC_ROOT}/objectmodel/Data.h
+    ${SRC_ROOT}/objectmodel/DDGLink.h
     ${SRC_ROOT}/objectmodel/DataFileName.h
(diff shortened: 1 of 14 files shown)


marques-bruno:

PR #1228, opened 2019-12-02
[SofaCore] Add links vector

marques-bruno:
Here we go as promised:

This PR stores within a SOFA Object (inheriting Base) links to the owners that given object in a vector, that can be accessed by calling getDDGLinkOwners().

Currently, pointers are stored within a vector of const Base*, but we could imagine using a map, that would store the link owners according to a CATEGORY, (ForceField, Mass, MO, Topo...), or having a multimap with the component names..? I don't know, I guess it depends on the use case.
@jnbrunet  @epernod that PR is specifically for you, so you probably have a better idea of how you'd like to manipulate that list. let me know what you think!

 

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaCore/CMakeLists.txt
+++ b/SofaKernel/modules/SofaCore/CMakeLists.txt
@@ -110,6 +110,7 @@ set(HEADER_FILES
     ${SRC_ROOT}/objectmodel/ContextObject.h
     ${SRC_ROOT}/objectmodel/DDGNode.h
     ${SRC_ROOT}/objectmodel/Data.h
+    ${SRC_ROOT}/objectmodel/DDGLink.h
     ${SRC_ROOT}/objectmodel/DataFileName.h
     ${SRC_ROOT}/objectmodel/DetachNodeEvent.h
     ${SRC_ROOT}/objectmodel/Event.h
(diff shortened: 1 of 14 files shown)


jnbrunet:
Hey @marques-bruno,

I think the idea was to have a collection of "DDGLinks" seen as one data. For example:

```
// Inputs
DDGLink<MechanicalObject> d_mechanical_object; // One mechanical object
DDGLinks<BaseForceField> d_forcefields; // Multiple forcefields acting on the MO
DDGLink<BaseMass> d_mass; // One mass acting on the MO

// Outputs
Data<VecCoord> d_solution;
```

Here we want to update the solution vector if either the MO, one of the forcefields, or the mass has changed.

marques-bruno:
> Hey @marques-bruno,
> 
> I think the idea was to have a collection of "DDGLinks" seen as one data. For example:
> 
> ```
> // Inputs
> DDGLink<MechanicalObject> d_mechanical_object; // One mechanical object
> DDGLinks<BaseForceField> d_forcefields; // Multiple forcefields acting on the MO
> DDGLink<BaseMass> d_mass; // One mass acting on the MO
> 
> // Outputs
> Data<VecCoord> d_solution;
> ```
> 
> Here we want to update the solution vector if either the MO, one of the forcefields, or the mass has changed.

Hi @jnbrunet, that's another issue right? This PR proposes to get some sort of "bidirectionality" in the Data Links without breaking the "acyclicity" of the DDG.
I guess it would not be a problem to overload DDGLink<T> to handle a vector of links of the same type instead of a single link of type T.

Just to understand a bit better: your example code would be the kind of code one would find in an ode solver for instance, right?

marques-bruno:

PR #1230, opened 2019-12-03
[SofaConstraint] Fix crash with PrecomputedConstraintCorrection

fredroy:
Fix crash when solver was not in the same context (node)
I am not really an expert of solvers and Vectors (VecId stuff) but the reference to the vector of position is changed everytime the odesolver solves if this one (and I guess the linearsolver) is not in the same context.
I roughly guess that it is because the system has not the same number of nodes and everything.
Anyway, reviews/comments from @ChristianDuriez would be most welcome 😺 

This PR: 

### DIFF
--- a/modules/SofaConstraint/PrecomputedConstraintCorrection.inl
+++ b/modules/SofaConstraint/PrecomputedConstraintCorrection.inl
@@ -302,9 +302,8 @@ void PrecomputedConstraintCorrection<DataTypes>::bwdInit()
         }
 
 
-        helper::WriteAccessor< Data< VecCoord > > posData = *this->mstate->write(core::VecCoordId::position());
-        VecCoord& pos = posData.wref();
-        const VecCoord prev_pos = pos;
+        helper::ReadAccessor< Data< VecCoord > > rposData = *this->mstate->read(core::ConstVecCoordId::position());
+        const VecCoord prev_pos = rposData.ref();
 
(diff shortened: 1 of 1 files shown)


guparan:
[ci-build][with-all-tests]

guparan:
Maybe @jnbrunet could help reviewing this?

jnbrunet:
Seems this PR doesn't change the algorithm, so should be non breaking.

It is however weird that the pointer get changed by the solver... While this patch will work around this, I would be quite interested to know why it was happening in the first place.

@fredroy , could you just change `WriteAccessor` to `WriteOnlyAccessor` so that we do not trigger a data update every time (we are overriding every entries of the position vector).

modules/SofaConstraint/PrecomputedConstraintCorrection.inl
```diff
@@ -393,6 +394,9 @@ void PrecomputedConstraintCorrection<DataTypes>::bwdInit()
         for (unsigned int i = 0; i < velocity.size(); i++)
             velocity[i] = prev_velocity[i];
 
+        helper::WriteAccessor< Data< VecCoord > > wposData = *this->mstate->write(core::VecCoordId::position());
```
jnbrunet: @fredroy you missed one here

jnbrunet:
All good for me

jnbrunet (approved):

PR #1231, opened 2019-12-03
[SofaBaseTopology] Fix SparseGrid obj loading + tests

fredroy:
It started from the case where SparseGrid could not open .OBJ files (because it is upper case 🐶 )
Now SparseGrid relies on the FactoryMesh mechanism for its extension management.

BUT MeshOBJ is super deprecated, and the legacy helper::io::MeshLoader (not the one in sofa::core::loader) should be not used anymore.
BUT SparseGrid does not really handle src="@loader" syntax very well now, and it will be upgraded in the future. 

I added some tests anyway 💛 

This PR: 


fredroy:
[ci-build][with-all-tests]

PR #1232, opened 2019-12-10
[SofaKernel] Remove deprecated SOFA_DEBUG macro 

epernod:
I suppose it is an old macro. Was not accesible from cmake config and always equal to true if compiled in debug mode.

only a few code was using it and were already checking NDebug as well.

This PR: 


epernod:
[ci-build][with-all-tests]

PR #1242, opened 2020-01-17
[SofaCUDA] Compilation without OpenGL

fredroy:
Small PR to allow compilation of SofaCUDA plugin with SOFA_NO_OPENGL flag (i.e disabling OpenGL)
Remarks:
- The components' draw() functions in SofaCUDA should share their code with their "normal"  version.
For example, CudaUniformMass with cudarigid3f should call the draw() of UniformMass<Rigid3>
Some work has to be done further in Sofa itself, that's why I just add preprocessor tests (#ifdef SOFA_NO_OPENGL) and did not use the drawtool functions 👅 
- Obviously, if you compile without OpenGL, you wont be able to share OpenGL buffer in CUDA. In other words it is useless to use CudaVisualModel without SOFA_NO_OPENGL.
(Actually CudaVisualModel should be renamed CudaOglModel)

This PR: 

### DIFF
--- a/applications/plugins/SofaCUDA/CMakeLists.txt
+++ b/applications/plugins/SofaCUDA/CMakeLists.txt
@@ -10,11 +10,6 @@ if(NOT WIN32)
     set(CUDA_PROPAGATE_HOST_FLAGS OFF)
 endif()
 
-if(${SOFA_NO_OPENGL})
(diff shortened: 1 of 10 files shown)


fredroy:
[ci-build][with-all-tests]

PR #1028, opened 2019-05-07
[SofaQtGui] Add a profiling window based on AdvanceTimer records

epernod:
I know the QtGui is going to be deprecated soon but I wanted/needed this feature since so long.

Basically allow to see the animation step duration (ms) in a graphView, then allow to navigate on the graph or on the sliders and see the different substeps exectuted during this animation step.
Substeps are displayed in a Tree, in the right order and with their respective time and percentage resp. to the full step duration.

All the mechanism to parse data could be kept for future runSofa2.

The windows is a "little" inspired from Unity3D profiler.

image

______________________________________________________

This PR: 
- [ ] builds with SUCCESS for all platforms on the CI.
- [ ] does not generate new warnings.
- [ ] does not generate new unit test failures.
- [ ] does not generate new scene test failures.
- [ ] does not break API compatibility.
- [ ] is more than 1 week old (or has fast-merge label).

**Reviewers will merge only if all these checks are true.**

PR #1244, opened 2020-01-18
[SofaHelper] AdvancedTimer wasn't using the good timer ids for the label assignments

jnbrunet:
PR #1028 added a getter for the records of the advanced timer that automatically assigns the string version of the record's **id** to a new field called **label**.

There was a small problem with this, records can have different types of ids (IdTimer, IdStep, IdVal or IdObj). Casting a record id to a wrong type causes the wrong string version of the id to be assigned to the label.

This PR fixes this.

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaHelper/src/sofa/helper/AdvancedTimer.cpp
+++ b/SofaKernel/modules/SofaHelper/src/sofa/helper/AdvancedTimer.cpp
@@ -1410,19 +1410,19 @@ helper::vector<Record> AdvancedTimer::getRecords(IdTimer id)
         switch (r.type) {
             case Record::RBEGIN: // Timer begins
             case Record::REND: // Timer ends
-                r.label = (std::string) AdvancedTimer::IdTimer(r.id);
+                r.label = IdTimer::IdFactory::getName(r.id);
                 break;
             case Record::RSTEP_BEGIN: // Step begins
             case Record::RSTEP_END: // Step ends
             case Record::RSTEP: // Step
-                r.label = (std::string) AdvancedTimer::IdStep(r.id);
+                r.label = IdStep::IdFactory::getName(r.id);
                 break;
             case Record::RVAL_SET: // Sets a value
             case Record::RVAL_ADD: // Adds a value
-                r.label = (std::string) AdvancedTimer::IdVal(r.id);
+                r.label = IdVal::IdFactory::getName(r.id);
                 break;
             default:
-                r.label = (std::string) AdvancedTimer::IdObj(r.id);
+                r.label = "Unknown";
                 break;
         }
     }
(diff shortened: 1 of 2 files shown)


SofaKernel/modules/SofaHelper/src/sofa/helper/AdvancedTimer.cpp
```diff
@@ -1406,8 +1406,27 @@ std::map<AdvancedTimer::IdStep, StepData> AdvancedTimer::getStepData(IdTimer id,
...
+            case Record::RSTEP_END: // Step ends
+            case Record::RSTEP: // Step
+                r.label = (std::string) AdvancedTimer::IdStep(r.id);
+                break;
+            case Record::RVAL_SET: // Sets a value
+            case Record::RVAL_ADD: // Adds a value
+                r.label = (std::string) AdvancedTimer::IdVal(r.id);
+                break;
+            default:
+                r.label = (std::string) AdvancedTimer::IdObj(r.id);
```
epernod: should not be IdStep by default? (just compared to the old version)

SofaKernel/modules/SofaHelper/src/sofa/helper/AdvancedTimer.cpp
```diff
@@ -1406,8 +1406,27 @@ std::map<AdvancedTimer::IdStep, StepData> AdvancedTimer::getStepData(IdTimer id,
 helper::vector<Record> AdvancedTimer::getRecords(IdTimer id)
 {
     TimerData& data = timers[id];
-    for (unsigned int i=0; i<data.records.size(); ++i)
-        data.records[i].label = std::string(AdvancedTimer::IdStep(data.records[i].id));
+    for (Record & r : data.records) {
+        switch (r.type) {
+            case Record::RBEGIN: // Timer begins
+            case Record::REND: // Timer ends
+                r.label = (std::string) AdvancedTimer::IdTimer(r.id);
```
epernod: use use std::string(... ) instead of (std::string)... ?

jnbrunet: It is exactly the same, see [here](https://stackoverflow.com/questions/45505861/function-style-cast-vs-constructor) and [here](https://web.archive.org/web/20070929132736/http://www.acm.org/crossroads/xrds3-1/ovp3-1.html). Apparently, since we are in c++, the good way to do it would be static_cast (which again, produce the exact same thing). I'll change it with static_cast if that's ok with you ;)

SofaKernel/modules/SofaHelper/src/sofa/helper/AdvancedTimer.cpp
jnbrunet: Now that I think about it, the default should never happen. If it does, it means we added a new Id type and we did not treat it in this function.

Casting it to a default id (IdStep or IdObj) will produce undefined behaviour since it will try to get the name value of the id from another factory of id. This will be hard to debug.

For example, if we have:
```
IdTimer         |        IdStep
----------------|-----------------------
1: "Timer1"     |       1: "Step 1"
2: "Timer2"     |       2: "Step 2"
```

Then
```
IdTimer id (1); //  "Timer 1"
std::cout << (std::string) IdStep(id) << std::endl;
//  "Step 1"
```
is wrong.

I suggest we use a msg_error for the default, so that we can easily fix it once we see the error.

What do you think?

PR #1248, opened 2020-01-29
[SofaSimulationCore] Add a way to fill a multi vector from a base vector with a MechanicalOperations (mops).

jnbrunet:
The MechanicalOperations (mops) already had the method `multiVector2BaseVector` that allows us to fill a BaseVector (one continuous memory vector) from a multi vector (group of vectors contained inside mechanical objects).

However, the were no way to do the inverse (build a multi vector from a continuous vector) beside creating a new visitor, which is now done automatically by the mops.

This PR is only adding a new method, it is non-breaking.

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/MechanicalOperations.h
+++ b/SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/MechanicalOperations.h
@@ -137,7 +137,7 @@ public:
 
     void multiVector2BaseVector(core::ConstMultiVecId src, defaulttype::BaseVector *dest, const sofa::core::behavior::MultiMatrixAccessor* matrix);
     void baseVector2MultiVector(const defaulttype::BaseVector *src, core::MultiVecId dest, const sofa::core::behavior::MultiMatrixAccessor* matrix);
-    void multiVectorPeqBaseVector(core::MultiVecDerivId dest, defaulttype::BaseVector *src, const sofa::core::behavior::MultiMatrixAccessor* matrix);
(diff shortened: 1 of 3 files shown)


SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/MechanicalOperations.h
```diff
@@ -136,6 +136,7 @@ class SOFA_SIMULATION_CORE_API MechanicalOperations
     void addSubMBK_ToMatrix(const sofa::core::behavior::MultiMatrixAccessor* matrix, const helper::vector<unsigned> & subMatrixIndex, SReal mFact, SReal bFact, SReal kFact);
 
     void multiVector2BaseVector(core::ConstMultiVecId src, defaulttype::BaseVector *dest, const sofa::core::behavior::MultiMatrixAccessor* matrix);
+    void baseVector2MultiVector(core::MultiVecId dest, defaulttype::BaseVector *src, const sofa::core::behavior::MultiMatrixAccessor* matrix);
```
guparan:
```cpp
void baseVector2MultiVector(defaulttype::BaseVector *src, core::MultiVecId dest, const sofa::core::behavior::MultiMatrixAccessor* matrix);
```
would be more consistent with multiVector2BaseVector and with the idea of "XXX to YYY", don't you think ?

jnbrunet: Done

jnbrunet:
I've also mark const the source arguments from both `baseVector2MultiVector` and `multiVectorPeqBaseVector` so that it is clear that the source will never be changed by those two functions.

PR #1266, opened 2020-03-04
[SofaKernel] Cleaner output when the creation of an object fails

jnbrunet:
Error message when the creation of an object fails are difficult to read and it isn't always obvious where the error is.

With this PR, the message is a little bit more clear. Now we only need to add error message inside the components `canCreate` methods (see the changes I made in the `ForceField` class in this PR).

Before:
image

After:
image

This PR: 


jnbrunet:
> CI on windows complains due to C++ revision incompatibility
> Otherwise good to me!

Should be good now.

Funny thing, the keyword NOT (!), which is a [valid c++ keyword](https://en.cppreference.com/w/cpp/keyword/not), isn't supported in Visual Studio unless you deactivate the VS C++ extensions (flag `/Za`). For them, removing the support of a keyword is a language extension.

PR #1268, opened 2020-03-10
[SofaSimulationCore] FIX resizing of bboxes in UpdateBoundingBoxVisitor

marques-bruno:
I don't understand why this check was made:
Without invalidating the bbox, the node's bbox will only be sized up, and never down with this visitor, to my understanding..

This PR: 


epernod:
[ci-build][with-all-tests]

PR #1272, opened 2020-03-12
[Modeler] Make private the tinyxml dependence

jnbrunet:
See #1271 

This PR: 

### DIFF
--- a/applications/projects/Modeler/lib/CMakeLists.txt
+++ b/applications/projects/Modeler/lib/CMakeLists.txt
@@ -55,9 +55,10 @@ if(Qt5Core_FOUND)
     set(FORM_FILES_GROUP "Generated")
 endif()
 
-add_library(${PROJECT_NAME} STATIC ${HEADER_FILES} ${MOC_FILES} ${FORM_FILES} ${SOURCE_FILES})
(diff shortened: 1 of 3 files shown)


applications/projects/Modeler/lib/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 cmake_minimum_required(VERSION 3.1)
 project(SofaModeler)
 
-find_package(TinyXML REQUIRED)
+sofa_find_package(TinyXML REQUIRED)
```
guparan: Note (FYI): this works but is not necessary because `SOFAMODELER_HAVE_TINYXML` is not needed in SofaModelerConfig.cmake.in (CMake level propagation) nor in config.h.in (C++ level propagation).

jnbrunet: Ha I see, I wasn't sure here since other packages that depends on tinyxml were doing it with `sofa_find_package` and it wasn't clear to me what this function was doing.  I'll put back the `find_package` since the modeler always have tinyxml

guparan (approved):

guparan:
Merged too quickly, there was a link error on Windows :disappointed:

guparan:
@jnbrunet Why did you set SofaModeler library SHARED instead of STATIC?

jnbrunet:
> @jnbrunet Why did you set SofaModeler library SHARED instead of STATIC?

@guparan Because static libraries bring their dependencies with them. This is especially bad in this case since the SofaModeler (the exec) is the only one depending on SofaModeler (the lib).

This was the main cause of having to find tinyxml in every project that links against Sofa.

PR #1237, opened 2019-12-23
[SofaKernel] replace all serr by msg_error/msg_warning

epernod:
Tried not to go too quickly....
- Replaced serr by msg_error/msg_warning and msg_info regarding the case.
- Remove all commented serr.
- Use stringstream to accumulate errors when inside a loop.

This PR only trigger classes in SofaKernel. See #1236 for changes in Sofa Modules

______________________________________________________

This PR: 
- [ ] builds with SUCCESS for all platforms on the CI.
- [ ] does not generate new warnings.
- [ ] does not generate new unit test failures.
- [ ] does not generate new scene test failures.
- [ ] does not break API compatibility.
- [ ] is more than 1 week old (or has fast-merge label).

**Reviewers will merge only if all these checks are true.**

PR #1280, opened 2020-03-21
 [SofaEngine] Fix engine unit tests

fredroy:
Following #1237 , some error messages were not expected in some unit tests; triggering a FAILED status.
This PR fixes that, and add two more (real) unit tests for RandomPointDistributionInSurface and JoinPoints engines.
Moreover, it fixes also an example of the ExtrudeEngine scene and adds a example scene for RandomPointDistributionInSurface .

This PR: 


fredroy:
[ci-build][force-full-build]

PR #1291, opened 2020-03-27
[SofaBaseMechanics] Change data name: handleTopologicalChanges in UniformM

hugtalbot:
All in the title as said in the dev meeting.

This PR: 


jnbrunet:
Just to be that guy, I think it should be handleTopologyChange**s** (plural).

Edit:
In fact, I think it should be **handleTopologicalChanges**

epernod (approved):

PR #1313, opened 2020-04-02
[All] Fix SOFA_LIBSUFFIX used in Debug by PluginManager

epernod:
The path to the pluginManager.cpp is incomplete in the CMakeLists.

Thus prefix was always empty and nod set to _d in debug mode.

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaHelper/CMakeLists.txt
+++ b/SofaKernel/modules/SofaHelper/CMakeLists.txt
@@ -334,7 +334,8 @@ endif()
 # since this is configuration specific it is a bit more convenient to put it as a debug compile definition for
 # PluginManager.cpp, at the expense of being much less visible compare to having it in the generated
 # SofaFramework/config.h
-set_property(SOURCE system/PluginManager.cpp APPEND PROPERTY COMPILE_DEFINITIONS_DEBUG "SOFA_LIBSUFFIX=_d" )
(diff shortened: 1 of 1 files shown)


fredroy (approved):

SofaKernel/modules/SofaHelper/CMakeLists.txt
```diff
@@ -334,7 +334,8 @@ endif()
 # since this is configuration specific it is a bit more convenient to put it as a debug compile definition for
 # PluginManager.cpp, at the expense of being much less visible compare to having it in the generated
 # SofaFramework/config.h
-set_property(SOURCE system/PluginManager.cpp APPEND PROPERTY COMPILE_DEFINITIONS_DEBUG "SOFA_LIBSUFFIX=_d" )
+set_property(SOURCE ${CMAKE_CURRENT_SOURCE_DIR}/src/sofa/helper/system/PluginManager.cpp APPEND PROPERTY COMPILE_DEFINITIONS_DEBUG "SOFA_LIBSUFFIX=_d" )
```
guparan:
```cmake
set_property(SOURCE ${SRC_ROOT}/system/PluginManager.cpp APPEND PROPERTY COMPILE_DEFINITIONS_DEBUG "SOFA_LIBSUFFIX=_d")
```

guparan (changes requested):
Thanks for the fix @epernod 
I only suggest to use `SRC_ROOT` instead of plain path for better maintainability.

PR #1318, opened 2020-04-03
[SofaGeneralEngine] Fix draw of the sphere in SphereROI

hugtalbot:
Sphere drawn was actually half the size of the Sphere used..

Thank you @etagliabue for the notice !
 
Problem was:

This PR: 

### DIFF
--- a/modules/SofaGeneralEngine/SphereROI.inl
+++ b/modules/SofaGeneralEngine/SphereROI.inl
@@ -475,7 +475,7 @@ void SphereROI<DataTypes>::draw(const core::visual::VisualParams* vparams)
         {
 
             drawcenters.push_back(c[i]);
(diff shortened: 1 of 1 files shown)


modules/SofaGeneralEngine/SphereROI.inl
```diff
@@ -475,7 +475,7 @@ void SphereROI<DataTypes>::draw(const core::visual::VisualParams* vparams)
         {
 
             drawcenters.push_back(c[i]);
-            drawradii.push_back((float)(r[i] * 0.5));
+            drawradii.push_back((float)r[i]);
```
epernod: as you are on this line.
float(r[i])

epernod (approved):

fredroy (approved):

PR #1321, opened 2020-04-06
[Regression] Update hash

hugtalbot:
As simple as this.
Ciao for 2 regression errors theoretically!

This PR: 


epernod:
@hugtalbot as we are on those tests. Could you quickly test if this one is working fine on your computer:
Regression_test/TopologyRegression_test.sceneTest/RemovingTetra2TriangleProcess_withOptions

I'm not sure to understand what is the problem here.

PR #1328, opened 2020-04-10
[CImgPlugin] fix debug lib copying

fredroy:
- On Windows, Cmake was not copying the debug version of the libpng/zlib/libtiff DLLs, but was still linking with the debug library (lib) files. So it was not loading the CImgPlugin when required while running runSofa_d

+ change the libsuffix with a newer version of add_definition() in debug mode in multiconfig mode (ouf)
(https://cmake.org/cmake/help/v3.17/policy/CMP0043.html#policy:CMP0043) 

This PR: 


guparan:
What's the status here @fredroy ?

fredroy:
Nothing much ; as everybody is debugging in ReleaseWithDebugInfo mode 🤔 
I will let it on stand-by for now.

PR #1333, opened 2020-04-14
[SofaGeneralEngine] Remove remaining BoxROI after SofaEngine move

hugtalbot:
:dove: 

This PR: 


fredroy:
These files are not in the CMakeLists of SofaGeneralEngine, and were not updated for 14mo.
And the good BoxROI is still in SofaEngine, still updated and tested. So I guess there is no problem with deleting those (and it seems this PR does it well 🐱‍🐉)

fredroy (approved):

PR #1335, opened 2020-04-15
[SofaBaseTopology][SofaExporter] Fix failing tests due to changes in topology

hugtalbot:
Fix tests further to #1323 

- SofaExporter_test was expecting to export Hexa while option computeHexa was false in RGrid
- RegularGridTopology_test was expecting the wrong number of edges since the computation is now under the condition whether the triangles are used or not to compute the edges

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaBaseTopology/GridTopology.h
+++ b/SofaKernel/modules/SofaBaseTopology/GridTopology.h
@@ -184,8 +184,9 @@ public:
     /// Get Cube index, similar to \sa hexa method
     int cube(int x, int y, int z) const { return hexa(x,y,z); }
 
-	/// Get the actual dimension of this grid using Enum @sa Grid_dimension
-	Grid_dimension getDimensions() const;
+    /// Get the actual dimension of this grid using Enum @sa Grid_dimension
+    Grid_dimension getDimensions() const;
+
 public:
     /// Data storing the size of the grid in the 3 directions
     Data<Vec3i> d_n;
--- a/SofaKernel/modules/SofaBaseTopology/SofaBaseTopology_test/RegularGridTopology_test.cpp
+++ b/SofaKernel/modules/SofaBaseTopology/SofaBaseTopology_test/RegularGridTopology_test.cpp
@@ -158,12 +158,12 @@ bool RegularGridTopology_test::regularGridFindPoint()
                 );
     grid->init();
 
-    EXPECT_EQ(grid->findPoint(Coordinates{ .4, .4, .4 }),
-        -1); // No margin set means anything position rounded to a valid
+    EXPECT_EQ(grid->findPoint(Coordinates{ .4, .4, .4 }), -1);
+    // No margin set means anything position rounded to a valid
     // node will be returned
     EXPECT_EQ(grid->findPoint(Coordinates{ .51, .51, .51 }), 0);
-    EXPECT_EQ(grid->findPoint(Coordinates{ .51, .51, .51 }, Epsilon(0.01)),
-        -1); // Margin set means anything within a radius of 0.01*cell_size
+    EXPECT_EQ(grid->findPoint(Coordinates{ .51, .51, .51 }, Epsilon(0.01)), -1);
+    // Margin set means anything within a radius of 0.01*cell_size
     // will be returned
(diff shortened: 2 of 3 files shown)


guparan:
I quickly applied a simple formatter online and finished manually. This file was a mess.

SofaKernel/modules/SofaBaseTopology/SofaBaseTopology_test/RegularGridTopology_test.cpp
```diff
@@ -207,20 +231,34 @@ std::vector<std::vector<int>> dimvalues={
     {-2,1,1, 1, 1},
 };
 
-TEST_P(RegularGridTopology_test, regularGridSize )
+TEST_P(RegularGridTopology_test, regularGridSizeComputeEdgeFromTriangle)
```
epernod, line 234: why changing that?
the test is checking all the topology buffer sizes, not only edges.

SofaKernel/modules/SofaBaseTopology/SofaBaseTopology_test/RegularGridTopology_test.cpp
```diff
@@ -207,20 +231,34 @@ std::vector<std::vector<int>> dimvalues={
...
+    }
+}
+
+TEST_P(RegularGridTopology_test, regularGridSize)
+{
+    /// We check if this test should returns a warning.
+    if (GetParam()[3] == 1)
+    {
+        EXPECT_MSG_EMIT(Warning);
+        ASSERT_TRUE(regularGridSize(GetParam(), false));
```
epernod, line 254: very good idea to put the computeTriangle as a parameter. IT would be even better to duplicate the test or at least have some with the option set to false (this will test the old behavior in fact)

SofaKernel/modules/SofaBaseTopology/GridTopology.h
```diff
@@ -184,8 +184,9 @@ using MeshTopology::getHexahedron;
     /// Get Cube index, similar to \sa hexa method
     int cube(int x, int y, int z) const { return hexa(x,y,z); }
 
-	/// Get the actual dimension of this grid using Enum @sa Grid_dimension
-	Grid_dimension getDimensions() const;
+    /// Get the actual dimension of this grid using Enum @sa Grid_dimension
+    Grid_dimension getDimensions() const;
```
epernod, line 188: don't have "the eyes in front of the holes".. is there a change here?

epernod (changes requested):

SofaKernel/modules/SofaBaseTopology/SofaBaseTopology_test/RegularGridTopology_test.cpp
epernod: sorry I meant the opposite... set to true. To test the new behavior. This can be done in another PR if you want to fix the dashboard quickly!

epernod:
[ci-build][with-all-tests]

SofaKernel/modules/SofaBaseTopology/GridTopology.h
guparan: Nope, only formatting.

SofaKernel/modules/SofaBaseTopology/SofaBaseTopology_test/RegularGridTopology_test.cpp
hugtalbot: I am testing here both cases: with a triangle topology computed and without

epernod: sorry my bad, didn't see the lines above :+1:

epernod: sorry my bad, didn't see the lines above 👍

epernod:
you didn't answer my last question regarding change of name: ```TEST_P(RegularGridTopology_test, regularGridSizeComputeEdgeFromTriangle)```

epernod (approved):

PR #1339, opened 2020-04-17
[all] Clean warnings for option config, again

hugtalbot:
WAIT for #1283 (priority)

This PR: 

### DIFF
--- a/applications/sofa/gui/headlessRecorder/HeadlessRecorder.cpp
+++ b/applications/sofa/gui/headlessRecorder/HeadlessRecorder.cpp
@@ -346,7 +346,9 @@ int HeadlessRecorder::mainLoop()
         }
         else
         {
-            sleep(0.01);
+            //TODO(hugtalbot 2020-04-17): back to zero since sleep(0.01) creates implicit
+            //conversion from 'double' to 'unsigned int' changes value from 0.01 to 0
+            sleep(0);
         }
     }
     msg_info("HeadlessRecorder") << "Recording time: " << recordTimeInSeconds << " seconds at: " << fps << " fps.";
--- a/applications/sofa/gui/headlessRecorder/HeadlessRecorder.cpp
+++ b/applications/sofa/gui/headlessRecorder/HeadlessRecorder.cpp
@@ -346,9 +346,7 @@ int HeadlessRecorder::mainLoop()
         }
         else
         {
-            //TODO(hugtalbot 2020-04-17): back to zero since sleep(0.01) creates implicit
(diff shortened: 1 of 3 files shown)


applications/sofa/gui/headlessRecorder/HeadlessRecorder.cpp
```diff
@@ -346,7 +346,9 @@ int HeadlessRecorder::mainLoop()
         }
         else
         {
-            sleep(0.01);
+            //TODO(hugtalbot 2020-04-17): back to zero since sleep(0.01) creates implicit
+            //conversion from 'double' to 'unsigned int' changes value from 0.01 to 0
+            sleep(0);
```
fredroy: I don't see the point to sleep 0 second 😺 
and it would be better to use `sofa::helper::system::thread::CTime::sleep(10)` as it is system/OS independent (it takes double and it represents milliseconds)

Or just use `std::this_thread::sleep_for(std::chrono::milliseconds(10));`

fredroy (changes requested):

applications/sofa/gui/headlessRecorder/HeadlessRecorder.cpp
fredroy: 😄 obviously the 🧵  is ": thread :"

fredroy (approved):

applications/sofa/gui/headlessRecorder/HeadlessRecorder.cpp
```diff
@@ -346,7 +346,7 @@ int HeadlessRecorder::mainLoop()
         }
         else
         {
-            sleep(0.01);
+            sofa::helper::system::thread::CTime::sleep(10);
```
tgaugry, line 349: Whole module only works on linux, so not sure this is really needed, even if technically more correct.

However, i believe the module is not compiled on the CI and is not compiled by default, so I assume this lacks the proper includes, namely `#include <sofa/helper/system/thread/CTime.h>`

tgaugry (changes requested):

applications/sofa/gui/headlessRecorder/HeadlessRecorder.cpp
hugtalbot: Apparently CI worked as is.
It should therefore pass.

tgaugry: Just checked, it is compiled. I still have no idea where it get the include from. Resolved then.

tgaugry (approved):

PR #1348, opened 2020-04-25
[SofaKernel] Set the Eigen3 cmake search as NO_MODULE to ignore FindEigen3.cmake files

jnbrunet:
Found out that Eigen3 in Fedora 29 installs a `FindEigen3.cmake` that do not set the imported target **Eigen3::Eigen**, resulting in CMake failure when compiling SOFA or linking against it.

```
$ dnf locates /usr/share/cmake/Modules/FindEigen3.cmake
eigen3-devel-3.3.5-2.fc29.noarch : A lightweight C++ template library for vector and matrix math
Repo        : fedora
Matched from:
Filename    : /usr/share/cmake/Modules/FindEigen3.cmake
```

 There is also third party software that install their own, such as CGAL:

```
$ dnf locates /usr/lib64/cmake/CGAL/FindEigen3.cmake
CGAL-devel-4.13-0.2.beta1.fc29.x86_64 : Development files and tools for CGAL applications
Repo        : fedora
Matched from:
Filename    : /usr/lib64/cmake/CGAL/FindEigen3.cmake
```

Hence to avoid any of those, it is safer to set the NO_MODULE options to `find_package(Eigen3)` to look for **Eigen3Config.cmake** instead, which always define the imported target.

This PR: 


jnbrunet:
Well, it seems Windows is unable to find any EigenConfig.cmake

@guparan any idea here? To recap:

**FindEigen3.cmake** does not always define the imported target `Eigen3::Eigen`. **Eigen3Config.cmake** always does, but isn't there on Windows.

fredroy:
I got the problem with Windows (no Eigen3Config.cmake) with a custom out-of-tree plugin.
The problem is that this file is "unconfigured" (Eigen3Config.cmake.in) while extracting the Eigen archive directly.
You need a pass of Cmake/"compilation"/install to have a full directory with a configured Eigen3Config.cmake.

jnbrunet:
> The problem is that this file is "unconfigured" (Eigen3Config.cmake.in) while extracting the Eigen archive directly.

Which actually makes sense, since you are downloading the source code of Eigen, and not a packaged one (as with package managers apt, dnf, etc.). Even if Eigen is header only, it is still the raw source code you are downloading.

Not sure what's the good way to go here.

jnbrunet:
**Update**
It was working before on Windows because we already have a user defined `FindEigen3.cmake` in sofa/cmake (on which I did some modifications not so long ago :facepalm:, sometime I have the memory of a red fish).

The problem I had was probably because this `FindEigen3.cmake` isn't installed alongside SOFA, so if an external plugin links against SOFA, it will hit the `find_package(Eigen3 REQUIRED)` inside our ***Config.cmake and will fail to find it.

@guparan Do you think we should install the `FindEigen3.cmake` (like CGAL is doing)?

guparan:
Sorry this PR went under the radar. I actually added `FindEigen3.cmake` to the install of SOFA in #1397. 
See https://github.com/sofa-framework/sofa/blob/master/CMakeLists.txt#L74-L80

I guess we can close this PR now @jnbrunet ?

jnbrunet:
Yes we can, thanks @guparan !

PR #1353, opened 2020-04-30
[ManifoldTopologies] Remove CHECK_TOPOLOGY macro occurrences

epernod:
this PR depends on #1351  where Data d_checkTopology has been introduced

This PR: 


epernod:
[ci-build][with-all-tests]

epernod:
the checkTopology Data is only used to activate/unactivate the method CheckTopology. 
It should not change message conditio. If it is the case, I missed it. Could you point the line?

PR #1366, opened 2020-05-05
[Geomagic Plugin] Move all code related to device model display in a dedicated class.

epernod:
This PR is a clean before introducing more work on the GeomagicDriver and a GeomagicEmulator.

- Move all code related to model display from GeomagicDriver to an embedded class GeomagicVisualModel. This last class is not a component but it creates the node hierarchy to display the device.
- Various minor changes in Driver like adding a singleLink to the Forcefeedback.

If needed this work would allow to use GeomagicDriver without linking to SofaOpenglVisual nor SofaLoader. (With a few changes in the cmakelists and in code)

geomagic_demo

- Even if this display is only for debug, my eyes were bleeding too much to let that.
Update the mesh to have cleaner normals/vertices. New mesh display:

Geomagic-RigidCubes_00000003

This PR: 


epernod:
Even if this display is only for debug, my eyes were bleeding too much to let that.
Update the mesh to have cleaner normals/vertices. New mesh display: 

Geomagic-RigidCubes_00000003

PR #1377, opened 2020-05-09
[SofaKernel] Fixes a bug where the camera was not moving with the Qt viewer

jnbrunet:
The `BaseCamera` is computing the scene center point and scene radius based on the current bounding box. The `InteractiveCamera` is using these information to compute the zoom/translation speed when navigating in the scene with mouse events. 

Since the `BaseCamera` was computing the scene center point and radius only when `BaseCamera::p_computeZClip` was true, when manually giving the z-clip values, the interaction was not working at all (`sceneCenter` and `sceneRadius` were 0, hence the zoom/translation speed were 0).

This was only visible when using the `QtViewer` since the `QglViewer` is using its own internal camera and is ignoring the cameras added to the scene.

This PR also fixes the caduceus scene as it was using a `Camera` instead of `InteractiveCamera`, which isn't listening to mouse events at all, hence navigating with the mouse wasn't possible with the QtViewer. A little bit problematic when it is the default scene of runSofa.

This PR: 

### DIFF
--- a/examples/Demos/caduceus.scn
+++ b/examples/Demos/caduceus.scn
@@ -12,7 +12,7 @@
 
     <DefaultContactManager name="Response" response="FrictionContact" />
 
-    <Camera position="0 30 90" lookAt="0 30 0"/>
+    <InteractiveCamera position="0 30 90" lookAt="0 30 0"/>
     <LightManager />
     <SpotLight name="light1" color="1 1 1" position="0 80 25" direction="0 -1 -0.8" cutoff="30" exponent="1" />
     <SpotLight name="light2" color="1 1 1" position="0 40 100" direction="0 0 -1" cutoff="30" exponent="1" />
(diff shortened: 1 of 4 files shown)


examples/Demos/caduceus.scn
```diff
@@ -12,7 +12,7 @@
 
     <DefaultContactManager name="Response" response="FrictionContact" />
 
-    <Camera position="0 30 90" lookAt="0 30 0"/>
+    <InteractiveCamera position="0 30 90" lookAt="0 30 0"/>
```
fredroy, line 15: After further investigation, this was set to "Camera" a while back (537a865e1c13103396ced10e86c41985ed8d6375)
I guess this is/was? required by SofaQtQuick. I dont know if it still the case though.

jnbrunet: I see, let's ping @marques-bruno. 

Maybe once SofaQtQuick is release we could put it back? Because now this camera isn't doing anything with both the QtViewer (not responding to mouse/key events) and QglViewer (simply ignoring all cameras). And since we don't have any other viewer...

Issue #1371, opened 2020-05-07
Create Geomagic Emulator


epernod:
Reproduce work similar to what has been done for OmniEmulator.

This could be used to test the plugin on the CI

______________________________________________________

Suggested labels:

epernod:
To go faster I would need help from someone used to do inverse kinematics. 
Basically in the emulator it is now possible to control the virtual device using the angles value per articulation but it has to be computed from the end of the tool (tip) through the base of the tool. So right now the device it turning around the tip instead of turning around the base.

For example from this start:
![GeomagicEmulator-RigidCubes_00000001](https://user-images.githubusercontent.com/21199245/82423546-2c6e2d00-9a84-11ea-886e-ef4879b80f88.png)

Changing the angles does something like that:
![GeomagicEmulator-RigidCubes_00000002](https://user-images.githubusercontent.com/21199245/82423548-2d9f5a00-9a84-11ea-863d-ad7cf33c171e.png)

fredroy:
Any news of the emulator @epernod ?

PR #1378, opened 2020-05-09
[Geomagic] Fix several wrong behaviors in driver code

### LINKED ISSUE #996: [GeomagicDriver]: Strange behaviour of the device at initial point
The first three steps, the haptic device moves in the scene. This is weird since its position is chosen in the function init(). #STC 

Suggested labels:

epernod:
This PR is the following work on Geomagic started in PR #1366 (which is included in this one)

This PR:
- Totally rewrite the GeomagicDriver init. Now, the tool is directly well placed before simulation start and doesn't need 2 or 3 simulation step to be well init.
=> fix #996 

- HD (openHaptics) scheduler communication has been cleaned and fired msg_error with HD error code and description in case of problems.
=> fix the possibility to handle several devices.

- Clean a lot of old code not anymore used. Like the denavit matrix computation as OpenHaptics is doing it internally. 
=> As a consequence, 5 or 6 Data have been removed: dh_XX
- Others changes like variable names replaced to use "device" or "haptic" instead of "omni".  And add Doxygen documentation.

- Add a component GeomagicEmulator  to be able to virtually recreate device motion and interaction. See #1371
=> This is almost working.. just need to do some inverse kinematics to compute the good tools transform matrix from the set of arm angles. ... I beg for help here. 

This PR: 


hugtalbot (approved):

epernod:
@hugtalbot regarding the emulator. To go faster I would need help from someone used to do inverse kinematics. 
Basically in the emulator it is now possible to control the virtual device using the angles value per articulation but it has to be computed from the end of the tool (tip) through the base of the tool. So right now the device it turning around the tip instead of turning around the base.

For example from this start:
GeomagicEmulator-RigidCubes_00000001

Changing the angles does something like that:
GeomagicEmulator-RigidCubes_00000002

PR #1381, opened 2020-05-13
[SofaExporter] FIX Compilation of SofaExporter_test

marques-bruno:
missing libs in target_link_libraries for SofaBase, SofaCommon and SofaGeneral after "modularization"

This PR: 


marques-bruno:
> I guess you made sure these libs were really needed from the code @marques-bruno ?

Yes, at least in WriteState_test.cpp

fredroy (approved):
Weird that it is/was passing on the CI and stuff without those...
(and you are not obliged to add SofaBase and SofaCommon as they are deps of SofaGeneral, but whatever)

PR #1383, opened 2020-05-13
[SofaKernel] Filerepository gettemppath

faichele:
Adds a helper function to the FileRepository class that returns a platform-specific path to a temporary directory.
Windows and Linux platforms supported. 

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaHelper/src/sofa/helper/system/FileRepository.cpp
+++ b/SofaKernel/modules/SofaHelper/src/sofa/helper/system/FileRepository.cpp
@@ -345,6 +345,37 @@ std::string FileRepository::relativeToPath(std::string path, std::string refPath
     return path;
 }
 
+// Zykl.io begin
+std::string FileRepository::getTempPath() const
+{
+    std::string retval;
+
+#ifdef _WIN32
+    TCHAR wbuf [255];
+    GetTempPath (255, wbuf);
+
+    // conversion
+    char buf[255] = {0};
+    if ( !WideCharToMultiByte(CP_UTF8, 0, wbuf, -1, buf, 255, 0, 0) ) {
+        msg_error(this) << "widechar to multibyte encoding failed";
+    }
+    retval = std::string(buf);
+#else
+
+#ifdef __linux__
+    char const *folder = getenv("TMPDIR");
+    if (folder == 0)
+        folder = "/tmp/";
+    retval = std::string(folder);
+
+#else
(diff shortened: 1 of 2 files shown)


faichele:
The added function is used in a few places in plugins I would like to make available in addition to the ROS connector. It is not specifically tied to the multithreaded collision pipeline, but it is potentially useful independent of the relation to the plugins in question, and easy to patch into a core class as it introduces no further dependencies on additional modifications.

jnbrunet:
Hi @faichele ,

Thank you for this PR.

However, now that SOFA is using c++17, I think we should use a standard way of doing this:

https://en.cppreference.com/w/cpp/filesystem/temp_directory_path

This is compatible with all OS (given a c++17 compliant compiler, which is now required by SOFA).

guparan:
I chose to stick to boost::filesystem since std::filesystem is not present yet on some of our minimal supported compilers: GCC 7, Clang 5, AppleClang 9.1.0 (Xcode 9.4.1)

fredroy:
> 
> 
> I chose to stick to boost::filesystem since std::filesystem is not present yet on some of our minimal supported compilers: GCC 7, Clang 5, AppleClang 9.1.0 (Xcode 9.4.1)

Actually 💁‍♂️ for those old versions of clang and gcc, filesystem is in "experimental" so theoretically it could work with 
```
#include <experimental/filesystem>
using fs = std::experimetal::filesystem
``` 
But if the boost version is working for now, let's not make the code more complicated... 
(even if I am normally against using boost when possible 😛)

guparan:
> 
> 
> > I chose to stick to boost::filesystem since std::filesystem is not present yet on some of our minimal supported compilers: GCC 7, Clang 5, AppleClang 9.1.0 (Xcode 9.4.1)
> 
> Actually tipping_hand_man for those old versions of clang and gcc, filesystem is in "experimental" so theoretically it could work with
> 
> ```
> #include <experimental/filesystem>
> using fs = std::experimetal::filesystem
> ```

This would force us to handle both cases and add a specific compiler options when using experimental::filesystem.
At the end, it's a lot of changes for something that should be redone in a couple years (when all our minimal versions we will support std::filesystem).

That's why I prefer the boost::filesystem solution, keeping in mind that when std::filesystem covers all our compilers we will be able to switch all boost::filesystem (there are already a few in SOFA) to std::filesystem :+1:

fredroy:
Agreed ! 💪

PR #1369, opened 2020-05-07
[SofaConstraint] Set the use of integration factor true by default

hugtalbot:
This should avoid confusion in constraint resolution when mixing integration factors.
Let's see how this affect the CI.

This affects the UncoupledConstraintCorrection : the flag _useOdeSolverIntegrationFactors_ is now true by default. This enforces to respect values returned by the ODESolver.

See this [forum topic](https://www.sofa-framework.org/community/forum/topic/bilateralinteractionconstraint-reduces-gravity-by-100x/)

______________________________________________________

This PR: 
- [ ] builds with SUCCESS for all platforms on the CI.
- [ ] does not generate new warnings.
- [ ] does not generate new unit test failures.
- [ ] does not generate new scene test failures.
- [ ] does not break API compatibility.
- [ ] is more than 1 week old (or has fast-merge label).

**Reviewers will merge only if all these checks are true.**

PR #1386, opened 2020-05-15
[SofaConstraint] Fix test further to #1369

hugtalbot:
A test is run with UncoupledCC alone without ODE solvers so no solver found, error, bad.
This fixes it!

This PR: 


epernod:
[ci-build][with-all-tests]

PR #1393, opened 2020-05-21
[Geomagic] Fix compilation

hugtalbot:
Two thing:
- @epernod some of your code includes changes not merged yet on component**S**tate
- ```Sleep()``` on Linux is not working

This PR: 

### DIFF
--- a/applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
+++ b/applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
@@ -334,7 +334,7 @@ void GeomagicDriver::initDevice()
     }
     
     // 2.6- Need to wait several ms for the scheduler to be well launched and retrieving correct device information before updating information on the SOFA side.
-    std::this_thread::sleep_for(42);
+    std::this_thread::sleep_for(std::chrono::milliseconds(42));
     updatePosition();
 
     sofa::core::objectmodel::BaseObject::d_componentstate.setValue(sofa::core::objectmodel::ComponentState::Valid);
--- a/applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
+++ b/applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
@@ -28,6 +28,8 @@
 
 #include <sofa/core/visual/VisualParams.h>
 #include <Geomagic/GeomagicVisualModel.h>
+#include <thread>
+#include <chrono>
 
 namespace sofa::component::controller
 {
--- a/applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
+++ b/applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
@@ -173,7 +173,7 @@ GeomagicDriver::GeomagicDriver()
     this->f_listening.setValue(true);
     m_forceFeedback = nullptr;
     m_GeomagicVisualModel = std::make_unique<GeomagicVisualModel>();
-    sofa::core::objectmodel::BaseObject::d_componentState.setValue(sofa::core::objectmodel::ComponentState::Loading);
+    sofa::core::objectmodel::BaseObject::d_componentstate.setValue(sofa::core::objectmodel::ComponentState::Loading);
 }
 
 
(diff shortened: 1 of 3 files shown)


applications/plugins/Geomagic/src/Geomagic/GeomagicEmulator.cpp
```diff
@@ -167,7 +175,11 @@ void GeomagicEmulator::initDevice()
     }
 
     // 2.6- Need to wait several ms for the scheduler to be well launched and retrieving correct device information before updating information on the SOFA side.
-    Sleep(42);
+#ifndef WIN32
```
fredroy: Please use 
` std::this_thread::sleep_for(std::chrono::milliseconds(x)); `
instead of sleep/Sleep and the ifdefs

hugtalbot: Done @fredroy
Should this later be propagated to all SOFA?
What about ctime::sleep ?

fredroy: CTime was done because of a lack of time API standard; now that  there is a stdlib for that, I would advocate to remove CTime altogether...

applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
```diff
@@ -334,11 +334,7 @@ void GeomagicDriver::initDevice()
     }
     
     // 2.6- Need to wait several ms for the scheduler to be well launched and retrieving correct device information before updating information on the SOFA side.
-#ifndef WIN32
-        usleep(42);
-#else
-        Sleep(42);
-#endif
+    std::this_thread::sleep_for(42);
```
fredroy: You are sure that std::this_thread::sleep_for(42) will sleep for 42 ms?

hugtalbot: nope, sorry! missing `std::chrono::milliseconds`

applications/plugins/Geomagic/src/Geomagic/GeomagicEmulator.cpp
hugtalbot: OK I will work on a more general PR then

applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
hugtalbot: done

epernod: in fact this will sleep at least for 43ms... ;)  I discover that sleep_for method take more than 1 ms to act....

applications/plugins/Geomagic/src/Geomagic/GeomagicEmulator.cpp
epernod: I will be interested to see the changes because right now I'm using CTime tick mechanism to be more precise than method sleep_for for example. See my other comment below. 
I probably missed some methods.

epernod:
I can't test the behavior, but I assume the std lib method works and we don't need to be precise to the ms here.
I Just pushed a compilation fix on windows.

applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
```diff
@@ -173,7 +175,7 @@ GeomagicDriver::GeomagicDriver()
     this->f_listening.setValue(true);
     m_forceFeedback = nullptr;
     m_GeomagicVisualModel = std::make_unique<GeomagicVisualModel>();
-    sofa::core::objectmodel::BaseObject::d_componentState.setValue(sofa::core::objectmodel::ComponentState::Loading);
+    sofa::core::objectmodel::BaseObject::d_componentstate.setValue(sofa::core::objectmodel::ComponentState::Loading);
```
epernod, line 178: Sorry, too many changes in my dev branch. 
I'll try to finish the PR regarding that changes for the next meeting to avoid moving backward here.

applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
```diff
@@ -334,10 +336,10 @@ void GeomagicDriver::initDevice()
     }
     
     // 2.6- Need to wait several ms for the scheduler to be well launched and retrieving correct device information before updating information on the SOFA side.
-    Sleep(42);        
+    std::this_thread::sleep_for(std::chrono::milliseconds(42));
```
epernod, line 339: I can't test the behavior, but I assume the std lib method works and we don't need to be precise to the ms here.
I Just pushed a compilation fix on windows.

applications/plugins/Geomagic/src/Geomagic/GeomagicEmulator.cpp
hugtalbot: See #1394

PR #1396, opened 2020-05-27
[headlessRecorder] Update headless recorder to use the new ffmpeg recorder

guparan:
This is a redo and fix #919 based on current master.

To be tested by @Younesssss :wink: 

This PR: 


adagolodjo:
It works now :+1: .

PR #1399, opened 2020-05-27
[SofaKernel] Totally remove the macro CHECK_TOPOLOGY from BaseMeshTopology

epernod:
The use of the macro CHECK_TOPOLOGY  has been removed from the code through several PR. This one totally remove it.

Remove duplicate call to createEdgeSetArray from Modifier when adding an edge to the topology. As all topology buffers are init at start, if hasEdges return 0, it means the topology is really empty not that the topology is not init.
This behavior has been set in the PR #967

This PR: 


epernod:
[ci-build][with-all-tests]

guparan:
I guess all the scene errors like "method must be implemented by a child topology" on CI are normal? Do you plan to fix them in another PR @epernod ?

epernod:
@guparan no, I will try to investigate that before tomorrow meeting :)

PR #1401, opened 2020-05-29
[SofaMiscFem] Fix add computeBBox method to the hyperelasticFEM

sergeiNikolaev:
This is a fix to update a bound box for a hyperelastic force field. Without updating the rotation and translation of a virtual camera works incorrectly. This method is a copy of a bound box computation from TetrahedronFEMForceField.

This PR: 

### DIFF
--- a/modules/SofaMiscFem/TetrahedronHyperelasticityFEMForceField.inl
+++ b/modules/SofaMiscFem/TetrahedronHyperelasticityFEMForceField.inl
@@ -734,6 +734,33 @@ void TetrahedronHyperelasticityFEMForceField<DataTypes>::saveMesh( const char *f
 	fclose( file );
 }
 
+
+
+template<class DataTypes>
+void TetrahedronHyperelasticityFEMForceField<DataTypes>::computeBBox(const core::ExecParams*, bool onlyVisible)
+{
+    if( !onlyVisible ) return;
+
+    helper::ReadAccessor<DataVecCoord> x = this->mstate->read(core::VecCoordId::position());
+
+    static const Real max_real = std::numeric_limits<Real>::max();
+    static const Real min_real = std::numeric_limits<Real>::lowest();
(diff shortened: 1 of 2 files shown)


modules/SofaMiscFem/TetrahedronHyperelasticityFEMForceField.inl
```diff
@@ -734,6 +734,33 @@ void TetrahedronHyperelasticityFEMForceField<DataTypes>::saveMesh( const char *f
 	fclose( file );
 }
 
+
+
+template<class DataTypes>
+void TetrahedronHyperelasticityFEMForceField<DataTypes>::computeBBox(const core::ExecParams*, bool onlyVisible)
+{
+    if( !onlyVisible ) return;
+
```
hugtalbot, line 743: Could you please add this check:
```cpp
if (!this->mstate) return;
```

hugtalbot (changes requested):

modules/SofaMiscFem/TetrahedronHyperelasticityFEMForceField.inl
hugtalbot: Hi @sergeiNikolaev 
I suggested one change in your PR.
I don't know why I can not do myself thus directly from the GitHub GUI.

sergeiNikolaev: Hi @hugtalbot.

Yes, I have seen this.
Sorry, I was busy during the last days. I will do it right now.

sergeiNikolaev: Done

hugtalbot: Great thanks Sergei !

hugtalbot (approved):

PR #1404, opened 2020-06-02
[SofaGeneralEngine] Refresh MeshBarycentricMapperEngine

fredroy:
Because why not.

Doing the same as before but with proper guidelines and stuff.
Breaking ⛔ because of the change in the fields name, but this engine is not used anywhere 🙋‍♂️

This PR: 

### DIFF
--- a/modules/SofaGeneralEngine/MeshBarycentricMapperEngine.cpp
+++ b/modules/SofaGeneralEngine/MeshBarycentricMapperEngine.cpp
@@ -20,20 +20,14 @@
 * Contact information: contact@sofa-framework.org                             *
 ******************************************************************************/
 #define SOFA_COMPONENT_ENGINE_MESHBARYCENTRICMAPPERENGINE_CPP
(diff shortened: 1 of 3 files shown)


modules/SofaGeneralEngine/MeshBarycentricMapperEngine.cpp
```diff
@@ -20,20 +20,14 @@
...
 #include <sofa/defaulttype/RigidTypes.h>
 
-namespace sofa
-{
-
-namespace component
-{
-
-namespace engine
+namespace sofa::component::engine
```
marques-bruno, line 29: I like it, but didn't we say we'd keep clear of namespace concatenation & pragmas in sofa modules & only apply them to plugins?

hugtalbot: not that I remember
Fine to me this way

PR #1407, opened 2020-06-05
[SofaCore] ADD directory DataFileNames

marques-bruno:
This PR adds an attribute pathType (m_pathType) in DataFileName and DataFileNameVector to be able to easily identify whether a component holding a DataFileName accepts files, directories or both. This is useful to provide custom fileDialogs in user interfaces such as runSofa2, give more information about the type of values a dataFileName takes, and  provide better error messages.

To use it: in the ctor of the component, on a DataFileName datafield, call setPathType(<PathType>) (<PathType> being an enum value of either FILE, DIRECTORY or BOTH. Default value is FILE)

The PathType of a DataFileName can be queried by calling getPathType().

In addition, checks are made within the updatePath() method of DataFileName, called when the value changes, to guarantee that the value set is consistent with the defined path type.

Since the default pathtype is set to FILE, there are only few components that are directly affected by the pull request (AddResourceRepository is one of them (takes only directories), MeshExporters have a DataFileNameVector which can accept both directories and files. Those 2 components had to be modified to behave properly, and I've been able to identify them thanks to the existing unit tests & scene tests that we compile on the CI. There's no guarantee that there isn't any untested component that should be updated too, but the fix is easy (setPathType(WHATEVER) in the ctor of the component). Same goes for pluginized components of course.

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/DataFileName.cpp
+++ b/SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/DataFileName.cpp
@@ -20,6 +20,7 @@
 * Contact information: contact@sofa-framework.org                             *
 ******************************************************************************/
 #include <sofa/helper/system/FileRepository.h>
+#include <sofa/helper/system/FileSystem.h>
 #include <sofa/core/objectmodel/DataFileName.h>
 #include <sofa/core/objectmodel/Base.h>
 
@@ -33,6 +34,8 @@ namespace core
 
 namespace objectmodel
 {
+
+namespace fs = sofa::helper::system;
 
 DataFileNameVector::~DataFileNameVector()
 {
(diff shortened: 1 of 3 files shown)


damienmarchal (commented):
You may add a consistency check if the given path is actually a dir when it is supposed to be. I'm ok with the rest of the code.

marques-bruno:
> You may add a consistency check if the given path is actually a dir when it is supposed to be. I'm ok with the rest of the code.

I've updated the code to check for path types according to isDirectory.
I also propagated the mechanism to DataFileNameVectors (if one of the paths in the vector is not of the right type, I simply set it to "". Dunno if thats the right way but it seems right)

marques-bruno:
[ci-build][with-all-tests]

marques-bruno:
[ci-build][with-all-tests]

hugtalbot (approved):

PR #1414, opened 2020-06-11
[ManifoldTopologies] Update the license in init.cpp

hugtalbot:
and I also propose to update the APIChange file .. 
I did two in one (but separate commits) since it's super minor!

This PR: 


fredroy:
I don't mind the change in the API string but I would suggest to not squash this PR then

guparan (approved):

PR #1417, opened 2020-06-11
[SofaKernel] Fix hexahedra detection in BoxROI

andrea-mendizabal:
A break is present in the ```doUpdate``` function making the detection of all hexahedra fail

This PR: 


epernod (approved):

epernod:
Nice one, this is indeed a bug. Behavior was changed in PR #108 and this loop was forgotten.

PR #1421, opened 2020-06-15
[all] Fix warnings due to visibility attribute

hugtalbot:
Just not 100% sure for LCPcalc.h  and OglModelPolicy.h 
Else we should get back to original warning number.

This PR: 


damienmarchal (approved):
Hi Hugo,

Thanks for the PR. You are right about the typedefs...they does not need visilibty declaration as these lines are not "something" that results in the binary files. 
Everything is ok for me.

marques-bruno (approved):

PR #1423, opened 2020-06-16
[SofaOpenGLVisual] Use updateCallback to update OglModel's internal buffers

marques-bruno:
Needs #1408 
OglModel recomputes its internal data on topology change.

There's 2 things I need review for here:
1. which initialization methods should be called to properly update an OglModel? Currently, I'm calling more or less the whole thing: delete buffers, then init / reinit / initVisual & updateBuffers. Not sure all of them are necessary
2. Which datafields should be taken into consideration for a total reinit:
 Ideally I'd like to trigger the full update of the OglModel only when a change in the loader triggers the recomputation: positions can vary during simulation and shouldn't trigger a reload of the component.
Edges / triangles & other topological info change when the loader changes geometry, but could also change during a topological change for instance, and in that 2nd case shouldn't trigger a reload from the loader, whose topology isn't up-to-date anymore.

This component needs heavy restructuring I believe if we want to handle proper data update using callbacks, and might be a good example for other components if we do it right

This PR: 

### DIFF
--- a/modules/SofaOpenglVisual/src/SofaOpenglVisual/OglModel.cpp
+++ b/modules/SofaOpenglVisual/src/SofaOpenglVisual/OglModel.cpp
@@ -95,9 +95,23 @@ OglModel::OglModel()
     primitiveTypeOptions->setNames(4, "DEFAULT", "LINES_ADJACENCY", "PATCHES", "POINTS");
     primitiveTypeOptions->setSelectedItem(0);
     primitiveType.endEdit();
+
+   addUpdateCallback("topology", {&m_edges, &m_quads, &m_triangles},
+                     [&]() -> sofa::core::objectmodel::ComponentState
+   {
+       m_componentstate = sofa::core::objectmodel::ComponentState::Loading;
+       clearLoggedMessages();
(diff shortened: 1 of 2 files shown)


fredroy:
Hum, you would be better to have a look at VisualModelImpl, which is the real location managing the data, especially updateVisual().

fredroy:
And you have to make sure that the GL buffers are created/updated/modified while having an OpenGL context (so either in initVisual()/updateVisual()/drawVisual()...)

modules/SofaOpenglVisual/src/SofaOpenglVisual/OglModel.cpp
```diff
@@ -370,14 +389,16 @@ void OglModel::internalDraw(const core::visual::VisualParams* vparams, bool tran
     if (vparams->displayFlags().getShowWireFrame())
         glPolygonMode(GL_FRONT_AND_BACK, GL_LINE);
 
+
+
+
     const VecCoord& vertices = this->getVertices();
```
hugtalbot, line 395: Very nice !

PR #1360, opened 2020-05-01
[ColladaSceneLoader][SofaAssimp] Move ColladaSceneLoader plugin content into SofaAssimp plugin

epernod:
Remove ColladaSceneLoader plugin and its embedded old version of assimp. 
- Move class, doc and examples inside SofaAssimp plugin.
- Merge the CMakeLists and update the findAssimp with a the version provided by Assimp 5.0.
- Ensure scenes of both plugins are working with up to date version of assimp

SofaAssimp plugin still need the pluginization refresh. To be done in another PR.

______________________________________________________

This PR: 
- [ ] builds with SUCCESS for all platforms on the CI.
- [ ] does not generate new warnings.
- [ ] does not generate new unit test failures.
- [ ] does not generate new scene test failures.
- [ ] does not break API compatibility.
- [ ] is more than 1 week old (or has fast-merge label).

**Reviewers will merge only if all these checks are true.**

PR #1424, opened 2020-06-16
[SofaGuiQt] Fix node graph

marques-bruno:
In order for SofaGuiQt to find NodeEditor a custom FindNodeEditorConfig.cmake was created in cmake/Modules but this file was obfuscating the real NodeEditorConfig.cmake that is packaged with the NodeEditor library when trying to find_package it in external plugins. I modified the CMakeLists.txt in sofa/gui/qt/ to use the proper find_package and link correctly to the NodeEditor::nodes target.

This PR also fixes some unintentional renamings that were causing compile issue when SOFAGUIQT_ENABLE_NODEGRAPH is set to ON (it would be nice to be able to test this on the CI, or at least to build with  SOFAGUIQT_ENABLE_NODEGRAPH=ON so that this doesn't happen again.

This PR: 


hugtalbot (approved):

applications/sofa/gui/qt/CMakeLists.txt
```diff
@@ -294,12 +294,8 @@ if(NOT SOFA_NO_OPENGL)
...
 endif()
 
-if(NODEEDITOR_FOUND)
-    target_link_libraries(${PROJECT_NAME} PUBLIC ${NODEEDITOR_LIBRARY})
-    if (SOFA_BUILD_RELEASE_PACKAGE OR CMAKE_SYSTEM_NAME STREQUAL Windows)
-        sofa_install_libraries(PATHS ${NODEEDITOR_LIBRARY})
-    endif()
-    target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${NODEEDITOR_INCLUDE_PATH}>")
+if(SOFAGUIQT_ENABLE_NODEGRAPH)
+    target_link_libraries(${PROJECT_NAME} PUBLIC nodes)
```
guparan: NodeEditor::nodes

marques-bruno: @guparan it works with NodeEditor::nodes. I corrected the PR, I let you mark it as ready if you don't object

epernod (changes requested):
Here there is the same problem as for Eigen or for Assimp in pending PR #1360 
The custom find.cmake was there for windows as the findpackage of Node editor is not found.
you should do like for Eigen, first search for the official findpackage and if not found use the custom one.

guparan (approved):

marques-bruno:
@guparan thx for your motifs :)

Can we label this PR as "ready"?

Issue #1442, opened 2020-07-30
Moving plugins out in external sofa-framework repositories (STC#9)


hugtalbot:
As decided at the last STC, a serie of plugins will be moved out from the SOFA repository and saved into their own repository.
List of related plugins [available here](https://docs.google.com/spreadsheets/d/1FVQN4rghI1SSYuhT1Y3L5fUZdL7XXPETgQ1GQ28KlD8/edit#gid=0).

This will help in the process of deprecation and pluginization in SOFA.

- Extraction process is clearly explained in this [documentation](https://docs.github.com/en/github/using-git/splitting-a-subfolder-out-into-a-new-repository)
- Remaining: making a pull-request in SOFA to
  - remove all plugin files
  - add a plugin folder containing one file `ExternalProjectConfig.cmake.in`
  - edit _applications/plugin/CMakeLists.txt_ to replace `sofa_add_plugin` by `sofa_add_plugin_external`
- Finalize the GitHub repository:
  - add a description
  - add reviewers (sofa-framework/reviewers, sofa-framework/consortium-staff and appropriate watcher)
  - add the "plugin" topic (flag)

______________________________________________________

Suggested labels:

alxbilger:
@bakpaul @hugtalbot @olivier-roussel We discussed last week that `SofaNewMat` and probably `SofaDistanceGrid` must be externalized, according to https://github.com/sofa-framework/sofa/issues/4221

PR #1443, opened 2020-07-30
[InvertibleFVM] Externalize the plugin

hugtalbot:
First plugin to be moved out, as detailed in issue #1442 
Plugin is now located here: https://github.com/sofa-framework/InvertibleFVM

All feedback are more than welcome.

This PR: 


guparan (changes requested):
You should not mess with `sofa_add_plugin` calls order in applications/plugins/CMakeLists.txt, it's this way because of dependencies between plugins.
TODO: reset this file and just replace `sofa_add_plugin(InvertibleFVM InvertibleFVM)` by `sofa_add_plugin_external(InvertibleFVM InvertibleFVM)`

guparan:
[ci-build][with-all-tests]

PR #1444, opened 2020-07-30
Micro PR to make a function virtual in FrictionContact.

olivier-goury:
With the purpose to override it in a child class in a plugin.

This PR: 


hugtalbot (approved):

epernod:
do you think other methods could be overridden in the future or by another derivative implementation? 
It could be the good opportunity to turn other methods as virtual as well, maybe.

olivier-goury:
All the other methods in FrictionContact are virtual, except `setInteractionTags`, I don't see a reason to "virtualize" it for the moment...

PR #1445, opened 2020-08-03
[SofaCore] FIX simu unload crash caused by missing checks on slaves ptrs

marques-bruno:
Got some crashes on some simulations, when trying to unload them.
Here's the fix

This PR: 


guparan (approved):

guparan:
[ci-build][with-all-tests]

PR #1447, opened 2020-08-12
[BaseConstraintCorrection][LinearSolver] add a virtual function to split some computation

epernod:
This PR is a cherry pick of the changes proposed in PR #869 

Here is the initial description of the PR by @fspadoni 

---------

When solving constraints using a BaseConstraintCorrection component the function 
addComplianceInConstraintSpace compute the compliance matrix and add the contribution of the compliance matrix to the constraint solver matrix. 

the goal is to split the computation of the compliance matrix (local for each deformable object) and the sum of the compliance matrix to the  constraint solver (global) matrix in two different functions.

- BaseConstraintCorrection::computeComplianceInConstraintSpace:
 should be used to compute the compliance. This computation is independent and can be easily parallelized

-  BaseConstraintCorrection::addComplianceInConstraintSpace:
 should be used only to add the contribution of the compliance matrix to the constraint solver matrix. 

This change is propagated to the LinearSolver class hierarchy because a ConstraintCorrection component might delegate the computation of the compliance matrix to the the LinearSolver

This PR: 


epernod:
[ci-build][with-all-tests]

epernod:
I cherry picked the changes of #869 and merge it carefully with current master branch.
Could someone double check this PR diff compared to the diff of #869. If it looks the same we could at least close the #869 and review only this one.

modules/SofaConstraint/GenericConstraintCorrection.h
```diff
@@ -49,6 +49,8 @@ class SOFA_CONSTRAINT_API GenericConstraintCorrection : public core::behavior::B
 
     void computeMotionCorrectionFromLambda(const core::ConstraintParams* cparams, core::MultiVecDerivId dx, const defaulttype::BaseVector * lambda) override;
 
+    void computeComplianceInConstraintSpace(const core::ConstraintParams *cparams, defaulttype::BaseMatrix* W) override;
+
     void addComplianceInConstraintSpace(const core::ConstraintParams *cparams, defaulttype::BaseMatrix* W) override;
```
hugtalbot, line 54: PR #869 seems to virtualize many function here, but I guessed you set as virtual only the useful one. Is it correct?

modules/SofaConstraint/PrecomputedConstraintCorrection.h
```diff
@@ -80,6 +80,8 @@ class PrecomputedConstraintCorrection : public sofa::core::behavior::ConstraintC
 public:
     void bwdInit() override;
 
+    virtual void computeComplianceInConstraintSpace(const core::ConstraintParams *cparams, defaulttype::BaseMatrix* W) override {}
+
     void addComplianceInConstraintSpace(const sofa::core::ConstraintParams *cparams, sofa::defaulttype::BaseMatrix* W) override;
```
hugtalbot, line 85: Many virtualization here as well

modules/SofaConstraint/GenericConstraintCorrection.h
epernod: in fact before the #869 I think several methods were already virtual and nearly all method were virtual at the end of #869 but since that time I think some "clang clean PR" have removed many ```virtual``` keyword .
So when I rebased this this PR it was really a mess to know who should be virtual or not...

Maybe not the best way to act but I would suggest to concentrate in making the suggested split of method works without loss of precision nor performance and we will unify the virtual later.

PR #1508, opened 2020-09-30
[SofaQtGui] Restore GraphWidget for Momentum and Energy using QtCharts instead of Qwt

epernod:
- Restore QGraphStatWidget, QMomentumWidget and QEnergyWidget which have been removed by error in a PR #750 
- Replace use of Qwt by QtCharts

graph_energy_new
graph_momentum_new

This PR: 

### DIFF
--- a/modules/SofaGuiQt/src/sofa/gui/qt/ModifyObject.cpp
+++ b/modules/SofaGuiQt/src/sofa/gui/qt/ModifyObject.cpp
@@ -275,7 +275,7 @@ void ModifyObject::createDialog(core::objectmodel::Base* base)
             }
         }
 
-#ifdef SOFAGUIQT_HAVE_QT5_CHARTS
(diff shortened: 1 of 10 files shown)


epernod:
[ci-build][with-all-tests]

modules/SofaGuiQt/src/sofa/gui/qt/ModifyObject.cpp
```diff
@@ -275,6 +275,29 @@ void ModifyObject::createDialog(core::objectmodel::Base* base)
             }
         }
 
+#ifdef SOFAGUIQT_HAVE_QT5_CHARTS
```
guparan: SOFAGUIQT_HAVE_QT5_CHARTS is **always defined** (0 or 1).
You must use `#if`.

guparan (changes requested):

PR #1514, opened 2020-10-01
[All] Change index_type from size_t to uint

fredroy:
#1453 was about to make uniform the type representing indices. But this "universal" index_type was a std::size_t which is a bit too much for our simulations (and breaks a lot of memory-dependent code such as CUDA things)
This PR change this type to unsigned int;  and fixes some forgotten index_type mistakes by the way.

This PR: 


fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):

epernod (approved):

PR #1516, opened 2020-10-03
[SofaGeneralEngine] Fix BarycentricMapperEngine parse() function

fredroy:
Bug introduced in #1487, with an infinite loop when calling parse()
(Fix timeout with MeshBarycentricMapperEngine.scn on the CI)

This PR: 

### DIFF
--- a/modules/SofaGeneralEngine/MeshBarycentricMapperEngine.h
+++ b/modules/SofaGeneralEngine/MeshBarycentricMapperEngine.h
@@ -83,7 +83,7 @@ public:
         {
             msg_warning() << "input data 'InputMeshName' changed for 'topology', please update your scene (see PR#1487)";
         }
-        MeshBarycentricMapperEngine::parse(arg);
+        core::DataEngine::parse(arg);
     }
 
 private:


fredroy:
[ci-build][with-all-tests]

modules/SofaGeneralEngine/MeshBarycentricMapperEngine.h
```diff
@@ -83,7 +83,7 @@ class MeshBarycentricMapperEngine : public core::DataEngine
         {
             msg_warning() << "input data 'InputMeshName' changed for 'topology', please update your scene (see PR#1487)";
         }
-        MeshBarycentricMapperEngine::parse(arg);
+        core::DataEngine::parse(arg);
```
epernod, line 86: does it make a diff to do that call or ```Inherit1::parse(arg);``` ?

fredroy: Inherit1 ?
You mean to make a typedef Inherit1 = core::DataEngine and calling it ?

epernod: my bad I thought it was a smarter implementation. 
ok for me

epernod (approved):

PR #1518, opened 2020-10-05
[SofaHelper] Quater does not depend on Vec and Mat from defaulttype anymore

fredroy:
+ some cleaning and hard coded epsilon removals.

Should not be breaking, EXCEPT: 
 - for code using Vec/Mat without including them (👎)
 - rare code depending on Vec3 as parameters 🧨

This PR: 

### DIFF
--- a/SofaKernel/modules/SofaHelper/src/sofa/helper/Quater.h
+++ b/SofaKernel/modules/SofaHelper/src/sofa/helper/Quater.h
@@ -22,12 +22,14 @@
 #ifndef SOFA_HELPER_QUATER_H
 #define SOFA_HELPER_QUATER_H
 
+#include <sofa/helper/config.h>
+
 #include <sofa/defaulttype/Vec.h>
 #include <sofa/defaulttype/Mat.h>
 #include <cmath>
 #include <cassert>
 #include <iostream>
-#include <sofa/helper/config.h>
+#include <array>
 
 namespace sofa
 {
@@ -39,12 +41,12 @@ template<class Real>
 class SOFA_HELPER_API Quater
 {
 private:
-    Real _q[4];
+    std::array<Real, 4> _q;
 
 public:
-
     typedef Real value_type;
     typedef std::size_t size_type;
+    using Vector3 = std::array<Real, 3>;
 
     Quater();
     ~Quater();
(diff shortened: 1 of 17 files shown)


SofaKernel/modules/SofaHelper/src/sofa/helper/Quater.h
```diff
@@ -101,10 +99,68 @@ class SOFA_HELPER_API Quater
         _q[3]=1.0;
     }
 
-    void fromFrame(defaulttype::Vec<3,Real>& x, defaulttype::Vec<3,Real>&y, defaulttype::Vec<3,Real>&z);
+    void fromFrame(Vector3& x, Vector3&y, Vector3&z);
```
damienmarchal, line 78: I think that we should put the code in a .inl, have two explicit extern template on double and float and in the .cpp the two float/double instanciation.

damienmarchal:
Thanks @fredroy  for the PR, good to see there is more and more cleaning on the core sofa. 
I think we should keep the code definition in .inl  instead of having everything in a .h. 
This is especially true for Quaternion as in 99% of the sofa code base we only have float and double specialization.

SofaKernel/modules/SofaHelper/src/sofa/helper/fixed_array.h
```diff
@@ -278,7 +278,7 @@ class fixed_array
     }
 
     // size is constant
-    static size_type size()
+    static constexpr size_type size()
```
damienmarchal, line 281: if this is a constexpr is the static usefull anymore ?

fredroy: If I remove the static, people can only do <instanciation>.size()

SofaKernel/modules/SofaHelper/src/sofa/helper/Quater.h
fredroy: I can do it for some functions (only templated on float/double) but it wont be possible to do it with one other template (typically the matrix type or the vec type)

fredroy:
> > I think we should keep the code definition in .inl instead of having everything in a .h.
> This is especially true for Quaternion as in 99% of the sofa code base we only have float and double specialization.

Some functions needs the vector or the matrix type (99.9% of the cases, from defaulttype), as such it wont be possible to do an explicit instanciation in the cpp. It could be possible to do it later (in a special file in SofaDefaulttype?) but I think it will be too much a mess

damienmarchal:
About the templated methods.
I never payed attention but at first sight this does not seems the way to go. Despite looking more general...it also allows mixing different precisions in calculus in a subtile and implicit way. 

So I would recommand having Quatf with float only, Quatd with double only signatures, and thus drop the templated method in Quat :)

applications/plugins/CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.12)
 find_package(SofaFramework)
 
 sofa_add_subdirectory_external(SofaHighOrder SofaHighOrder)
-sofa_add_subdirectory_external(SofaPython3 SofaPython3)
+#sofa_add_subdirectory_external(SofaPython3 SofaPython3)
```
guparan: ???

fredroy: Collateral damage 😅 (will fix it along onther things) 
but I could not just only disable Python3 compilation in the cmake by the way

fredroy:
Not relevant anymore due to Sofa.Type work #1527

PR #1532, opened 2020-10-12
[SofaCarving] Fix method doCarve should be called at AnimateEndEvent

epernod:
Method doCarve in CarcingManager should be called at AnimateEndEvent and not at CollisionEndEvent to avoid any topology synchronization error during step.

This PR: 


epernod:
[ci-build][with-all-tests]

epernod:
[ci-build][with-all-tests]

epernod:
[ci-build][with-all-tests]

PR #1550, opened 2020-10-18
[plugins] Fix CUDA includes of SofaSphFluid

courtecuisse:
Fix cuda include compilation

This PR: 

### DIFF
--- a/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaSpatialGridContainer.h
+++ b/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaSpatialGridContainer.h
@@ -34,7 +34,7 @@
 #ifndef SOFA_GPU_CUDA_CUDASPATIALGRIDCONTAINER_H
 #define SOFA_GPU_CUDA_CUDASPATIALGRIDCONTAINER_H
 
-#include <SofaSphFluid/SpatialGridContainer.h>
+#include <SofaSphFluid/src/SofaSphFluid/SpatialGridContainer.h>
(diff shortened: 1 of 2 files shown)


applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaSpatialGridContainer.h
```diff
@@ -34,7 +34,7 @@
 #ifndef SOFA_GPU_CUDA_CUDASPATIALGRIDCONTAINER_H
 #define SOFA_GPU_CUDA_CUDASPATIALGRIDCONTAINER_H
 
-#include <SofaSphFluid/SpatialGridContainer.h>
+#include <SofaSphFluid/src/SofaSphFluid/SpatialGridContainer.h>
```
guparan, line 37: Needing to do this is not normal.

To use SofaSphFluid headers, you must have `find_package(SofaSphFluid)` and `target_link_library(${PROJECT_NAME} PUBLIC SofaSphFluid)` in your CMakeLists.

If it does not work, the problem comes from SofaSphFluid.

guparan (changes requested):

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaSpatialGridContainer.h
guparan: Here is a full example of how to handle a soft dependency to SofaSphFluid: https://github.com/sofa-framework/sofa/search?q=SofaSphFluid+OpenCL

PR #1577, opened 2020-10-30
[SofaBaseTopology] Fix ambiguity causing compilation error with MSVC

alxbilger:
The error pops up in debug mode


fredroy:
Make sense that this error does not appear on the CI, as the assert() is not processed by the compiler in other mode than Debug (and I suppose the CI compiles in Release mode with MSVC)
and the ambiguity comes from the fact there are two invalidID things, one defined in the sofa namespace and one other defined in the defaulttype namespace for compatibility.

fredroy (approved):

PR #1587, opened 2020-11-03
[SofaGeneralAnimationLoop] Fix mechanical matrix mapper

olivier-goury:
Fixes a bug in the component, that happens when using a child class that has its own implementation of _accumulateJacobiansOptimized_ and leads to an undefined J1 and J2 matrices after initialization, thus leading to a crash when setting J1eig to match its size.


modules/SofaGeneralAnimationLoop/src/SofaGeneralAnimationLoop/MechanicalMatrixMapper.inl
```diff
@@ -381,17 +381,19 @@ void MechanicalMatrixMapper<DataTypes1, DataTypes2>::addKToMatrix(const Mechanic
...
     const MatrixDeriv1 &J1 = c[ms1].read()->getValue();
     const MatrixDeriv2 &J2 = c[ms2].read()->getValue();
-
+    if (this->getContext()->getTime() == 0)
+        nbColsJ1 = J1.begin().row().size()*DerivSize1;
     Eigen::SparseMatrix<double> J1eig;
     Eigen::SparseMatrix<double> J2eig;
-    J1eig.resize(K->nRow, J1.begin().row().size()*DerivSize1);
-    unsigned int nbColsJ1 = 0, nbColsJ2 = 0;
+    J1eig.resize(K->nRow, nbColsJ1);
```
hugtalbot: Could this all be done in the init / bwdInit function ?
it seems strange to check time in all addKToMatrix calls.

modules/SofaGeneralAnimationLoop/src/SofaGeneralAnimationLoop/MechanicalMatrixMapper.inl
```diff
@@ -381,17 +381,19 @@ void MechanicalMatrixMapper<DataTypes1, DataTypes2>::addKToMatrix(const Mechanic
...
 
     optimizeAndCopyMappingJacobianToEigenFormat1(J1, J1eig);
     if (bms1 != bms2)
     {
         double startTime2= (double)timer->getTime();
-        J2eig.resize(K->nRow, J2.begin().row().size()*DerivSize2);
+        if (this->getContext()->getTime() == 0)
+            nbColsJ2 = J2.begin().row().size()*DerivSize2;
+        J2eig.resize(K->nRow, nbColsJ2);
         optimizeAndCopyMappingJacobianToEigenFormat2(J2, J2eig);
```
hugtalbot: same same

hugtalbot (approved):

modules/SofaGeneralAnimationLoop/src/SofaGeneralAnimationLoop/MechanicalMatrixMapper.inl
olivier-goury: Ideally yes, but in practice, it is not easy, because I need the mparams and the matrixAccessor inputs from addKtoMatrix to set up the dimension. I made a new commit to improve that version a little bit.

olivier-goury: Oh wait, my bad, it's not a problem actually, I will commit a better version in a minute

olivier-goury:
Yes it is now!

PR #1649, opened 2020-12-04
[SofaFramework] Isolate OpenGL code into a single module (Sofa.GL)

fredroy:
Fix #1645 (needs #1646 , #1644 , #1650  )

# TL;DR
OpenGL code in the core is now enclosed in the new module Sofa.GL.
If you need to call OpenGL code, you will need to **find_package()** it and add it to **target_link_library()**.

# Description
OpenGL is spread a bit everywhere in the core, and it is difficult to bypass it if you dont need it (typically for batch mode or headless system).
The macro **SOFA_NO_OPENGL** is a really Quick and Dirty solution to this problem.
For example, people include sofa/helper/system/gl.h (or any gl-related header), even without needing OpenGL actually. 
And in this header, **SOFA_NO_OPENGL** does include (or not) the good gl.h header according to your OS. So if you enable this header, you still include gl.h, which does nothing.
This is quite nonsensical.
Moreover, if the macro is not well handled in CMake, you may have some weird things happening.

# Solution
This PR regroups all OpenGL related code in the core in a single package, needing only SofaHelper and SofaDefaulttype.
If you wish to be able to call OpenGL things, just find_package() it and link it to your library
```
find_package(Sofa.GL REQUIRED)
...
target_add_library(${Target} PUBLIC .... Sofa.GL)
```

You can even conditionnally test if OpenGL is present or not with 
```
if(Sofa.GL_FOUND)
...
endif()
```
if needed (to compile specific components in your module for example).

And even more, if you want to compile specific parts of your code with OpenGL (not a good idea but still), you can use this like an external dependency.
Example with **ColourPickingVisitor** in **SofaGuiCommon**.
CMakeLists.txt:
```
sofa_find_package(Sofa.GL) # create the cmake var SOFAGUICOMMON.GL_HAVE_SOFA.GL
```
config.h.in
```
#define SOFAGUICOMMON_HAVE_SOFA_GL @SOFAGUICOMMON_HAVE_SOFA.GL@
```
ColourPickingVisitor.cpp
```
void ColourPickingVisitor::processTriangleModel(simulation::Node * node, sofa::component::collision::TriangleCollisionModel<sofa::defaulttype::Vec3Types> * tmodel)
{
#ifdef SOFAGUICOMMON_HAVE_SOFA_GL
    using namespace sofa::core::collision;
    using namespace sofa::defaulttype;
    glDisable(GL_LIGHTING);
    ....
#endif // SOFAGUICOMMON_HAVE_SOFA_GL
```
[...]

PR #1650, opened 2020-12-05
[SofaHelper] Fix CMake config with SOFA_NO_OPENGL and dependency package (Windows)

fredroy:
Stumbled on it with #1649 

DiffusionSolver was unable to find "png.h" when SOFA_NO_OPENGL is activated.

It appears that if SOFA_NO_OPENGL is not activated, adding glew somewhat adds the dependency package directory (<sofaroot>/include) to the include_directories of SofaHelper (and the subsequent modules needing it).

But if SOFA_NO_OPENGL is not activated, this directory is not added.
DiffusionSolver needs Cimg, which adds only for himself the include directory (i.e no target_include_directories).
So CImg compiles well but the modules needing it do not. That's why DiffusionSolver cannot find "png.h"

My fix does add this include_directory to the target, but I don't know if it is the best way.

Joy of CMake 🥴

### DIFF
--- a/SofaKernel/modules/SofaHelper/CMakeLists.txt
+++ b/SofaKernel/modules/SofaHelper/CMakeLists.txt
@@ -332,6 +332,7 @@ target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${CMAKE_CUR
 target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../SofaDefaultType/src>")
 target_include_directories(${PROJECT_NAME} PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/../SofaCore/src>")
 if(CMAKE_SYSTEM_NAME STREQUAL Windows)
+    target_include_directories(${PROJECT_NAME} SYSTEM PUBLIC "$<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/include>")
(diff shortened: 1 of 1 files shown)


guparan:
Are you building DiffusionSolver out-of-tree?
`<src>/include` is supposed to by added to CMAKE_INCLUDE_PATH here: https://github.com/sofa-framework/sofa/blob/master/CMakeLists.txt#L138

I think the issue is coming from the fact that CImgPlugin does not use imported targets for its dependencies (JPEG, TIFF and PNG).
We should have something like `target_link_libraries(${PROJECT_NAME} PUBLIC PNG::PNG JPEG::JPEG TIFF::TIFF)` in CImgPlugin/CMakeLists.txt

Exporting `<src>/include` in the include directories of SofaHelper is overkill :smile:

guparan:
I have a branch were I already did the cleaning for CImgPlugin, I can PR it :+1:

fredroy:
> 
> 
> Are you building DiffusionSolver out-of-tree?

Nope
> 
> I think the issue is coming from the fact that CImgPlugin does not use imported targets for its dependencies (JPEG, TIFF and PNG).
> We should have something like `target_link_libraries(${PROJECT_NAME} PUBLIC PNG::PNG JPEG::JPEG TIFF::TIFF)` in CImgPlugin/CMakeLists.txt

A solution as well

> Exporting `<src>/include` in the include directories of SofaHelper is overkill 😄

Exporting this directory is already done if you have GLEW 👺 (+SOFA_NO_OPENGL disabled)

fredroy:
> 
> 
> I have a branch were I already did the cleaning for CImgPlugin, I can PR it 👍

Whatever works for me :F
But I guess it wont solve the problem for other people needing PNG and stuff, and without activating CimgPlugin (not really a common case indeed)

fredroy:
Superseded by #1651 , closing.

PR #1728, opened 2021-01-22
[tools] Handles Qt plugins installed from system's package managers

jnbrunet:
This will enable creating SOFA packages on Linux when Qt is installed from the package manager instead of the Qt installer. It also fixes some missing dependencies on Linux distros other than ubuntu 18.04.


tools/postinstall-fixup/linux-postinstall-fixup.sh
```diff
@@ -93,7 +106,7 @@ get-lib-deps-assoc() {
 get-lib-deps-assoc "$BUILD_DIR" "$INSTALL_DIR" "$OUTPUT_TMP"
 
 # Copy libs
-groups="libQt libpng libicu libmng libxcb libxkb libOpenGL"
+groups="libQt libpng libicu libmng libxcb libxkb libpcre2 libjbig libwebp libjpeg"
```
fredroy: removing libOpenGL, on purpose ?

jnbrunet: Yes, we need to backport https://github.com/sofa-framework/sofa/pull/1721 into master (which removes libOpenGL as it isn't working)

jnbrunet: Done in https://github.com/sofa-framework/sofa/pull/1729

tools/postinstall-fixup/linux-postinstall-fixup.sh
```diff
@@ -48,21 +48,34 @@ rm -rf "$INSTALL_DIR/plugins/platforms"
...
+# Qt installed via Fedora RPM
+if ! [ -d "$QT_PLUGIN_DIR" ] && [ -d "/usr/lib64/qt5/plugins" ] ; then
+  QT_PLUGIN_DIR="/usr/lib64/qt5/plugins"
+fi
+
+# Qt installed via Ubuntu DEB
+if ! [ -d "$QT_PLUGIN_DIR" ] && [ -d "/usr/lib/x86_64-linux-gnu/qt5/plugins" ] ; then
+  QT_PLUGIN_DIR="/usr/lib/x86_64-linux-gnu/qt5/plugins"
+fi
+
```
guparan: What is the (I guess wrong) value of QT_DIR with Fedora RPM and Ubuntu DEB?
I think it would be better to fix QT_DIR instead of adding hardcoded guesses.

jnbrunet: > What is the (I guess wrong) value of QT_DIR with Fedora RPM and Ubuntu DEB?
> I think it would be better to fix QT_DIR instead of adding hardcoded guesses.

There isn't:

```
/usr/lib64/libQt5Core.so
/usr/lib/x86_64-linux-gnu/libQt5Core.so
```

guparan: I don't understand. My question was unclear I think.
What do you get if you add `message("Qt5_DIR = ${Qt5_DIR}")` at the end of SOFA main CMakeLists?

guparan:
I pushed some change to handle the paths at a higher level, avoiding hard coded paths as much as possible.
Also, QtWebEngine copy at the end of the script had to be fixed.

jnbrunet:
> I pushed some change to handle the paths at a higher level, avoiding hard coded paths as much as possible.

This is way better than what I had, thanks a lot !

> Also, QtWebEngine copy at the end of the script should be fixed.

Not sure I'm super happy about this ;-)

PR #1761, opened 2021-01-29
[SofaDefaultType][SofaCUDA] FIX compilation errors

fredroy:
- IncompleteTypeInfo was defined with SOFA_DEFAULTTYPE_API  but this is incorrect as IncompleteTypeInfo  is inline only. It was apparently harmless in Release/RelWithDebug but was complaining in Debug config with an undefined reference in SofaBaseMechanics 🧙‍♂️
- Since a few version of MSVC2019, SofaCUDA could not compile and was complaining on a missing pow() function. Apparently, one has to use powf() instead.

### DIFF
--- a/SofaKernel/modules/SofaDefaultType/src/sofa/defaulttype/typeinfo/models/IncompleteTypeInfo.h
+++ b/SofaKernel/modules/SofaDefaultType/src/sofa/defaulttype/typeinfo/models/IncompleteTypeInfo.h
@@ -27,7 +27,7 @@
 namespace sofa::defaulttype
 {
 template<class TDataType>
-struct SOFA_DEFAULTTYPE_API IncompleteTypeInfo
+struct IncompleteTypeInfo
 {
     /// Template parameter.
     typedef TDataType DataType;
--- a/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaStandardTetrahedralFEMForceField.cu
+++ b/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaStandardTetrahedralFEMForceField.cu
@@ -748,8 +748,8 @@ __device__ void StandardTetrahedralFEMForceFieldCuda_BoyceAndArruda_deriveSPKTen
         }
     }
 
-    //SPKTensorGeneral=((inversematrix*(Real)(-1.0/3.0)*I1+ID)*(Real)(1.0/2.0)*pow(sinfo->J,(Real)(-2.0/3.0))+(inversematrix*(Real)(-2.0/3.0)*I1*I1+ID*(Real)2.0*I1)*(Real)(1.0/160.0)*pow(sinfo->J,(Real)(-4.0/3.0))+(ID*(Real)3.0*I1*I1-inversematrix*I1*I1*I1)*(Real)(11.0/(1050.0*8*8))*pow(sinfo->J,(Real)(-2.0))
-    ////	+(inversematrix*(Real)(-4.0/3.0)*pow(I1,(Real)4.0)+ID*(Real)4.0*pow(I1,(Real)3.0))*(Real)(19.0/(7000.0*8.0*8.0*8.0))*pow(sinfo->J,(Real)(-8.0/3.0))+(inversematrix*(Real)(-5.0/3.0)*pow(I1,(Real)5.0)+ID*(Real)5.0*pow(I1,(Real)4.0))*(Real)(519.0/(673750.0*8.0*8.0*8.0*8.0))*pow(sinfo->J,(Real)(-10.0/3.0)))*2.0*mu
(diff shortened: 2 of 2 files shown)


damienmarchal:
Looks ok to me. Any one for a quick feedback on reviewing this fast-merge PR ?

SofaKernel/modules/SofaDefaultType/src/sofa/defaulttype/typeinfo/models/IncompleteTypeInfo.h
```diff
@@ -27,7 +27,7 @@
 namespace sofa::defaulttype
 {
 template<class TDataType>
-struct SOFA_DEFAULTTYPE_API IncompleteTypeInfo
+struct IncompleteTypeInfo
```
damienmarchal, line 30: I'm not sure to understand why having the declspec is problematic in this case.

fredroy: Because the struct IncompleteTypeInfo is purely inline, so it does not need to export any symbols in the generated .lib

damienmarchal: I'm not sure that being inlineable means that the structure will be inlined in any cases.

fredroy: by pure inline, I wanted to say not present in a translation unit (aka a cpp).
In this case, you know the content of IncompleteTypeInfo (i.e the definitions) by including the header, not by linking a .lib
In MSVC you **must** say "I want to export the symbol(s)" when you compile a cpp. Here that not the case as there is not any symbol defined in a translation unit (cpp)

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaStandardTetrahedralFEMForceField.cu
```diff
@@ -762,7 +762,7 @@ __device__ void StandardTetrahedralFEMForceFieldCuda_BoyceAndArruda_deriveSPKTen
         {
             result[j][k] = inversematrix[j][k] * (-1.0/3.0) * I1;
             result[j][k] = result[j][k] + ID[j][k];
-            result[j][k] = result[j][k] * (1.0/2.0) * pow(J, (real)(-2.0/3.0));
+            result[j][k] = result[j][k] * (1.0/2.0) * powf(J, (real)(-2.0/3.0));
```
epernod: pow in float even if real is double?

SofaKernel/modules/SofaDefaultType/src/sofa/defaulttype/typeinfo/models/IncompleteTypeInfo.h
epernod: 👍

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaStandardTetrahedralFEMForceField.cu
fredroy: you are right that it seems that you need to use pow if double and powf if float.😢
Never tried as I never compile with double with CUDA. 🙄

fredroy: it was compiling fine with CUDA_DOUBLE enabled but I guess it was casting the double in float and then back to float.
I pushed a templated version which *should* use the double version if needed

epernod (approved):

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaStandardTetrahedralFEMForceField.cu
epernod: In fact I don't understand why std::pow was not working, it is supposed to handle double or float
https://en.cppreference.com/w/cpp/numeric/math/pow

fredroy: err... this is a CUDA file so you should look for:
 - https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__SINGLE.html 
 - https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__DOUBLE.html

epernod: 👍

PR #1770, opened 2021-02-02
[SofaHelper] Remove SofaSimulationCore dependency from AdvancedTimer

fredroy:
AdvancedTimer has a function end(), which wants a Node just to get a time and a dt...
And add an ugly (implicit) dependency on SofaSimulationCore (header-wise)

Consequently, this PR :
 - adds a fonction end() which takes a time and a dt instead of a Node*
 - deprecates the original function end(... , Node* node) but change its behavior (calling the standard end())

Maybe it would even better to just delete the end() taking a Node*, as only BatchGUI, runSofa and its own test are using it. (in the SOFA code base at least)

And as a side commit, clean the includes (json double-quoted include 🥱 )


fredroy:
... we did not say that we should be throw an error into the end(.... Node*) function ? 😏
I will do an other PR to add it

jnbrunet:
Ha crap, I completely forgot !

PR #1775, opened 2021-02-04
[SofaKernel] Port Mat from Defaulttype to Sofa.Type

fredroy:
## Description:
Move Mat (and MatSym) and their satellites file (algo...) to the new Sofa.Type module.
Like #1734  There is a compatibility layer for transition, so it should be as least breaking as possible.

- sofa::defaulttype::Mat becomes sofa::type::Mat
- sofa::defaulttype::MatSym becomes sofa::type::MatSym

## Breaking
Same as #1734:
 - if inclusion of _defaulttype_ and _type_ namespace, the compiler cannot chose which one to chose for the free functions like invert, etc. when the code did not called them from their respective namespace.It was not a problem because the compiler could find them previously thanks to [ADL](https://en.wikipedia.org/wiki/Argument-dependent_name_lookup).
Fixing is easy by just adding the respective namespace.

## Talking

- An other contribution I would like to spread to the sofa code base: 
Instead of throwing a text information  if the function cannot do (like matrix non invertible, etc), 
this should always return a boolean value and apply the [[nodiscard]] flag to oblige the user to use the return value.
This makes the compiler generates a warning; I would even push to make it a compilation error
More generally, it would be cool to use the  [[nodiscard]] flag much more for critical code.
~~I tried to even use the [[nodiscard "MESSAGE"]] but this feature seems only included with c++20 so I may change this PR to run successfully on the CI.~~
As expected, this is not supported by the CI 🙁

- the compatibility layer still holds the dependency on helper for the messaging, so we should discuss about that.


fredroy:

fredroy:
(forgot to mention:) #1521

PR #1779, opened 2021-02-04
[SofaMiscFEM] Minor changes in TrianglePressureFF and TriangularFEMFF

epernod:
```TriangularPressureForceField```:
- Set componentState to Invalid if topology is not found
- Fix force vectors display

```TriangularFEMForceField```:
- Remove GetFracturableEdge deprecated method which could not work.
- Various fix in the display to avoid computation inside draw method
- Replace default display by stress value

image

### DIFF
--- a/modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
+++ b/modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
@@ -92,7 +92,6 @@ TriangularFEMForceField<DataTypes>::TriangularFEMForceField()
     , f_damping(initData(&f_damping,(Real)0.,"damping","Ratio damping/stiffness"))
     , m_rotatedInitialElements(initData(&m_rotatedInitialElements,"rotatedInitialElements","Flag activating rendering of stress directions within each triangle"))
     , m_initialTransformation(initData(&m_initialTransformation,"initialTransformation","Flag activating rendering of stress directions within each triangle"))
-    , f_fracturable(initData(&f_fracturable,false,"fracturable","the forcefield computes the next fracturable Edge"))
     , hosfordExponant(initData(&hosfordExponant, (Real)1.0, "hosfordExponant","Exponant in the Hosford yield criteria"))
     , criteriaValue(initData(&criteriaValue, (Real)1e15, "criteriaValue","Fracturable threshold used to draw fracturable triangles"))
     , showStressValue(initData(&showStressValue,false,"showStressValue","Flag activating rendering of stress values as a color in each triangle"))
@@ -505,35 +504,6 @@ void TriangularFEMForceField<DataTypes>::getFractureCriteria(int elementIndex, D
         value = 0;
     }
 }
-
-template<class DataTypes>
-typename TriangularFEMForceField<DataTypes>::Index TriangularFEMForceField<DataTypes>::getFracturedEdge()
-{
-    helper::vector<EdgeInformation>& edgeInf = *(edgeInfo.beginEdit());
-
-    if (f_fracturable.getValue())
-    {
-        std::size_t nbEdges = m_topology->getNbEdges();
-
-        for(Size i=0; i<nbEdges; i++ )
-        {
-            if (edgeInf[i].fracturable)
-            {
-                return i;
-            }
-        }
-    }
-
-    edgeInfo.endEdit();
-
-    return sofa::InvalidID;
-}
-
-
-
-
-
-
 
 
 // --------------------------------------------------------------------------------------
@@ -1422,7 +1392,7 @@ void TriangularFEMForceField<DataTypes>::draw(const core::visual::VisualParams* 
     const VecCoord& x = this->mstate->read(core::ConstVecCoordId::position())->getValue();
     unsigned int nbTriangles=m_topology->getNbTriangles();
 
-    if (!f_fracturable.getValue() && !this->showFracturableTriangles.getValue())
+    if (!this->showFracturableTriangles.getValue())
     {
         for(unsigned int i=0; i<nbTriangles; ++i)
         {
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by fredroy** [comment]

File: modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl [hunk 1]
```diff
@@ -1415,56 +1390,29 @@ void TriangularFEMForceField<DataTypes>::draw(const core::visual::VisualParams*
...
+    const helper::vector<TriangleInformation>& triangleInf = triangleInfo.getValue();
+    Size nbTriangles = m_topology->getNbTriangles();
+      
     if (showStressVector.getValue())
     {
-        color = sofa::helper::types::RGBAColor(1,0,1,1);
-        for(unsigned int i=0; i<nbTriangles; ++i)
+        const VecCoord& x = this->mstate->read(core::ConstVecCoordId::position())->getValue();
+        std::vector<sofa::defaulttype::Vector3> vertices;
+        for(unsigned int i=0; i< nbTriangles; ++i)
```
hugtalbot: use Size instead of unsigned int

modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
```diff
@@ -92,10 +92,9 @@ TriangularFEMForceField<DataTypes>::TriangularFEMForceField()
     , f_damping(initData(&f_damping,(Real)0.,"damping","Ratio damping/stiffness"))
     , m_rotatedInitialElements(initData(&m_rotatedInitialElements,"rotatedInitialElements","Flag activating rendering of stress directions within each triangle"))
     , m_initialTransformation(initData(&m_initialTransformation,"initialTransformation","Flag activating rendering of stress directions within each triangle"))
-    , f_fracturable(initData(&f_fracturable,false,"fracturable","the forcefield computes the next fracturable Edge"))
     , hosfordExponant(initData(&hosfordExponant, (Real)1.0, "hosfordExponant","Exponant in the Hosford yield criteria"))
     , criteriaValue(initData(&criteriaValue, (Real)1e15, "criteriaValue","Fracturable threshold used to draw fracturable triangles"))
-    , showStressValue(initData(&showStressValue,false,"showStressValue","Flag activating rendering of stress values as a color in each triangle"))
+    , showStressValue(initData(&showStressValue,true,"showStressValue","Flag activating rendering of stress values as a color in each triangle"))
```
hugtalbot, line 97: why changing the default value ? (just a curious question)

modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
```diff
@@ -1549,8 +1505,7 @@ void TriangularFEMForceField<DataTypes>::draw(const core::visual::VisualParams*
             }
             vparams->drawTool()->drawTriangles(vertices,color);
```
hugtalbot: should it not be colorVector here?

modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
```diff
@@ -1512,10 +1463,15 @@ void TriangularFEMForceField<DataTypes>::draw(const core::visual::VisualParams*
         vparams->drawTool()->drawTriangles(vertices,colorVector);
         vertices.clear();
         colorVector.clear();
+        vertexInfo.endEdit();
     }
 
     if (showFracturableTriangles.getValue())
```
hugtalbot, line 1469: If I understand we can still display which triangles meet the fracturable criterion, is it correct?

epernod:
I would not backport this one. No relevant reason.

modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
epernod: see response below

modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
epernod: see response of another question below

modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
epernod: yes, even if I couldn't see one during the tests. I only remove the default FEM rendering which was showing green/blue triangles without meaning. 
Now by default if you render this FEM it will show the stressMap value. But you can still uncheck this rendering and thus it will render nothing.

modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
epernod: no because this one is inside the for loop which compute the color depending on the criteria value.

modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
```diff
@@ -1512,13 +1463,18 @@ void TriangularFEMForceField<DataTypes>::draw(const core::visual::VisualParams*
...
     {
+        std::vector<sofa::helper::types::RGBAColor> colorVector;
```
hugtalbot, line 1471: then this one is unused, isn't it?

hugtalbot: and same for lines 1499, 1501 and 1503

epernod: no in fact you were right the colorVector should be used. In fact the drawing should be outside the for loop

hugtalbot (approved):

PR #1780, opened 2021-02-04
[SofaCUDA] Fix CudaFixedConstraint when indices are not contiguous

epernod:

### DIFF
--- a/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaFixedConstraint.inl
+++ b/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaFixedConstraint.inl
@@ -73,8 +73,8 @@ template<class TCoord, class TDeriv, class TReal>
 void FixedConstraintInternalData< gpu::cuda::CudaVectorTypes<TCoord,TDeriv,TReal> >::init(Main* m)
 {
     Data& data = *m->data;
-    data.minIndex = -1;
-    data.maxIndex = -1;
+    data.minIndex = sofa::InvalidID;
+    data.maxIndex = sofa::InvalidID;
     data.cudaIndices.clear();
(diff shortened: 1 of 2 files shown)


applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaFixedConstraint.inl
```diff
@@ -342,12 +346,12 @@ void FixedConstraintInternalData<gpu::cuda::CudaVec1fTypes>::projectResponse(Mai
...
 void FixedConstraintInternalData<gpu::cuda::CudaVec3fTypes>::projectResponse(Main* m, VecDeriv& dx)
-{
+{    
     Data& data = *m->data;
     if (m->d_fixAll.getValue())
-        FixedConstraintCuda3f_projectResponseContiguous(dx.size(), ((float*)dx.deviceWrite()));
-    else if (data.minIndex >= 0)
-        FixedConstraintCuda3f_projectResponseContiguous(data.maxIndex-data.minIndex+1, ((float*)dx.deviceWrite())+3*data.minIndex);
+        FixedConstraintCuda3f_projectResponseContiguous(dx.size(), ((float*)dx.deviceWrite()));        
+    else if (data.minIndex != sofa::InvalidID)
```
epernod, line 353: @hugtalbot here is the bug I fixed but actually the code was certainly failing at other places because before data.minIndex was an int. So the test >= 0 means a valid int value. 
But as data.minIndex was changed to a Sofa:Index (aka unsigned int) in a previous PR and still initialised to -1, the invalid value was 42x10^9 something (max unsigned int). So the test >=0 was always true. 
So to fix I changed all the if condition to test against an InvalidID value.

guparan (approved):
LGTM

PR #1815, opened 2021-02-11
[SofaBaseMechanics] fixes reinit of BarycentricMapping

EulalieCoevoet:
This PR fixes reinit() of BarycentricMapping when using `useRestPosition` option.


SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/BarycentricMapping.inl
```diff
@@ -262,10 +262,7 @@ void BarycentricMapping<TIn, TOut>::init()
     if (!this->toModel)
```
hugtalbot, line 262: This check of valid pointer could be integrated within the ```initMapper()``` function to avoid crash, don't  you think?

EulalieCoevoet: What about using `d_componentState.getValue()==Valid` instead?

SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/BarycentricMapping.inl
```diff
@@ -259,27 +259,35 @@ void BarycentricMapping<TIn, TOut>::init()
...
-    {
-        d_mapper->clear();
-        d_mapper->init (((const core::State<Out> *)this->toModel)->read(core::ConstVecCoordId::position())->getValue(), ((const core::State<In> *)this->fromModel)->read(core::ConstVecCoordId::position())->getValue() );
-    }
+    if (d_componentState.getValue() != ComponentState::Valid)
+        return;
+
+    d_mapper->clear();
+    initMapper();
 }
```
hugtalbot, line 278: Hey @EulalieCoevoet 
In your design the BMapping is never set back to Valid (if Invalid), should it not be the role of the reinit() function?

EulalieCoevoet: Well maybe, but that's not what initMapper() does, it just computes the barycentric map, and we shouldn't try if the component is not Valid.

PR #1824, opened 2021-02-16
[SofaHelper] Add two search paths for every prefixes of the plugin manager

jnbrunet:
When the plugin manager searches for a plugin, it will loop through the prefixes defined inside the PluginRepository, and append it the plugin name. If this file exists, it will be directly returned. Else, it will go recursively through each prefixes trying to find the file, which is inefficient and does not follow symbolic links. Especially since the prefixes contain large directories such as "/usr/lib".

This PR adds two hard coded path to each prefixes which will help the plugin manager find common plugins usually located in _"$SOFA_ROOT/plugins/PLUGIN_NAME/lib/libPLUGIN_NAME.so"_ **just before** doing the recursive search.

This should improve the speed for locating common plugins, and moreover allow some symbolic links to external plugins since the recursive search do not follow these links.

### DIFF
--- a/SofaKernel/modules/SofaHelper/src/sofa/helper/system/PluginManager.cpp
+++ b/SofaKernel/modules/SofaHelper/src/sofa/helper/system/PluginManager.cpp
@@ -29,6 +29,7 @@ using sofa::helper::system::FileSystem;
 
 #include <boost/filesystem.hpp>
 #include <fstream>
+#include <array>
 
 using sofa::helper::Utils;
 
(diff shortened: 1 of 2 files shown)


SofaKernel/modules/SofaHelper/src/sofa/helper/system/PluginManager.cpp
```diff
@@ -321,12 +321,19 @@ std::string PluginManager::findPlugin(const std::string& pluginName, const std::
...
-    for (std::vector<std::string>::iterator i = searchPaths.begin(); i!=searchPaths.end(); i++)
-    {
-        const std::string path = *i + "/" + libName;
-        if (FileSystem::isFile(path))
-            return path;
+    for (const auto & prefix : searchPaths) {
+        const std::array<std::string, 3> paths = {
+                prefix + "/" + libName,
+                prefix + "/" + pluginName + "/" + libName,
+                prefix + "/" + pluginName + "/lib/" + libName
```
guparan, line 330: And what about the other 69.5% of our users who are on Windows? :smile:

guparan:
Thanks for the PR @jnbrunet :+1:
Maybe you could also lower the default value of `maxRecursiveDepth` from 6 to 3?

SofaKernel/modules/SofaHelper/src/sofa/helper/system/PluginManager.cpp
jnbrunet: It should be fine, no? They will get caught into the first `prefix + "/" + libName` (which is exactly the same path as before). The other two paths should be quickly discarded as non-existent.

jnbrunet:
> Thanks for the PR @jnbrunet +1
> Maybe you could also lower the default value of `maxRecursiveDepth` from 6 to 3?

Sure thing. Note that this will make the PR "breaking" (more of a behaviour change).

SofaKernel/modules/SofaHelper/src/sofa/helper/system/PluginManager.cpp
guparan: Nope, plugins are located in `$SOFA_ROOT/plugins/PLUGIN_NAME/bin/PLUGIN_NAME.dll` so, while non-breaking, this change will not help Windows users.

jnbrunet: Haaa ok, I though there were no such things as relocatable plugins in Windows. I'll add this then.

jnbrunet:

PR #1850, opened 2021-02-22
[SofaGuiQt] Clean QtGLViewer with key events

hugtalbot:
Minor fix of a line causing a double call to the keyPressEvent function

### DIFF
--- a/modules/SofaGuiQt/src/sofa/gui/qt/viewer/qgl/QtGLViewer.cpp
+++ b/modules/SofaGuiQt/src/sofa/gui/qt/viewer/qgl/QtGLViewer.cpp
@@ -868,7 +868,6 @@ void QtGLViewer::keyPressEvent ( QKeyEvent * e )
         default:
         {
             SofaViewer::keyPressEvent(e);
-            QGLViewer::keyPressEvent(e);
         }
         }
     }


modules/SofaGuiQt/src/sofa/gui/qt/viewer/qgl/QtGLViewer.cpp
```diff
@@ -868,7 +868,6 @@ void QtGLViewer::keyPressEvent ( QKeyEvent * e )
         default:
         {
             SofaViewer::keyPressEvent(e);
-            QGLViewer::keyPressEvent(e);
```
epernod, line 871: ok but are you sure both viewer were doing the same thing?

epernod: Checking again, SofaViewer doesn't inherite from QGLViewer... so you might loose some QGLViewer key commands. 
Are they important ? I let you decide.

hugtalbot: we checked with @fredroy it seemed fine to us: either you call the one from SofaViewer OR QGLViewer OR you implement your inherit in QtGLViewer but not two in a row. This is confusing and prone to error.
what was weird is that the event keyPressEvent in SofaViewer was thus called twice !

epernod: ok!

PR #1862, opened 2021-02-24
[SofaConstraint] GenericConstraintSolver: compute compliances concurrently

EulalieCoevoet:
The computation of the compliance matrices is expensive. When we have multiple objects in our scene it can be interesting to compute these matrices concurrently.  
I'm not an expert in multithreading... so any feedback would be much appreciated :)  
Here's the speed up I get (percentage wrt the whole simulation):

With two identical FEM models:
image
image with multithreading

With four identical FEM models:
image
image with multithreading


modules/SofaConstraint/src/SofaConstraint/GenericConstraintSolver.h
```diff
@@ -104,6 +104,7 @@ class SOFA_SOFACONSTRAINT_API GenericConstraintSolver : public ConstraintSolverI
     Data<bool> allVerified; ///< All contraints must be verified (each constraint's error < tolerance)
     Data<bool> schemeCorrection; ///< Apply new scheme where compliance is progressively corrected
     Data<bool> unbuilt; ///< Compliance is not fully built
+    Data<bool> multithreading; ///< Compliances built concurrently
```
hugtalbot: data naming should now be d_ 
maybe we should keep in mind to clean the GenericCS !

modules/SofaConstraint/src/SofaConstraint/GenericConstraintSolver.cpp
```diff
@@ -292,13 +296,48 @@ bool GenericConstraintSolver::buildSystem(const core::ConstraintParams *cParams,
...
+
+            for (unsigned int i=0; i<constraintCorrections.size(); i++)
+            {
+                Ws[i].resize(current_cp->W.rowSize(),current_cp->W.colSize());
+                Ps[i] = *cParams;
+                auto Wi = &Ws[i];
+                auto Pi = Ps[i];
+                core::behavior::BaseConstraintCorrection* cc = constraintCorrections[i];
+                threads[i] = std::thread([Pi, Wi, cc] {
+                    if (cc->isActive()) {
```
jnbrunet: Maybe just get this condition out of the lambda function. It will avoid starting a thread for no reason when the cc isn't activated.

modules/SofaConstraint/src/SofaConstraint/GenericConstraintSolver.cpp
```diff
@@ -292,13 +296,48 @@ bool GenericConstraintSolver::buildSystem(const core::ConstraintParams *cParams,
...
+                        cc->addComplianceInConstraintSpace(&Pi, Wi);
+                    }
+                });
+            }
+            for (auto& t : threads)
+                t.join();
+
+            for (unsigned int i=0; i<constraintCorrections.size(); i++)
```
jnbrunet: I'm not sure if it will change something here because of the virtual calls to `current_cp->W.add()` and `Ws[i].element()`, but in time sensitive loops, I usually tend to strip out functions calls from the end condition of the loop. This will allow the compiler to be certain that the end condition will not change during the loop. i.e.:

```cpp
auto & W = current_cp->W;
const auto rows = W.rowSize();
const auto cols = W.colSize();
const auto n_constraintCorrections = constraintCorrections.size();

// Accumulate the contribution of each constraints
// into the system's compliant matrix W
for (sofa::Index i = 0; i < n_constraintCorrections; i++) {
    const auto * cc = constraintCorrections[i];
    const auto & Wi = Ws[i];
    
    if (!cc->isActive()) {
        continue;
    }
    
    for (sofa::Index j = 0; j < rows; ++j) {
        for (sofa::Index l = 0; l < cols; ++l) {
            W.add(j, l, Wi.element(j,l));
        }
    }
}
```

I also refactored a little bit the code to the format I usually like to read, feel free to take/reject whatever suits you ;)

modules/SofaConstraint/src/SofaConstraint/GenericConstraintSolver.cpp
```diff
@@ -31,6 +31,9 @@
 #include <sofa/core/behavior/MultiVec.h>
 #include <sofa/simulation/Node.h>
 
+#include <thread>
+#include <functional>
```
epernod, line 36: Hi, thank you very much for this kind of contribution @EulalieCoevoet !
It is funny because yesterday I was thinking of reworking the Gauss Seidel code to parallelize it. 

Back to your PR I think the clean approach would be to enable the multi-thread plugin from SOFA which give access to a MT api where you "just" need to create a runner and give him some work.
I will be happy to help you on that and test your changes. Could you describe which scene you used to test and how did you log it.

EulalieCoevoet: Hi, thanks @epernod sounds good, I wasn't aware of this plugin, I'll have a look. 
Well in my scene I have 38 constraints per FEM model (700nodes). For the lod I just used `Log Time` of the `Stats` tab of runSofa..

modules/SofaConstraint/src/SofaConstraint/GenericConstraintSolver.h
EulalieCoevoet: sure, sorry I have the tendency to blend my code in the existing when I don't feel like changing everything.. bad habit (de toqué).

fspadoni (commented):
Hello, I've just had a quick look. 

There is the TaskScheduler class in SOFA to help to  dynamically parallelize this kind of loop without directly manipulating the threads.

The TaskSchduler is an abtract class and the DefaultTaskScheduler is an implementation.
They are in SofaKernel/modules/SofaSimulationCore/src/sofa/simulation so no need to link to MultiThreading plugin.
The MultiThreading plugin contains a few component that use the TaskScheduler to parallelize some computation so it's helpful to see how to use the task scheduler to define , create and synchronize tasks.

modules/SofaConstraint/src/SofaConstraint/GenericConstraintSolver.cpp
```diff
@@ -292,13 +296,60 @@ bool GenericConstraintSolver::buildSystem(const core::ConstraintParams *cParams,
...
 
-        for (unsigned int i=0; i<constraintCorrections.size(); i++)
-        {
-            core::behavior::BaseConstraintCorrection* cc = constraintCorrections[i];
-            if (!cc->isActive()) continue;
-            sofa::helper::AdvancedTimer::stepBegin("Object name: " + cc->getName());
-            cc->addComplianceInConstraintSpace(cParams, &current_cp->W);
-            sofa::helper::AdvancedTimer::stepEnd("Object name: " + cc->getName());
+        if(d_multithreading.getValue()){
+            helper::vector<std::thread> threads(constraintCorrections.size());
```
hugtalbot: Risk here to have N threads incoming without limits

hugtalbot: Multithreading plugin by @fspadoni prevents this with the scheduler

alxbilger:
I recommend adding tests, and repeat them several times due to the non-deterministic nature of a multi-threaded approach.

modules/SofaConstraint/src/SofaConstraint/GenericConstraintSolver.cpp
EulalieCoevoet: @fspadoni @hugtalbot I've made the changes.

fspadoni (commented):
Hello, the task scheduler create function should not be explicity call.  the task scheduler just need to be initialized once so I would do it in the init() function of the component and not during the simulation. 

It's ok to  call theThe TaskScheduler init() function with no arguments, in this case it creates one worker thread for each cpu core.

TaskSchduler* taskscheduler =  TaskSchduler::getInstance();
taskscheduler ->init();

The TaskScheduler stop function should be called in the destructor or  clean up function of the component.

The doc in the MultiThreading plugin is not up to date but it explains what the task scheduler does.

fspadoni (approved):
It looks OK.

fredroy:
I was testing a benchmark scene with quite a lot of constraints (frictions) and:
(exactly the same scene)
```
master  : 5000 iterations done in 203.986 s ( 24.5115 FPS)
this PR : 5000 iterations done in 134.177 s ( 37.2643 FPS)
```
with just having to set `multithread="true"` (and building multithreading plugin) this is **cool** 🥳

Edit: and it was with Windows/MSVC

EulalieCoevoet:
You don't need the MultiThreading plugin. Since the TaskScheduler is in SofaSimulationCore, I think it's nice to always have the feature, if that's okay. 

However... something is off. I guess I did something wrong with the TaskScheduler. Now when I use the feature, 40% of my CPU is used by runSofa without necessarily running the simulation... only one step is enough to make everything go crazy.

fredroy:
[ci-build][with-all-tests]

EulalieCoevoet:
Yes, sorry.

EulalieCoevoet:
No, sorry :/
Well, for me it happens with all simulations. For instance: 
`examples/Components/constraint/BilateralInteractionConstraint.scn`

fspadoni:
Hopefully the PR #1997 should fix the busy CPU problem when simulation is not running.

fredroy:
Testing this PR coupled with #1997  on Windows/MSVC and there is no more CPU "business" if the simulation is stopped.
So I guess this PR is ready IMO.

fredroy:
Ready once #1997 is merged

PR #1927, opened 2021-03-16
[SofaSimulationCore] FIX Task scheduler memory leak

alxbilger:
Fix memory leak in task scheduler:

- Add comments
- Add unit tests
- Use shared pointer instead of raw pointer: 

The static raw pointer had two bad effects:
- Memory leak
- Task scheduler not being stopped

Note: The current API for the task scheduler is very confusing. I suggest to clarify the expected behavior.


fredroy:
[ci-build][with-all-tests]

fspadoni (commented):
Good to use a smart pointer and fix the memory leak.

What about using a unique_ptr ? 
The task scheduler should manage itself and no other class should control its lifetime storing a refernce to it.

alxbilger:
@fspadoni you're right, I'll change it

alxbilger:
@fspadoni Is the task scheduler designed to have local schedulers? For example, can I have a task scheduler specific to a component? Is it restricted to have only one instance per derived type?

fspadoni:
> @fspadoni Is the task scheduler designed to have local schedulers? For example, can I have a task scheduler specific to a component? Is it restricted to have only one instance per derived type?

There is only one global instance of the task scheduler class and this is shared between all components. 
Any components can create and submit tasks to task scheduler. 
The tasks are queued and process by the worker threads.
The task scheduler manages the worker threads (only one per cpu core by default)

alxbilger:
@hugtalbot @guparan I am not sure I just need to switch from EXPECT_EQ to EXPECT_NE. When I'll refactor the interface of the TaskScheduler, a large majority of the test will not make sense. I suggest to merge this PR now, and I'll change the tests in my next PR.

alxbilger:
[ci-build][with-all-tests]

alxbilger:
I removed completely the unit tests. I'll add them later in another pull request.

hugtalbot (approved):

PR #1937, opened 2021-03-18
[Sofa.Type] Add forgotten function declaration, used in Debug

fredroy:
Debug compilation was failing due to the fact that the header is not supposed to know `vector_access_failure` , defined in the cpp file.

### DIFF
--- a/SofaKernel/modules/Sofa.Type/src/sofa/type/stdtype/vector_T.h
+++ b/SofaKernel/modules/Sofa.Type/src/sofa/type/stdtype/vector_T.h
@@ -40,6 +40,8 @@ namespace sofa::type::stdtype
 
 static constexpr bool isEnabledVectorAccessChecking {SOFA_VECTOR_CHECK_ACCESS};
 
+extern void SOFA_TYPE_API vector_access_failure(const void* vec, unsigned size, unsigned i, const std::type_info& type);
+
 // standard vector dont use the CPUMemoryManager given as template
 template <typename T>
 class CPUMemoryManager;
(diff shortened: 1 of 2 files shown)


SofaKernel/modules/SofaBaseLinearSolver/src/SofaBaseLinearSolver/FullVector.inl
```diff
@@ -36,7 +36,7 @@ void FullVector<Real>::checkIndex(Index n) const
     {
         msg_error("FullVector") << "in vector<" << sofa::helper::gettypename(typeid(*this)) << "> " << std::hex << (long)this << std::dec << " size " << cursize << " : invalid index " << (int)n;
         sofa::helper::BackTrace::dump();
-        assert(i < size);
+        assert(n < cursize);
```
epernod, line 39: how can this assert happen if inside this if: https://github.com/sofa-framework/sofa/blob/47cde44a64dbc81c54e7f10d0deb17edcebaa1f2/SofaKernel/modules/SofaBaseLinearSolver/src/SofaBaseLinearSolver/FullVector.inl#L35

SofaKernel/modules/Sofa.Type/src/sofa/type/stdtype/vector_T.h
```diff
@@ -40,6 +40,8 @@ namespace sofa::type::stdtype
 
 static constexpr bool isEnabledVectorAccessChecking {SOFA_VECTOR_CHECK_ACCESS};
 
+extern void SOFA_TYPE_API vector_access_failure(const void* vec, unsigned size, unsigned i, const std::type_info& type);
```
epernod, line 43: hard to naviguate between files, is there a def somewhere?

SofaKernel/modules/SofaBaseLinearSolver/src/SofaBaseLinearSolver/FullVector.inl
fredroy: http://www.cplusplus.com/reference/cassert/assert/

```
If the argument expression of this macro with functional form compares equal to zero (i.e., the expression is false), a message is written to the standard error device and abort is called, terminating the program execution.
```
it is made that it will always assert;
and that was the code present before actually 😗

epernod: ah yes... sorry I always read assert in the wrong sens

PR #1939, opened 2021-03-19
[SofaBaseMechanics] Restore useRestPos Data initialization for BarycentricMapping

fredroy:
The initialization of the data disappeared at some point in the past 🤥
This PR restores it.

Credits to:
 - [InSimo](https://github.com/InSimo)(@JeremieA , @fjourdes ) 
 - @remibessard who reported it back to us

### DIFF
--- a/SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/BarycentricMapping.h
+++ b/SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/BarycentricMapping.h
@@ -63,7 +63,7 @@ public:
     typedef typename Inherit1::ForceMask ForceMask;
 
 public:
-    Data< bool > useRestPosition; ///< Use the rest position of the input and output models to initialize the mapping    
+    Data< bool > d_useRestPosition; ///< Use the rest position of the input and output models to initialize the mapping    
 
     SingleLink<BarycentricMapping<In,Out>,Mapper,BaseLink::FLAG_STRONGLINK> d_mapper;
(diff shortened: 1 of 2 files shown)


alxbilger:
Can we take the opportunity to renaming it `d_useRestPosition` ?

guparan (approved):

guparan:
> 
> 
> Can we take the opportunity to renaming it `d_useRestPosition` ?

I'm not sure it's a good idea to transform a fix into a breaking change. Waiting for @fredroy insight :mag:

fredroy:
> 
> 
> > Can we take the opportunity to renaming it `d_useRestPosition` ?
> 
> I'm not sure it's a good idea to transform a fix into a breaking change. Waiting for @fredroy insight 🔍

Exactly my thoughts.
Wondering the fact that indeed this is badly needed BUT it will be breaking.

Arguments for renaming:
 - does not follow the guideline and the other Data<> in BarycentricMapping are following it,
 - referenced only **ONCE** in the whole sofa code base
 - does not breaking anything scene-wise (the stringName wont change)

Arguments for not renaming:
- breaking plugins using this Data<> directly

Conclusion: 
My opinion would be to rename it indeed 🧐

epernod:
as all the others Data of this components already have the d_. I'm for renaming it as well.

fredroy:
Actually this is not really true, the problem is that the other things with a 'd_' are NOT data 🤪, they are Link and thus should be prefixed with 'l_' . 
So this class is not following the correct guidelines 😬 (and members not having 'm_' etc...)

epernod:
ok... I think we should do an afternoon of brainless coding sprint to update all those components with the good guidelines... 
so this one is ready to go I guess. 
LGTM !

PR #1913, opened 2021-03-11
[All] Rename SOFA_NO_OPENGL + Relocate Sofa.GL + Improve sofa_add_* macros

guparan:
1. The CMake option to enable/disable OpenGL is now `SOFA_WITH_OPENGL`.
2. SOFA_NO_OPENGL is completely removed from SOFA sources (only the CMake variable persists).
3. Sofa.GL is added by SofaKernel/modules/CMakeLists.txt (not by SofaFramework/CMakeLists.txt anymore).
4. sofa_add_* macros are now capable of creating dependent options. 
Example:   
```cmake
sofa_add_module(Sofa.GL Sofa.GL ON # the 3 usual sofa_add_* arguments
    WHEN_TO_SHOW "SOFA_WITH_OPENGL" # [new arg] if SOFA_WITH_OPENGL is true, then the option is showed in CMake-GUI
    VALUE_IF_HIDDEN OFF # [new arg] if SOFA_WITH_OPENGL is false, then the option is hidden and set to OFF
    )
```
5. SofaComponentAll does not fail anymore at configure time if some modules are not found. It will cover as many modules as possible. **This will fix our nightly minimal builds.**
______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #1941, opened 2021-03-19
[Sofa.GL] Relocate the module (again) + clarify module vs plugin definitions

guparan:
The issue with #1913 was that Sofa.GL was added to plugin_list.conf.default 
This PR aims to clarify things between modules and plugins.
`sofa_add_plugin` will now be the ONLY macro that adds targets to plugin_list.conf.default, which totally makes sense.

Also, for the record:
> Plugin = Something that CAN be loaded by the PluginManager, through a RequiredPlugin for example. The entry point of a Plugin is initExternalModule (and that name could be better).
> 
> Module = Something that CANNOT be loaded by the PluginManager. Most of the time, modules are not loaded manually, they are loaded implicitly during dependency resolution. The entry point of a Module is init (and that name could also be better).
> 
> From a CMake point of view, both Plugin and Module can be standalone packages or simple targets added to a bigger package.


SofaKernel/modules/CMakeLists.txt
```diff
@@ -2,12 +2,16 @@
...
+sofa_add_plugin(SofaRigid SofaRigid ON)
+sofa_add_plugin(SofaDeformable SofaDeformable ON)
+sofa_add_plugin(SofaObjectInteraction SofaObjectInteraction ON)
+sofa_add_plugin(SofaMeshCollision SofaMeshCollision ON)
+sofa_add_plugin(SofaEngine SofaEngine ON)
+sofa_add_plugin(SofaExplicitOdeSolver SofaExplicitOdeSolver ON)
+sofa_add_plugin(SofaImplicitOdeSolver SofaImplicitOdeSolver ON)
+sofa_add_plugin(SofaLoader SofaLoader ON)
+
+sofa_add_module(Sofa.GL Sofa.GL ON
```
fredroy: It was in the previous PR but Sofa.GL should be at the top of this file, as SofaSimpleFem, SofaRigid, etc _could need_ Sofa.GL ; in this current setup they will never find it.

guparan: Ok I update this :+1:

fredroy (approved):

fredroy:

PR #1951, opened 2021-03-26
[Sofa.GL] static variable belongs to the class

alxbilger:
Two variables with internal linkage were defined with the same symbol, and the same value. Instead, I propose to associate these variable to the classes where they are used. I did not made them global variables because we could imagine having two different values for both classes.


alxbilger:
[ci-build][with-all-tests]

PR #1955, opened 2021-03-26
[SofaBaseTopology] Remove definition of real

alxbilger:
"real" is replaced with its only possible type: float. float is assumed
by SparseGridTopology.cpp. Removing "real" avoids issues when some
templated functions have a template parameter named "real".

### DIFF
--- a/SofaKernel/modules/SofaBaseTopology/src/SofaBaseTopology/polygon_cube_intersection/fast_polygon_cube_intersection.cpp
+++ b/SofaKernel/modules/SofaBaseTopology/src/SofaBaseTopology/polygon_cube_intersection/fast_polygon_cube_intersection.cpp
@@ -70,7 +70,7 @@ namespace polygon_cube_intersection
 
 #define TEST_AGAINST_PARALLEL_PLANES(posbit, negbit, value, limit)	\
 	if (mask & (posbit|negbit)) {					\
-        /*register*/ real temp = value;				\
+        /*register*/ float temp = value;				\
 		if ((mask & posbit) && temp > limit)			\
 			outcode |= posbit;				\
 		else if ((mask & negbit) && temp < -limit)		\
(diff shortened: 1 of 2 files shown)


alxbilger:
[ci-build][with-all-tests]

SofaKernel/modules/SofaBaseTopology/src/SofaBaseTopology/polygon_cube_intersection/fast_polygon_cube_intersection.cpp
```diff
@@ -70,7 +70,7 @@ namespace polygon_cube_intersection
 
 #define TEST_AGAINST_PARALLEL_PLANES(posbit, negbit, value, limit)	\
 	if (mask & (posbit|negbit)) {					\
-        /*register*/ real temp = value;				\
+        /*register*/ float temp = value;				\
```
hugtalbot, line 73: shouldn't we use the Real instead?

alxbilger: Here the function is not templated. So real denotes a type, which was previously defined as a float. The components using those functions assume the parameters are of type float (check SparseGridTopology.cpp, line 817). That is why I chose float to replace real.
SReal is not an option neither because it can be switched between float and double. By default it is a double, so it makes incompatible the use of those functions with float.
A third option would be to template those functions. But we would also need to change all the components using them (e.g. SparseGridTopology).

If we look at SparseGridTopology, I'd say those function must take at least double as parameters (float also if templated). But it would require to modify some code calling those functions.

guparan:
Why not, for the sake of clarity. :+1:
Just note that this `real` thing is how the algorithm was written in the first place.
See https://github.com/erich666/GraphicsGems/blob/master/gemsv/ch7-2/pcube.h

guparan (approved):

alxbilger:
> Why not, for the sake of clarity. +1
> Just note that this `real` thing is how the algorithm was written in the first place.
> See https://github.com/erich666/GraphicsGems/blob/master/gemsv/ch7-2/pcube.h

I understand, but this definition leads to compilation error when compiling with unity builds. As I said, it's because the symbol "real" is used as a symbol for a class template somewhere else. 
An alternative would be to undef real after an include of this file.

PR #1962, opened 2021-03-29
[SofaSimulationCore] Add boolean variables to the base class BaseMech…

alxbilger:
### Context

All instances of the derived classes of `BaseMechanicalVisitor` execute the same functions `processNodeTopDown` and `processNodeBottomUp`, defined in the base class, and are generally **not** overriden.

In the functions `processNodeTopDown` and `processNodeBottomUp`, a set of other functions are executed. They have a prefix _fwd_ if they are executed in `processNodeTopDown` and _bwd_ if executed in `processNodeBottomUp`. The functions _fwd_* and _bwd_* are generally overriden, but not all of them. By default, they do nothing.

In the functions `processNodeTopDown` and `processNodeBottomUp`, the functions _fwd_* and _bwd_* are surrounded by boilerplate code which is executed even if the functions _fwd_* and _bwd_* are not overriden, i.e. even if they do nothing.

Moreover, all the _fwd_* and _bwd_* functions act on objects or containers, stored in a Node. However, in the context of concurrent tasks, those objects and containers can change. In particular, adding an element to a container invalidates the iterators. It is a problem if a visitor was in the middle of a loop over the container. This problem can occur even if the visitor was not design to work on this specific container through the override of _fwd_* and _bwd_* functions, because of the boilerplate code.

In a nutshell, there are race conditions which can be avoided if the derived class can execute only the code they are designed to execute, and nothing more.

### Proposition

I added boolean variables named `applyFwd*` and `applyBwd*`, each corresponding to the `fwd*` and `bwd*` functions. They are all set to false by default.
For example, if `applyFwdMechanicalState` is true, then the code related to `fwdMechanicalState` is executed. Remember that previously, the boilerplate code related to `fwdMechanicalState` was executed even if `fwdMechanicalState` does nothing. To change that, the derived class must set the right variables to true, so the right code is executed. For example, if a derived class wants to execute `fwdMechanicalState`, it must set `applyFwdMechanicalState` first, usually in its constructor.

### Pros

- Avoid unnecessary race conditions
- Does not execute unnecessary boilerplate code
- Simple to understand

### Cons

- Breaks all classes derived from `BaseMechanicalVisitor`
- Forces the developers of classes derived from `BaseMechanicalVisitor` to tune booleans according the goal of the visitor.
- Runtime check of the variables

This PR is not to merge yet. IIt is suggested as an alternative to PR #1963

### DIFF
--- a/SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/MechanicalComputeEnergyVisitor.cpp
+++ b/SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/MechanicalComputeEnergyVisitor.cpp
@@ -33,6 +33,8 @@ MechanicalComputeEnergyVisitor::MechanicalComputeEnergyVisitor(const core::Mecha
     , m_kineticEnergy(0.)
     , m_potentialEnergy(0.)
 {
+    applyFwdMass = true;
+    applyFwdForceField = true;
 }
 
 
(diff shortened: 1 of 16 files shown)


alxbilger:
No longer relevant after #2040, #1992 and #2125

PR #1963, opened 2021-03-30
[SofaSimulationCore] Add virtual functions to the base class BaseMech…

alxbilger:
### Context

All instances of the derived classes of `BaseMechanicalVisitor` execute the same functions `processNodeTopDown` and `processNodeBottomUp`, defined in the base class, and are generally **not** overriden.

In the functions `processNodeTopDown` and `processNodeBottomUp`, a set of other functions are executed. They have a prefix _fwd_ if they are executed in `processNodeTopDown` and _bwd_ if executed in `processNodeBottomUp`. The functions _fwd_* and _bwd_* are generally overriden, but not all of them. By default, they do nothing.

In the functions `processNodeTopDown` and `processNodeBottomUp`, the functions _fwd_* and _bwd_* are surrounded by boilerplate code which is executed even if the functions _fwd_* and _bwd_* are not overriden, i.e. even if they do nothing.

Moreover, all the _fwd_* and _bwd_* functions act on objects or containers, stored in a Node. However, in the context of concurrent tasks, those objects and containers can change. In particular, adding an element to a container invalidates the iterators. It is a problem if a visitor was in the middle of a loop over the container. This problem can occur even if the visitor was not design to work on this specific container through the override of _fwd_* and _bwd_* functions, because of the boilerplate code.

In a nutshell, there are race conditions which can be avoided if the derived class can execute only the code they are designed to execute, and nothing more.

### Proposition

I added two groups of function in `BaseMechanicalVisitor`:
- `processNodeTopDown_fwd*` and `processNodeBottomUp_bwd*`: in the base class they do nothing. In the derived class, they must execute the corresponding `processNodeTopDown_fwd*_impl` or `processNodeBottomUp_bwd*_impl` function.
- `processNodeTopDown_fwd*_impl` and `processNodeBottomUp_bwd*_impl` which have a content and will execute the goal of the visitor. Those functions are not meant to be overriden by the derived classes. They are provided so derived class can call them when they override `processNodeTopDown_fwd*` or `processNodeBottomUp_bwd*`.

### Example

`MechanicalGetDimensionVisitor` inherits from `BaseMechanicalVisitor` and overrides `fwdMechanicalState`. In order for `fwdMechanicalState` to be executed, it must also override `processNodeTopDown_fwdMechanicalState`. In `processNodeTopDown_fwdMechanicalState`, it just have to call `processNodeTopDown_fwdMechanicalState_impl`.

### Pros

- Avoid unnecessary race conditions
- Does not execute unnecessary boilerplate code
- Benefit from inheritance and avoid runtime checks compared to PR #1962 

### Cons

- Breaks all classes derived from `BaseMechanicalVisitor`
- Forces the developers of classes derived from `BaseMechanicalVisitor` to derive more functions

This PR is not to merge yet. It is suggested as an alternative to PR #1962 

### DIFF
--- a/SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/MechanicalComputeEnergyVisitor.h
+++ b/SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/MechanicalComputeEnergyVisitor.h
@@ -49,9 +49,17 @@ public:
 
     SReal getPotentialEnergy();
 
+    Result processNodeTopDown_fwdMass(simulation::Node* node, VisitorContext* ctx) override
+    {
+        return processNodeTopDown_fwdMass_impl(node, ctx);
+    }
     /// Process the BaseMass
     Result fwdMass(simulation::Node* /*node*/, core::behavior::BaseMass* mass) override;
 
+    Result processNodeTopDown_fwdForceField(simulation::Node* node, VisitorContext* ctx) override
+    {
+        return processNodeTopDown_fwdForceField_impl(node, ctx);
+    }
     /// Process the BaseForceField
     Result fwdForceField(simulation::Node* /*node*/, core::behavior::BaseForceField* f) override;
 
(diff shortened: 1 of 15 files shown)


alxbilger:
I used this approach in my attempt to parallelize two tasks. It did not exhibit any crash due to race condition compared to the previous implementation of the visitor.

alxbilger:
No longer relevant after #2040, #1992 and #2125

PR #1973, opened 2021-04-01
[SofaUserInteraction] Fix Bug of removing topological element when a Hexa2TetraTopologicalMapping is in the scene

epernod:
Removing element from TriangleModel (using the mouse) when an Hexa2TetraTopologicalmapping is in the scene was not possible because of a searchRoot even if the mapping has nothing to see with the current topology being manipulated. 

### DIFF
--- a/modules/SofaUserInteraction/src/SofaUserInteraction/TopologicalChangeManager.cpp
+++ b/modules/SofaUserInteraction/src/SofaUserInteraction/TopologicalChangeManager.cpp
@@ -88,7 +88,7 @@ Index TopologicalChangeManager::removeItemsFromTriangleModel(sofa::component::co
     {
         //Quick HACK for Hexa2TetraMapping
         sofa::component::topology::Hexa2TetraTopologicalMapping* badMapping;
-        model->getContext()->get(badMapping, sofa::core::objectmodel::BaseContext::SearchRoot);
+        model->getContext()->get(badMapping, sofa::core::objectmodel::BaseContext::SearchUp);
         if(badMapping) //stop process
         {
             msg_warning("TopologicalChangeManager") << " Removing element is not handle by Hexa2TetraTopologicalMapping. Stopping process." ;
--- a/modules/SofaUserInteraction/src/SofaUserInteraction/TopologicalChangeManager.cpp
+++ b/modules/SofaUserInteraction/src/SofaUserInteraction/TopologicalChangeManager.cpp
@@ -217,7 +217,7 @@ Index TopologicalChangeManager::removeItemsFromPointModel(sofa::component::colli
     {
         //Quick HACK for Hexa2TetraMapping
         sofa::component::topology::Hexa2TetraTopologicalMapping* badMapping;
-        model->getContext()->get(badMapping, sofa::core::objectmodel::BaseContext::SearchRoot);
+        model->getContext()->get(badMapping, sofa::core::objectmodel::BaseContext::SearchUp);
         if (badMapping) //stop process
         {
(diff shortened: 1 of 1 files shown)


modules/SofaUserInteraction/src/SofaUserInteraction/TopologicalChangeManager.cpp
```diff
@@ -88,7 +88,7 @@ Index TopologicalChangeManager::removeItemsFromTriangleModel(sofa::component::co
     {
         //Quick HACK for Hexa2TetraMapping
         sofa::component::topology::Hexa2TetraTopologicalMapping* badMapping;
-        model->getContext()->get(badMapping, sofa::core::objectmodel::BaseContext::SearchRoot);
+        model->getContext()->get(badMapping, sofa::core::objectmodel::BaseContext::SearchUp);
```
hugtalbot, line 91: This hack (previously here) seems to compensate for the lack of topological change management in component. Do you know why ?

modules/SofaUserInteraction/src/SofaUserInteraction/TopologicalChangeManager.cpp
```diff
@@ -88,7 +88,7 @@ Index TopologicalChangeManager::removeItemsFromTriangleModel(sofa::component::co
...
         if(badMapping) //stop process
         {
```
hugtalbot, line 93: Moreover same hack seems to be [here](https://github.com/sofa-framework/sofa/pull/1973/files#diff-2353e48df7b7a635938eabe04814682ad1fc81caead416b9301e26ac39e3b770L217-L223)

epernod:
yes it compensate the Hexa2TetraTopologicalMapping which is not handling change: https://github.com/sofa-framework/sofa/blob/2a3a085dcc05834058db9308e08276f7e6dc8a48/modules/SofaTopologyMapping/src/SofaTopologyMapping/Hexa2TetraTopologicalMapping.cpp#L208

But this mapping is usually used at init to create for example a Tetrahedron topology from a grid.  Then you only want to simulate or change your tetrahedron topology. 
The idea of this "hack" in this component is top stop the process of removing topological element from the mouse if the topology you are changing is below a Hexa2TetraTopologicalMapping, because removing a triangle, will remove a tetra which should remove a Hexa.
What is fixing this PR is the fact that if you remove a tetrahedron using the mouse and a triangleModel, it won't stop because a Hexa2TetraTopologicalMapping has been encountered in the whole scene which is not linked to the topology you are manipulating.

In others words, before this PR, if you have in a scene an ObjectA roughly simulated using a grid + Tetra and you are trying to carve an ObjectB. This was not possible because of the SearchRoot

hugtalbot (approved):

fredroy:
@epernod 
(and @hugtalbot I saw that you approved)

I see that the second 
```c++
model->getContext()->get(badMapping, sofa::core::objectmodel::BaseContext::SearchRoot);
```
([line 220](https://github.com/sofa-framework/sofa/blob/facaed90bb8260d555aaed91d22b5835b99c92e5/modules/SofaUserInteraction/src/SofaUserInteraction/TopologicalChangeManager.cpp#L220))
is still with SearchRoot and you did not touch it, is it normal ?

modules/SofaUserInteraction/src/SofaUserInteraction/TopologicalChangeManager.cpp
hugtalbot: Applying the change on the line 223 is indeed needed

hugtalbot (changes requested):

epernod:
lavais po compris

fredroy (approved):

hugtalbot (approved):

PR #1977, opened 2021-04-01
[SofaBaseMechanics] Update MechanicalObject resize method to exit if not necessary

epernod:
Fix ```MechanicalObject``` resize method to resize all its vectorsCoord only if the new size is different from the current one. 

In a normal case, resizing a vector with it's current size won't have consequences but here as the VectorCoords are Data, they were all set to dirty each time a not needed resize was performed. And this on the position / rest_position/  freePosition and resetPosition.

For example in 
```
void BarycentricMapping<TIn, TOut>::apply(...
...
d_mapper->resize(this->toModel);

```
was always setting all MechanicalObject VectorCoords Data to dirty even if not needed. 


SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/MechanicalObject.inl
```diff
@@ -669,25 +669,25 @@ void MechanicalObject<DataTypes>::resize(const Size size)
 
     if(size>0)
     {
+        if (d_size.getValue() == static_cast<int>(size))
+            return;
```
hugtalbot: this is the only diff. And we love it !

epernod:
My guess is this PR is really a kick in the 💩 
I came up to this single line fix because Data update were looping between engine using dofs as input and forcefield and mapping....
In the list of failing test could you pick one that you know is supposed to work and I'll check what changed.

epernod:
[ci-build][with-all-tests]

epernod:
[ci-build][with-all-tests]

PR #1978, opened 2021-04-01
[SofaGeneralEngine] Add option in NearestPointROI to use restPosition or position

epernod:
Add option to choose between restPosition and position. Also update the init method to better check the inputs. 

### DIFF
--- a/modules/SofaGeneralEngine/src/SofaGeneralEngine/NearestPointROI.inl
+++ b/modules/SofaGeneralEngine/src/SofaGeneralEngine/NearestPointROI.inl
@@ -51,22 +51,47 @@ NearestPointROI<DataTypes>::~NearestPointROI()
 template <class DataTypes>
 void NearestPointROI<DataTypes>::init()
 {
-    if(!mstate1 || !mstate2) {
-        msg_error() << "Cannot Initialize without valid objects";
-        //mstate1->set(static_cast<simulation::Node*>(this->getContext())->getMechanicalState());
+    // Test inputs
+    bool success = true;
+    if (mstate1 && !mstate1.get())
+    {
+        msg_error_when(!mstate1.get()) << "Cannot Initialize, mstate1 link is pointing to invalid object!";
+        success = false;
     }
+    else if (!mstate1)
+    {
+        msg_error_when(!mstate1) << "Cannot Initialize, mstate1 link is invalid!";
+        success = false;
+    }
+
+    if (mstate2 && !mstate2.get())
+    {
+        msg_error_when(!mstate2.get()) << "Cannot Initialize, mstate2 link is pointing to invalid object!";
+        success = false;
+    }
+    else if (!mstate2)
+    {
+        msg_error_when(!mstate2) << "Cannot Initialize, mstate2 link is invalid!";
+        success = false;
+    }
+
+    if(!success) {
+        this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
+        return;
+    }
+
+
 
     addInput(this->mstate1->findData("rest_position"));
     addInput(this->mstate2->findData("rest_position"));
     addOutput(&f_indices1);
     addOutput(&f_indices2);
-
-    reinit();
 }
 
 template <class DataTypes>
 void NearestPointROI<DataTypes>::reinit()
 {
+    this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
     if(f_radius.getValue() <= 0){
         msg_error() << "Radius must be a positive real.";
         return;
@@ -75,6 +100,8 @@ void NearestPointROI<DataTypes>::reinit()
         msg_error() << "2 valid mechanicalobjects are required.";
         return;
     }
+
+    this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Valid);
     doUpdate();
 }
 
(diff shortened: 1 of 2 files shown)


modules/SofaGeneralEngine/src/SofaGeneralEngine/NearestPointROI.inl
```diff
@@ -75,11 +110,39 @@ void NearestPointROI<DataTypes>::reinit()
...
+
+        computeNearestPointMaps(x1, x2);
+    }
+    else
+    {
+        const VecCoord& x1 = this->mstate1->read(core::ConstVecCoordId::position())->getValue();
+        const VecCoord& x2 = this->mstate2->read(core::ConstVecCoordId::position())->getValue();
+
+        if (x1.size() == 0 || x2.size() == 0)
+            return;
```
hugtalbot: maybe add a comment here, before returning

hugtalbot: Like:
``` cpp
msg_warning() << "Size of one of the mechanical object position fields is zero";
```

hugtalbot (approved):

modules/SofaGeneralEngine/src/SofaGeneralEngine/NearestPointROI.inl
epernod: you will have this warning at init for sure, because at init the mechanicalObject are not necessary yet filled. 
this method should be called the first time positions are changed but I know this method is called at init from the updateVisitor or something like this.

Moreover if one mechanicalObject is empty this means the output map will be empty and it could be totally right. You could imagine using this engine between and full mesh and new points accumulating in a second mechanicalObject

modules/SofaGeneralEngine/src/SofaGeneralEngine/NearestPointROI.inl
```diff
@@ -51,22 +53,55 @@ NearestPointROI<DataTypes>::~NearestPointROI()
...
+        msg_error_when(!mstate2.get()) << "Cannot Initialize, mstate2 link is pointing to invalid object!";
+        success = false;
+    }
+    else if (!mstate2)
+    {
+        msg_error_when(!mstate2) << "Cannot Initialize, mstate2 link is invalid!";
+        success = false;
+    }
+
+    if(!success) {
```
alxbilger: Cosmetic: put curly brace on the next line and add space between `if` and the parenthesis

modules/SofaGeneralEngine/src/SofaGeneralEngine/NearestPointROI.inl
```diff
@@ -75,11 +110,39 @@ void NearestPointROI<DataTypes>::reinit()
...
+    else
+    {
+        const VecCoord& x1 = this->mstate1->read(core::ConstVecCoordId::position())->getValue();
+        const VecCoord& x2 = this->mstate2->read(core::ConstVecCoordId::position())->getValue();
+
+        if (x1.size() == 0 || x2.size() == 0)
+            return;
+
+        computeNearestPointMaps(x1, x2);
+    }
```
alxbilger: I suggest to factorize this code by introducing a variable which should be equal to `position()` or `restPosition()` depending on the value of `d_useRestPosition`.

modules/SofaGeneralEngine/src/SofaGeneralEngine/NearestPointROI.inl
```diff
@@ -93,24 +156,22 @@ void NearestPointROI<DataTypes>::doUpdate()
...
-    const VecCoord& x1 = this->mstate1->read(core::ConstVecCoordId::restPosition())->getValue();
-    const VecCoord& x2 = this->mstate2->read(core::ConstVecCoordId::restPosition())->getValue();
 
-    for (unsigned int i2=0; i2<x2.size(); ++i2)
+    for (unsigned int i2 = 0; i2 < x2.size(); ++i2)
     {
         pt2 = x2[i2];
         auto el = std::min_element(std::begin(x1), std::end(x1), cmp);
-        if(dist(*el, pt2) < maxR) {
+        if (dist(*el, pt2) < maxR) {
```
alxbilger: Cosmetic: curly brace on the next line

modules/SofaGeneralEngine/src/SofaGeneralEngine/NearestPointROI.inl
epernod: I'm not sure to get it. I'm already factorizing using the computeNearestPointMaps(x1, x2);
I could move the 
```
if (x1.size() == 0 || x2.size() == 0)
     return;
```
at the begining of the method indeed.

alxbilger: What I suggest:
```
const auto positionId = d_useRestPosition.getValue() ? core::ConstVecCoordId::restPosition() : core::ConstVecCoordId::position();
const VecCoord& x1 = this->mstate1->read(positionId)->getValue();
const VecCoord& x2 = this->mstate2->read(positionId)->getValue();

if (x1.size() == 0 || x2.size() == 0)
    return;

computeNearestPointMaps(x1, x2);
```
I reduce the amount of duplicated code by merging lines 123-129 and 133-139.

Would that work?

epernod: yes probably, I'll test 👍

PR #1979, opened 2021-04-01
[SofaDeformable] Fix test in StiffSpringForceField doUpdateInternal

epernod:
Also avoid to call the method at ```init``` as input indices might probably not be updated yet. 

### DIFF
--- a/SofaKernel/modules/SofaDeformable/src/SofaDeformable/StiffSpringForceField.inl
+++ b/SofaKernel/modules/SofaDeformable/src/SofaDeformable/StiffSpringForceField.inl
@@ -55,8 +55,6 @@ void StiffSpringForceField<DataTypes>::init()
     {        
         this->trackInternalData(d_indices1);
         this->trackInternalData(d_indices2);
-
-        createSpringsFromInputs();
     }
 
     this->SpringForceField<DataTypes>::init();
@@ -65,10 +63,11 @@ void StiffSpringForceField<DataTypes>::init()
 template<class DataTypes>
 void StiffSpringForceField<DataTypes>::doUpdateInternal()
 {
-    if (!d_indices1.isSet() && !d_indices2.isSet()) // nothing to do in this case
-        return;
-
-    createSpringsFromInputs();
+    
+    if (this->hasDataChanged(d_indices1) || this->hasDataChanged(d_indices2))
+    {
+        createSpringsFromInputs();
+    }
 }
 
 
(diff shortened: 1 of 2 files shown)


SofaKernel/modules/SofaDeformable/src/SofaDeformable/StiffSpringForceField.inl
```diff
@@ -55,8 +55,6 @@ void StiffSpringForceField<DataTypes>::init()
     {        
         this->trackInternalData(d_indices1);
         this->trackInternalData(d_indices2);
```
hugtalbot: This was the previous internal update mechanism isn't  it @damienmarchal @marques-bruno ?
I would rather implement the associated callback

epernod: this is used by UpdateVisitor. Is it deprecated?
The visitor is called at each step...

alxbilger:
[ci-build][with-all-tests]

SofaKernel/modules/SofaDeformable/src/SofaDeformable/StiffSpringForceField.inl
hugtalbot: We should indeed deprecate the doUpdateInternal function (and its underlying API).

The way to go is to use `addUpdateCallback` as it was done in #1806 for example. Define the input data, and the output data (updated by this internal process).

epernod: 👍

SofaKernel/modules/SofaDeformable/src/SofaDeformable/StiffSpringForceField.inl
```diff
@@ -36,6 +36,12 @@ template<class DataTypes>
 StiffSpringForceField<DataTypes>::StiffSpringForceField(double ks, double kd)
     : StiffSpringForceField<DataTypes>(nullptr, nullptr, ks, kd)
 {
+    this->addUpdateCallback("updateSprings-ks-kd", { &d_indices1, &d_indices2, &d_length, &this->ks, &this->kd}, [this](const core::DataTracker& t)
+    {
+        SOFA_UNUSED(t);
+        createSpringsFromInputs();
+        return sofa::core::objectmodel::ComponentState::Valid;
+    }, {&this->springs});
```
epernod: First constructor is calling the second one. So I'm not 100% sure but you can remove this one otherwise your callback will be register twice.

PR #1996, opened 2021-04-09
[SofaSparseSolver] SparseLDLSolver optimizations

fredroy:
Two optimizations:
 - ~~SparseLDL is now instanciated by default with the template `CompressedRowSparseMatrix<Matrix3>` instead of `CompressedRowSparseMatrix<double>` : seems much faster with 3DOFs~~ warn the user to use `CompressedRowSparseMatrix<Matrix3>` if using only 3DOFs
 -  Optimize compliance computation : use the local constraint count instead of the global to compute the internal matrices Also rename Jdense to JLinv and Jminv to JLinvDinv to be coherent with what the current code is computing

(those descriptions comes directly from the commit messages)

With my scene test (a dozen of objects with a SparseLDLSolver each, and a lot of constraints due to the FrictionContacts):
```
- before : 5000 iterations done in 208.437 s ( 23.9881 FPS).
- after  : 5000 iterations done in 75.8347 s ( 65.9328 FPS).
```
I tried to keep the commit identical to the original one from the [ISSofa repo](https://github.com/InSimo/ISSofa) (author/message/date)

Credits to:
 - @JeremieA from [Insimo](http://insimo.com/)
 - @remibessard from [Mimesis](https://mimesis.inria.fr), for the report.


hugtalbot (approved):

fredroy:
[ci-build][with-all-tests]

fredroy:
More precise results for each optimisation:
```
no cache + double: 5000 iterations done in 199.849 s ( 25.0188 FPS).
no cache + 3x3   : 5000 iterations done in 170.294 s ( 29.3609 FPS).
cache + double   : 5000 iterations done in 108.942 s ( 45.8959 FPS).
cache + 3x3      : 5000 iterations done in 79.4698 s ( 62.917 FPS).
```

jnbrunet:
Thanks a lot @fredroy .

Can you share your test scene?

Also, what happen if the default 3x3 block sparse matrix is used with something else than 3D DOFs? For example, with rigids?

fredroy:
We still did not test with Rigid or Vec2/6.
In my opinion we should be conservative and just let the caching optimization (and let the default template as it is for the moment)
What do you think ?

fredroy:
[ci-build][with-all-tests]

jnbrunet:
> what's your opinion @jnbrunet ?

Isn't a warning when the scalar template is used a bit intense? I would put a warning if we actually checked that all mechanical objects are of 3d types, and the specified/default template is a scalar. But here, we output a warning without any logic other than the default scalar template was used. Maybe at least remove the warning if the user manually specified that he needed the scalar template?

**EDIT** Sorry, last time I checked the PR, the check was done on the template used, hence the warning was emitted when the scalar template was used. Now it seems to be emitted when the user doesn't manually specify which template he/she wants.

jnbrunet (approved):

jnbrunet:
@fredroy there is the tag "do not squash". Do you want to clean up a bit those 13 commits, or we should keep them all?

fredroy:
Right I will clean the branch a bit

fredroy:
[ci-build][with-all-tests]

PR #1997, opened 2021-04-11
[SofaSimulationCore]  Set default  TaskScheduler worker threads to idle state

fspadoni:
Set worker threads to idle state when there are no more tasks to execute.
It should fix the problem of the CPU busy when the simulation is  not running raised in PR #1862  


SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/DefaultTaskScheduler.h
```diff
@@ -235,7 +235,7 @@ namespace sofa  {
             
             unsigned m_workerThreadCount;
             
-            bool m_workerThreadsIdle;
+            volatile bool m_workerThreadsIdle;
```
alxbilger, line 161: Can you explain why `volatile` is desirable here?

fspadoni: If the task scheduler system works correctly there should be only one thread modifying this variable and multiple threads reading it. The volatile prevents the multiple threads to cache this variable.

fredroy:
Linked to the PR #1862; can be merged IMO

fredroy:
I cant solve the conflicts for you @fspadoni ... (no rights to push into your branch unfortunately)
The conflict comes from #2002 and you just need to add your change into `WorkerThread.cpp` (instead of `DefaultTaskScheduler.cpp`)

PR #1999, opened 2021-04-12
[SofaBaseCollision] REFACTOR BruteForceDetection

alxbilger:
I propose a more readable code for the component BruteForceDetection. It is a crucial component, used in many Sofa scenes, and yet it is hard to understand what it does.
Among the changes I propose:
- Split giant functions into multiple smaller functions
- Give meaningful names to functions, variables and types
- Merge simple operations which was hard to read into simple expressions
- Add a lot of comments

The algorithm stays the same.

Note the number of lines is larger than before. It is the price for readability. It is mainly due to the several functions I added

I did not notice any performances drop.

### DIFF
--- a/SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/BruteForceDetection.cpp
+++ b/SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/BruteForceDetection.cpp
@@ -69,40 +69,23 @@ void BruteForceDetection::addCollisionModel(core::CollisionModel *cm)
         return;
 	assert(intersectionMethod != nullptr);
 
+	dmsg_info() << "CollisionModel " << cm->getName() << "(" << cm << ") of class " << cm->getClassName() << " is added in broad phase (" << collisionModels.size() << " collision models)";
+
     // If a box is defined, check that both collision models are inside the box
     // If both models are outside, ignore them
-    if (boxModel)
+    if (boxModel && !intersectWithBoxModel(cm))
     {
-        bool swapModels = false;
-        core::collision::ElementIntersector* intersector = intersectionMethod->findIntersector(cm, boxModel.get(), swapModels);
-        if (intersector)
-        {
-            core::CollisionModel* cm1 = (swapModels?boxModel.get():cm);
-            core::CollisionModel* cm2 = (swapModels?cm:boxModel.get());
-
-            // Here we assume a single root element is present in both models
-            if (!intersector->canIntersect(cm1->begin(), cm2->begin()))
-                return;
-        }
+        return;
     }
 
-    if (cm->isSimulated() && cm->getLast()->canCollideWith(cm->getLast()))
+    if (doesSelfCollide(cm))
     {
-        // self collision
-        bool swapModels = false;
-        core::collision::ElementIntersector* intersector = intersectionMethod->findIntersector(cm, cm, swapModels);
-        if (intersector != nullptr)
-            if (intersector->canIntersect(cm->begin(), cm->begin()))
-            {
-                cmPairs.push_back(std::make_pair(cm, cm));
-            }
+        cmPairs.emplace_back(cm, cm);
     }
 
     // Browse all other collision models to check if there is a potential collision (conservative check)
-    for (sofa::helper::vector<core::CollisionModel*>::iterator it = collisionModels.begin(); it != collisionModels.end(); ++it)
+    for (auto* cm2 : collisionModels)
     {
-        core::CollisionModel* cm2 = *it;
-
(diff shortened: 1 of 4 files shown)


alxbilger:
[ci-build][with-all-tests]

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/BruteForceDetection.h
```diff
@@ -61,18 +80,117 @@ class SOFA_SOFABASECOLLISION_API BruteForceDetection :
...
+    static bool isSelfCollision(core::CollisionModel* cm1, core::CollisionModel* cm2);
+
+    /// Build a list of TestPair's from internal and external children of two CollisionModel's
+    static void initializeExternalCells(
+            core::CollisionModel *cm1,
+            core::CollisionModel *cm2,
+            std::queue<TestPair>& externalCells);
+
+    /// Store data related to two finnest CollisionModel's
+    struct FinnestCollision
```
fredroy: Finest with one 'n' ...

fredroy:
Setting to "breaking" as some functions signatures changed

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/BruteForceDetection.cpp
```diff
@@ -71,76 +69,114 @@ void BruteForceDetection::addCollisionModel(core::CollisionModel *cm)
...
+            return intersector->canIntersect(cm->begin(), cm->begin());
+        }
+    }
+
+    return false;
+}
+
+bool BruteForceDetection::isSelfCollision(core::CollisionModel* cm1, core::CollisionModel* cm2)
+{
+    return (cm1->getContext() == cm2->getContext());
```
fredroy, line 158: Should we rely on the context to determine the fact there are on the same object ?

alxbilger: In BruteForceDetection, the broad phase checks if an object must be tested against itself:
https://github.com/sofa-framework/sofa/blob/339241a4a36e8d4ca0f44107421f32c58dcd8540/SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/BruteForceDetection.cpp#L81
, based on the value of `bSelfCollision`:
https://github.com/sofa-framework/sofa/blob/4026a23b534056afe180b1d250fbc4302feadb07/SofaKernel/modules/SofaCore/src/sofa/core/CollisionModel.cpp#L167

If the object must be tested against itself, then `cm1==cm2` in the function you mention. So I guess, the context is also the same.
If `cm1!=cm2`, I don't know if they both can have the same context.

In any case, I copied the code which was initially here:
https://github.com/sofa-framework/sofa/blob/4026a23b534056afe180b1d250fbc4302feadb07/SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/BruteForceDetection.cpp#L169

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/BruteForceDetection.h
alxbilger: Good catch! And it's not the only place with this typo

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/BruteForceDetection.cpp
fredroy: Ah I did not see that it was the same before.
I have some doubt about the test of "the same context" but if it was the same as before, this is not really the scope of the PR.
(and selfCollision in SOFA is... not really here, lets say 😅)

SofaKernel/modules/SofaCore/src/sofa/core/collision/BroadPhaseDetection.h
```diff
@@ -71,6 +71,7 @@ class BroadPhaseDetection : virtual public Detection
 
     /// Get the potentially colliding pairs detected
     sofa::helper::vector<CollisionModelPair>& getCollisionModelPairs() { return cmPairs; }
+    const sofa::helper::vector<CollisionModelPair>& getCollisionModelPairs() const { return cmPairs; }
```
hugtalbot, line 74: duplicating the function with a const and non-const version ? why ?

alxbilger: It allows to call `getCollisionModelPairs` in a const member of `BroadPhaseDetection` or on a const object of type `BroadPhaseDetection`.

PR #2009, opened 2021-04-15
[SofaUserInteraction] Minor cleaning of RayTraceDetection

alxbilger:
Initially, I wanted to differentiate the functions from the broad phase, from the functions from the narrow phase, in the header file.
`void selfCollision (TriangleOctreeModel * cm1);` has been removed since it is not implemented. Other modifications are minor.

### DIFF
--- a/modules/SofaUserInteraction/src/SofaUserInteraction/RayTraceDetection.cpp
+++ b/modules/SofaUserInteraction/src/SofaUserInteraction/RayTraceDetection.cpp
@@ -62,6 +62,11 @@ RayTraceDetection ():bDraw (initData
 {
 }
 
+void RayTraceDetection::beginBroadPhase()
+{
+    core::collision::BroadPhaseDetection::beginBroadPhase();
+    collisionModels.clear();
+}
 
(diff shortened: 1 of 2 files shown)


modules/SofaUserInteraction/src/SofaUserInteraction/RayTraceDetection.cpp
```diff
@@ -240,10 +245,8 @@ void RayTraceDetection::addCollisionModel (core::CollisionModel * cm)
 {
     if (cm->empty ())
         return;
-    for (sofa::helper::vector < core::CollisionModel * >::iterator it =
-            collisionModels.begin (); it != collisionModels.end (); ++it)
+    for (auto* cm2 : collisionModels)
```
fredroy, line 248: auto* will not be deduced as core::CollisionModel** ? (as collisionModels is a vector of CollisionModel*)

alxbilger: No, I don't think so. See https://stackoverflow.com/a/34560387

fredroy: 👍

PR #2010, opened 2021-04-15
[SofaBaseCollision] Introduce brute force broad phase

alxbilger:
# This pull request

I introduce a new component: `BruteForceBroadPhase`

This component should not be confused with `BruteForceDetection`. `BruteForceDetection` is two components at once: a broad phase and a narrow phase. I extracted the broad phase code from `BruteForceDetection` and put it in `BruteForceBroadPhase`. `BruteForceBroadPhase` is only a broad phase. There is no narrow phase code inside.

The broad phase code extracted from `BruteForceDetection` is taken in #1999.

Looking at `RayTraceDetection`, it is also a mix of a broad phase and narrow phase. In this component, the broad phase is actually a brute force broad phase. The broad phase in `RayTraceDetection` did exactly the same thing than the broad phase in `BruteForceDetection`. It means duplicated code. With `BruteForceBroadPhase` we have an opportunity to factorize the broad phase by inheriting from `BruteForceBroadPhase` instead of `BroadPhaseDetection`. This is what I did in `RayTraceDetection`. Doing that, I could remove all the broad phase code from `RayTraceDetection`.

# What comes next

It is also my intention to inherit from `BruteForceBroadPhase` in `BruteForceDetection`, and remove all the broad phase code from `BruteForceDetection`. The broad phase code from `BruteForceDetection` and `RayTraceDetection` would be shared from `BruteForceBroadPhase`. However, I expect complex conflicts because of #1999. I'll do that if #1999 is accepted and merged.

It is also my intention to extract the narrow phase from `BruteForceDetection` and put it in another component which would inherit from `NarrowPhaseDetection`. The narrow phase in `BruteForceDetection` is not brute force, and should be named with a clearer and more meaningful name. It will be the case with the new component.

I also believe SAP can be implemented as a broad phase only component. We would have the choice between `BruteForceBroadPhase` and `SAPBroadPhase`. Any narrow phase would work with both broad phases (`BruteForceBroadPhase` and `SAPBroadPhase`).

Finally, with all those changes, we can consider more modularity on the collision detection and choose any combination of broad phase and narrow phase.


alxbilger:
[ci-build][with-all-tests]

PR #2030, opened 2021-04-20
[SofaGeneralMeshCollision] Direct SAP is written as a narrow phase

alxbilger:
SAP is usually a method to discriminate pairs of objects that are not in collision. The output is pairs of objects to be further investigated in the narrow phase.
However, in SOFA, the `DirectSAP` component is written as a component doing the broad phase and the narrow phase. Almost all the work is actually done in the narrow phase. That is why, we can consider this component is mainly a narrow phase component.
All the broad phase code has been removed. It is still a broad phase, but the code is located in its base class: `BruteForceBroadPhase`.
The narrow phase code has been extracted and put in a new component: `DirectSAPNarrowPhase`. `DirectSAP` inherits from `DirectSAPNarrowPhase`.
In the narrow phase, the result of the broad phase is used to skip the boxes which have been totally ignored in the broad phase. Initially, the `DirectSAP` component did not perform any discrimination in its broad phase. Now it performs a `BruteForceBroadPhase`. It is not yet clear if it is always a good choice regarding the performances. In the case of the caduceus scene, the performances are slightly better. I guess if the broad phase takes a lot of time (many objects), it is better to avoid it. `DirectSAP` would not be the best choice. I would use a more performant broad phase, or no broad phase, when used with `DirectSAPNarrowPhase`.

From now, the user can decide to design its scene with two different ways resulting to the same behavior:
```
<DirectSAP/>
```
or
```
<BruteForceBroadPhase/>
<DirectSAPNarrowPhase/>
```
In the second option, it will be possible to change the combination of broad phase/narrow phase.
Note: I kept the component `DirectSAP`, but I don't think its name is meaningful.

This is the following of the work started in #2010 


alxbilger:
[ci-build][with-all-tests]

hugtalbot (approved):

PR #2040, opened 2021-04-22
[SofaConstraint] Parallel free motion and collision detection

alxbilger:
In `FreeMotionAnimationLoop`, the collision detection and the free motion can be computed concurrently. A Data<bool> allows to specify if both steps are executed in parallel or in sequence. By default, the algorithm is sequential.

This parallel approach is the most useful when free motion and collision detection takes about the same time. An example scene is provided to validate the speed up. The details about the speed up are here: https://github.com/sofa-framework/sofa/commit/0f94b063ae3f3efbd2e0dc2f50d4bff706cd4734:

> On my computer, with the single threaded scene, FreeMotion+CollisionDetection runs in 24.92 ms in average. FreeMotion runs in 7.7 ms and CollisionDetection in 16.36 ms. If these two steps runs in parallel, we expect a gain of 7.7 ms (the minimum between 7.7 and 16.36).
On my computer, with the multi-threaded scene, FreeMotion+CollisionDetection runs in 16.93 ms in average. It is a gain of 7.99 ms (=24.92-16.93), which is about what we expected.
It validates the implementation of the parallelization of the collision detection and the free motion steps in the FreeMotionAnimationLoop.

cubes


examples/Benchmark/Performance/benchmark_cubes.pscn
```diff
@@ -0,0 +1,84 @@
...
+
+
+    <VisualStyle displayFlags="showBehavior showCollisionModels" />
+    <OglSceneFrame/>
+
+    <!-- Basic Components to perform the collision detection -->
+    <FreeMotionAnimationLoop name="FreeMotionAnimationLoop" parallel="true" />
+    <DefaultPipeline name="CollisionPipeline" />
+    <BruteForceDetection name="BruteForceDetection" />
+    <LocalMinDistance name="Proximity" alarmDistance="0.2" contactDistance="0.09" angleCone="0.0" />
```
hugtalbot, line 21: Here @alxbilger, with Rigid squares on top of another, you should open a bit the `coneFactor` to avoid small oscillatory motions as you are getting in your video
See le [LocalMinDistance doc page](https://www.sofa-framework.org/community/doc/components/collisions/localmindistance/).

examples/Benchmark/Performance/benchmark_cubes.scn
```diff
@@ -0,0 +1,1550 @@
...
+
+
+    <VisualStyle displayFlags="showBehavior showCollisionModels" />
+    <OglSceneFrame/>
+
+    <!-- Basic Components to perform the collision detection -->
+    <FreeMotionAnimationLoop name="FreeMotionAnimationLoop" parallel="true" />
+    <DefaultPipeline name="CollisionPipeline" />
+    <BruteForceDetection name="BruteForceDetection" />
+    <LocalMinDistance name="Proximity" alarmDistance="0.2" contactDistance="0.09" angleCone="0.0" />
```
hugtalbot, line 19: obviously same same

modules/SofaConstraint/src/SofaConstraint/FreeMotionAnimationLoop.cpp
```diff
@@ -185,32 +211,7 @@ void FreeMotionAnimationLoop::step(const sofa::core::ExecParams* params, SReal d
...
-        getContext()->executeVisitor(&freePosEqPosPlusFreeVelDt);
-    }
-    dmsg_info() << " SolveVisitor for freeMotion performed" ;
-
-    // Collision detection and response creation
-    {
-        ScopedAdvancedTimer timer("Collision");
-        computeCollision(params);
-    }
+    FreeMotionAndCollisionDetection(params, cparams, dt, pos, freePos, freeVel, &mop);
```
hugtalbot, line 214: when changing the lifecycle of such important AnimationLoop we have to keep in mind to update the doc : 

https://www.sofa-framework.org/community/doc/components/animationloops/freemotionanimationloop/
&
https://raw.githubusercontent.com/sofa-framework/doc/master/images/animationloop/FreeMotionAnimationLoop.png

examples/Benchmark/Performance/benchmark_cubes.pscn
alxbilger: I don't understand how I should use this parameter. It is a factor, and yet the default value is 0.5. I expected it to be 1. In the case of the rigid cubes, what value do you suggest? More than 1?
If I understand it correctly, the goal is to open the angle of the cone, but not too much to avoid too many contacts. Right?

modules/SofaConstraint/src/SofaConstraint/FreeMotionAnimationLoop.cpp
alxbilger: Yep, I'll do it when/if this PR is accepted

alxbilger:
[ci-build][with-all-tests]

examples/Benchmark/Performance/benchmark_cubes.pscn
hugtalbot: Hi @alxbilger I must admit it is confusing. I got into the code [LocalMinDistance.cpp](https://github.com/sofa-framework/sofa/blob/b5a6f19a704cacb01a7482a5c1d3e3220ac99450/modules/SofaConstraint/src/SofaConstraint/LocalMinDistance.cpp#L1274) and I thought it came from the fact only the angle is considered as:



but we definitely to have a validation because the tests I did worked even with: ` angleCone="0.0" coneFactor="0.0005" `, weird ..

PR #2047, opened 2021-04-27
[SofaKernel] Switch from map_ptr_stable_compare to std::unordered_map

alxbilger:
This is a discussion about the need for `sofa::helper::map_ptr_stable_compare`, in `NarrowPhaseDetection`, but also in Sofa in general (it is also used in `DefaultContactManager`).

According to benchmarks comparing 3 different types of maps, `sofa::helper::map_ptr_stable_compare` is several orders of magnitude slower than `std::map` and `std::unordered_map`.

## `std::unordered_map`
```
------------------------------------------------------------------------------------------
Benchmark                                                Time             CPU   Iterations
------------------------------------------------------------------------------------------
BM_NarrowPhaseDetection_getDetectionOutputs/8         19.8 us         19.7 us        32412
BM_NarrowPhaseDetection_getDetectionOutputs/16        42.8 us         42.8 us        16528
BM_NarrowPhaseDetection_getDetectionOutputs/32         121 us          121 us         5764
BM_NarrowPhaseDetection_getDetectionOutputs/64         401 us          401 us         1750
BM_NarrowPhaseDetection_getDetectionOutputs/128       2228 us         2228 us          316
BM_NarrowPhaseDetection_getDetectionOutputs/256       9252 us         9253 us           75
```
## `std::map`
```
------------------------------------------------------------------------------------------
Benchmark                                                Time             CPU   Iterations
------------------------------------------------------------------------------------------
BM_NarrowPhaseDetection_getDetectionOutputs/8         18.4 us         18.4 us        36921
BM_NarrowPhaseDetection_getDetectionOutputs/16        54.1 us         54.1 us        12873
BM_NarrowPhaseDetection_getDetectionOutputs/32         177 us          177 us         3973
BM_NarrowPhaseDetection_getDetectionOutputs/64         631 us          631 us         1116
BM_NarrowPhaseDetection_getDetectionOutputs/128       2587 us         2587 us          268
BM_NarrowPhaseDetection_getDetectionOutputs/256      10649 us        10649 us           67
```
## `sofa::helper::map_ptr_stable_compare`
```
------------------------------------------------------------------------------------------
Benchmark                                                Time             CPU   Iterations
------------------------------------------------------------------------------------------
BM_NarrowPhaseDetection_getDetectionOutputs/8         38.3 us         38.3 us        18336
BM_NarrowPhaseDetection_getDetectionOutputs/16         234 us          234 us         3010
[...]

PR #2067, opened 2021-05-07
[SofaCore] Revert #2047 for deterministic contacts

alxbilger:
In this PR, `map_ptr_stable_compare` is preferred over `unordered_map`, as it allows reproducible contact response. However, `unordered_map` is still a lot faster.

#2047 leads to a crash in `examples/Components/collision/RayTraceCollision.scn`. I suspect the cause of this crash existed before #2047, and that #2047 revealed the crash by processing the contacts in a non-deterministic order. With the non-deterministic order, I can see that the crash does not happen always at the same current time step. It even happened that there was no crash.

Currently, I am not able to fix the crash. But I will continue investigating. As a workaround I restore the stable map to have deterministic contacts. IMO, it hides the cause of the crash.

In the end, I think the stable map should be proposed to the user (as a parameter?) by default. Let's see if we can accelerate it (perhaps with [Insimo version](https://github.com/InSimo/ISSofa/commits/issofa/framework/sofa/helper/map_ptr_stable_compare.h)).

### DIFF
--- a/SofaKernel/modules/SofaCore/src/sofa/core/collision/NarrowPhaseDetection.h
+++ b/SofaKernel/modules/SofaCore/src/sofa/core/collision/NarrowPhaseDetection.h
@@ -41,18 +41,13 @@ class SOFA_CORE_API NarrowPhaseDetection : virtual public Detection
 public:
     SOFA_ABSTRACT_CLASS(NarrowPhaseDetection, Detection);
 
-private:
-
-    struct CollisionModelPairHash
-    {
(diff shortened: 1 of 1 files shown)


fredroy:
[ci-build][with-all-tests]

fredroy:
Should be apply it right now, and discuss it in a devmeeting ? or would you want to wait to discuss before merging ?
(IMO we should apply it now)

alxbilger:
We can apply it now. It will satisfy the CI.
In the meantime, I continue working on the issue.

PR #2076, opened 2021-05-11
[SofaMiscCollision] Fix multiple bugs in group manager

alxbilger:
The initial objective of this Pull Request was to fix the scene crash detected in #2047 (https://www.sofa-framework.org/dash/?branch=origin/PR-2047).

What this PR really does:
1. Clean and make `DefaultCollisionGroupManager` and `DefaultContactManager` more readable
2. Fix absence of collision response when using `DefaultCollisionGroupManager` with a stationary object. A scene test is provided
3. Fix freeze happening in `examples/Components/collision/RayTraceCollision.scn` after some time
4. Fix crash detected in #2047. 

Everything is related to `DefaultCollisionGroupManager`.

Note:
In the collision pipeline, the group manager responsibilities are:
1. Combine `Node`s which are involved in a contact (which has been detected before). It is a manipulation of the scene graph which depends on the collision detection.
2. Create the contact response

# Cleaning
- Split large functions into smaller to make them more readable
- More null pointer checks
- Add lots of comments
- Minor cleaning such as namespace concatenation, ...

# Fix absence of collision response with stationary object
The contact with stationary objects are supposed to be handled by the collision pipeline before creating groups with the group manager. In the group manager, it supposes all the objects in contacts are associated with an ODE solver (since they cannot be stationary). Actually, the collision pipeline check if an object is stationary based on the value of `isSimulated()`. It means it's up to the user to set this Data. If it not done, the object is considered moving and goes to the group manager. The ODE solver is not found, so it is skipped:
https://github.com/sofa-framework/sofa/blob/477747e21ae520fba08cb08f7b66e6f4250a91f8/applications/plugins/SofaMiscCollision/src/SofaMiscCollision/DefaultCollisionGroupManager.cpp#L91
And collision response is not created.
In this PR, this case is not skipped and a collision response is created.

A scene test has been added to check collision is working with a stationary object.

# Fix freeze
This freeze happened before #2047.
It is due to an infinite loop here:
https://github.com/sofa-framework/sofa/blob/477747e21ae520fba08cb08f7b66e6f4250a91f8/applications/plugins/SofaMiscCollision/src/SofaMiscCollision/DefaultCollisionGroupManager.cpp#L235
This happened because a map was not updated after some merge operations. And we relied on the value in the map, for example here:
https://github.com/sofa-framework/sofa/blob/477747e21ae520fba08cb08f7b66e6f4250a91f8/applications/plugins/SofaMiscCollision/src/SofaMiscCollision/DefaultCollisionGroupManager.cpp#L137

# Fix crash detected in #2047

## Before #2047
The scene freezes always at the same time. In #2067 I explain that the simulation is deterministic. Now that the freeze is fixed, we can notice a crash still occurs, but after the time step of the freeze.

## After #2047 
A crash occurs, but not always at the same time step. This is because the simulation is no longer deterministic (because of #2047). 

Conclusion: #2047 revealed this crash, but was not the cause. It means that a non-deterministic simulation is not a source of crash.

## Fix
The crash is fixed also by updating the values of the map, similarly to the freeze fix.


SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/DefaultContactManager.cpp
```diff
@@ -122,24 +114,44 @@ void DefaultContactManager::changeInstance(Instance inst)
...
-        if (contactInsert.second)
+        if (contactInsert.second) //insertion success
         {
             // new contact
-            CollisionModel* model1 = outputsIt->first.first;
-            CollisionModel* model2 = outputsIt->first.second;
+            core::CollisionModel* model1 = outputsIt->first.first;
+            core::CollisionModel* model2 = outputsIt->first.second;
+
+            msg_error_when(model1 == nullptr) << "Contact found with a nullptr collision model";
```
jnbrunet: Too me, an error message usually tagets users, not developers. How can a collision model be null? Can this be avoided by the user setting things differently in its sene? Or does it only happen when another component malfunction?

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/DefaultContactManager.cpp
```diff
@@ -199,72 +191,93 @@ void DefaultContactManager::createContacts(const DetectionOutputMap& outputsMap)
...
+{
     for (ContactMap::iterator contactIt = contactMap.begin(), contactItEnd = contactMap.end();
-        contactIt != contactItEnd;)
+         contactIt != contactItEnd;)
     {
-        bool remove = false;
-        DetectionOutputMap::const_iterator outputsIt = outputsMap.find(contactIt->first);
-        if (outputsIt == outputsMap.end())
+        core::collision::Contact::SPtr contact = contactIt->second;
+        msg_error_when(contact == nullptr) << "Reading nullptr contact";
```
jnbrunet: Again here. What is the user supposed to understand here? Should a bug report be made?

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/DefaultContactManager.cpp
```diff
@@ -199,72 +191,93 @@ void DefaultContactManager::createContacts(const DetectionOutputMap& outputsMap)
...
             nbContactsMap[cms.second]++;
     }
 
-    // TODO: this is VERY inefficient, should be replaced with a visitor
-    helper::vector< CollisionModel* > collisionModels;
-    simulation::Node* context = dynamic_cast< simulation::Node* >(getContext());
-    context->getTreeObjects< CollisionModel >(&collisionModels);
-
-    for (unsigned int i = 0; i < collisionModels.size(); ++i)
+    for (const auto& [collisionModel, nbContacts] : nbContactsMap)
```
jnbrunet, line 274: Nice catch ;)

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/DefaultPipeline.cpp
```diff
@@ -90,19 +92,23 @@ void DefaultPipeline::doCollisionReset()
...
-    core::objectmodel::BaseContext* scene = getContext();
     // clear all contacts
     if (contactManager!=nullptr)
     {
         const helper::vector<Contact::SPtr>& contacts = contactManager->getContacts();
-        for (helper::vector<Contact::SPtr>::const_iterator it = contacts.begin(); it!=contacts.end(); ++it)
+        for (const auto& contact : contacts)
         {
-            (*it)->removeResponse();
+            if (contact != nullptr)
```
jnbrunet, line 116: Could you remember me how does the contact manager can return null contacts? Shouldn't be its job to make sure returned contacts are valid?

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/DefaultPipeline.cpp
```diff
@@ -215,12 +221,18 @@ void DefaultPipeline::doCollisionResponse()
 {
     core::objectmodel::BaseContext* scene = getContext();
     // then we start the creation of contacts
-    if (contactManager==nullptr) return; // can't go further
+    if (contactManager==nullptr)
+    {
+        msg_warning() << "Cannot compute collision response without a ContactManager";
```
jnbrunet: A user is reading this. Maybe a warning of the style "A ContactManager component is required to compute collision respones and was not found in the current scene." or something like this would be easier to understand...

SofaKernel/modules/SofaCore/src/sofa/core/collision/NarrowPhaseDetection.cpp
```diff
@@ -79,14 +79,17 @@ auto NarrowPhaseDetection::getDetectionOutputs() const -> const DetectionOutputM
 
 DetectionOutputVector*& NarrowPhaseDetection::getDetectionOutputs(CollisionModel *cm1, CollisionModel *cm2)
 {
+    msg_error_when(cm1 == nullptr) << "Requested a collision detection result with nullptr first collision model";
```
jnbrunet, line 82: Again, rework the error message. Is this really a user error?

SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/BaseNode.cpp
```diff
@@ -77,8 +77,17 @@ void BaseNode::setObjectContext(BaseObject::SPtr obj)
...
+    if (obj != nullptr)
+    {
+        if (obj->getContext() == this->getContext())
+        {
+            obj->l_context.reset();
+        }
+    }
+    else
+    {
+        msg_warning() << "Trying to clear a nullptr object context";
```
jnbrunet, line 89: To me this should throw an exception... This is clearly the caller's job to check this and print an error message if it knows why it happened.

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/Node.cpp
```diff
@@ -602,11 +602,13 @@ bool Node::doAddObject(BaseObject::SPtr sobj)
 /// Remove an object
 bool Node::doRemoveObject(BaseObject::SPtr sobj)
 {
+    msg_warning_when(sobj == nullptr) << "Trying to remove a nullptr object";
```
jnbrunet: Throw?

applications/plugins/SofaMiscCollision/src/SofaMiscCollision/DefaultCollisionGroupManager.cpp
```diff
@@ -285,8 +304,132 @@ simulation::Node* DefaultCollisionGroupManager::getIntegrationNode(core::Collisi
...
+bool DefaultCollisionGroupManager::checkEndlessLoop(const MergeGroupsMap& mergedGroups, simulation::Node* firstGroup, simulation::Node* currentGroup)
+{
+    if (currentGroup == firstGroup)
+    {
+        std::stringstream errorStream;
+        for (const auto& [a, b] : mergedGroups)
+        {
+            errorStream << a << " " << b << "\n";
+        }
+        msg_fatal() << "Detected an endless loop:\n" << errorStream.str();
```
jnbrunet: Throw?

jnbrunet (commented):
Good PR Alex. It must not have been easy to understand all that moving nodes/solvers around to get them under the same "group", although I'm not really a fan of this whole "SOFA will decide which solver has precedence and automatically merge and move things around for the user" thing...

Anyway, the only comments I had were related to messages displayed to users. You added super useful and clear comments for devs directly inside the source code, but maybe messages intended to users seems to have got a little bit less of love here?

My take would be:
 1. Can the user do something about it ? -> msg_error/msg_warning explaining to him what he did wrong and how to fix it.
 2. Else, throw an exception. The user should send us a bug report so that we can:
    a. Fix it if it is a bug
    b. Display a msg_error to the user explaining to him that the X component is incompatible with the Y component, or whatever the reason that created the error in the first place and how to avoid it (as a user).

fredroy:
[Related to #2089] 
this does not fix the multiple creation of vectors at every timestep in the mechanical objects.

alxbilger:
> [Related to #2089]
> this does not fix the multiple creation of vectors at every timestep in the mechanical objects.

Ok, thank you for having done the test.

applications/plugins/SofaMiscCollision/src/SofaMiscCollision/DefaultCollisionGroupManager.cpp
alxbilger: I'll change the message to the user.
I am surprised `msg_fatal` does not throw an exception or exit the program.
In my case, I call `std::exit`. What do you think?

alxbilger:
[ci-build][with-all-tests]

applications/plugins/SofaMiscCollision/src/SofaMiscCollision/DefaultCollisionGroupManager.cpp
jnbrunet: I would throw an exception so that you can easily debug where the error came from the call stack. It should then be the role of the caller (found using the call stack) to prevent calling this method with invalid parameters, and suggesting a warning or error message to the user. If it isn't able to do it, than go up again it the call stack until some method has enough "high level" knowledge to tell the user what he did wrong.

SofaKernel/modules/SofaBaseCollision/src/SofaBaseCollision/DefaultPipeline.cpp
alxbilger: The `DefaultContactManager` makes sure there is no null contacts when creating them (phew). Does it mean I can trust him? No. What about `RuleBasedContactManager`? I don't know this component. There is no contract visible to `DefaultPipeline` so a ContactManager keeps only non-null contacts. It is also possible they have been nullified since the contact manager created them.
It's always a good practice to check nullptr. We can never anticipate that a pointer is always valid.

alxbilger:
Thanks @jnbrunet. I made some messages clearer. Some of them was meant to facilitate debugging (it helped me) and are not necessarily meant for the user. I converted them to developers messages (but still made them clearer).

PR #2112, opened 2021-05-21
[MultiThreading] ADD scene showing the perf improvements

alxbilger:
A scene is added in the MultiThreading plugin. Its goal is to show the difference between the parallel collision detection algorithms and the single threaded equivalents, in terms of performances.

My results after 1000 time steps:

# Single thread
43.6893 s ( 22.8889 FPS).  
`doCollisionDetection` runs in ~28.99 ms (mean)

# Multi thread
22.9585 s ( 43.5568 FPS).  
`doCollisionDetection` runs in ~7.81 ms (mean)


applications/plugins/MultiThreading/examples/ParallelCollisionDetection.scn
```diff
@@ -0,0 +1,96 @@
...
+    </Node>
+
+
+    <Node name="Floor">
+        <MeshObjLoader name="loader" filename="mesh/floor2b.obj"/>
+        <MeshTopology src="@loader"/>
+        <MechanicalObject src="@loader" dy="-10" scale="0.5"/>
+        <TriangleCollisionModel simulated="false"/>
+        <LineCollisionModel simulated="false"/>
+        <PointCollisionModel simulated="false"/>
```
hugtalbot: could be `moving="false"` as well

hugtalbot (approved):

applications/plugins/MultiThreading/examples/ParallelCollisionDetection.scn
```diff
@@ -0,0 +1,96 @@
...
+    <!--
+    Parallel collision detection.
+    To compare to the single thread collision detection, replace the two following components with:
+    <BruteForceBroadPhase/>
+    <BVHNarrowPhase/>
+    -->
+    <ParallelBruteForceBroadPhase/>
+    <ParallelBVHNarrowPhase/>
+
+    <NewProximityIntersection name="Proximity" alarmDistance="2" contactDistance="0.7"/>
```
hugtalbot: Should not we use :
- MinProximityIntersection
- or LocalMinDistance 
instead?

hugtalbot: NewProximityIntersection is optimized for triangle proximities

applications/plugins/MultiThreading/examples/ParallelCollisionDetection.scn
alxbilger: `moving="false"` is indeed appropriate here. It prevents to recompute the bounding tree. However, it is a negligible optimization for this collision model (4 triangles). I assume the optimization is desirable for highly detailed meshes.

applications/plugins/MultiThreading/examples/ParallelCollisionDetection.scn
alxbilger: MinProximityIntersection is faster than NewProximityIntersection. Since it is faster, the difference between the single thread and the multi thread scene is reduced.
LocalMinDistance gives incorrect results.

hugtalbot (approved):

PR #2125, opened 2021-05-26
[SofaSimulationCore] Simplify repetitive code in BaseMechanicalVisitor

alxbilger:
In this PR:

- Added static class member `Visitor::runVisitorTask`
- Used `Visitor::runVisitorTask` in BaseMechanicalVisitor to replace all the boilerplate code
- Simplify the logic in `BaseMechanicalVisitor::processNodeTopDown` even though I have some concerns
- `Visitor::for_each` is now static
- `Visitor::for_each` accepts an optional string parameter giving the type of visitor "fwd" or "bwd"
- `Visitor::for_each_r` is removed in favour of an overload of `Visitor::for_each`
- SOFA_VERBOSE_TRAVERSAL surrounds the calls of `debug_write_state_before` and `debug_write_state_after`, instead of the functions themselves. Doing that, it allows to actually compile the debug content of both function. And I noticed that they had compilation errors (we could not detect it as SOFA_VERBOSE_TRAVERSAL is never defined).

My concerns:

1. I wanted to use `Visitor::runVisitorTask` in other visitors than BaseMechanicalVisitor. But the logic is not the same. BaseMechanialVisitor is the only visitor checking the tags. Should we always check the tags? Also, the debug boilerplate code is not consistent over all the visitors.
2. `Visitor::for_each` returns only the result of the last call in the loop. In another loop on the solvers at the beginning of `BaseMechanicalVisitor::processNodeTopDown`, we check the intermediate values. It is not consistent. I suspect that `Visitor::for_each` must also check intermediate values.
3. In `BaseMechanicalVisitor::processNodeTopDown`, between the process of `mechanicalMapping` and `mechanicalState`, we don't check if `res == RESULT_PRUNE`. It means we do the process of `mechanicalState` even if the process of `mechanicalMapping` returned `RESULT_PRUNE`. It's not consistent with the rest of the code.
4. It is not possible to define SOFA_VERBOSE_TRAVERSAL through cmake


alxbilger:
[ci-build][with-all-tests]

alxbilger:
Unit and scene tests are not happy. Setting back to WIP

alxbilger:
[ci-build][with-all-tests]

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/BaseMechanicalVisitor.cpp
```diff
@@ -46,199 +46,123 @@ namespace sofa::simulation
...
-                if(testTags(node->mechanicalState))
+                if (testTags(node->mechanicalState))
                 {
-                    ctime_t t = begin(node, node->mechanicalState, "bwd");
-                    this->bwdMappedMechanicalState(ctx, node->mechanicalState);
-                    end(node, node->mechanicalState, t);
-                    t = begin(node, node->mechanicalMapping, "bwd");
-                    this->bwdMechanicalMapping(ctx, node->mechanicalMapping);
-                    end(node, node->mechanicalMapping, t);
+                    runVisitorTask(this, ctx, &BaseMechanicalVisitor::bwdMappedMechanicalState, &*node->mechanicalState, bwdVisitorType);
```
fredroy, line 146: Maybe move the first runVisitorTask on mappedMS out of the tag test (as runVisitor will also test the tags)...
but this comment is mostly to show I did a thorough review 😁

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/BaseMechanicalVisitor.h
```diff
@@ -240,6 +240,10 @@ class SOFA_SIMULATION_CORE_API BaseMechanicalVisitor : public Visitor
     sofa::helper::vector< sofa::core::ConstMultiVecId > readVector;
     sofa::helper::vector< sofa::core::MultiVecId > writeVector;
 #endif
+
+private:
+    static const std::string fwdVisitorType;
```
fredroy, line 245: it wont change anything I suppose but constexpr may be better in this case ? (and you can also inline initialize the value here)

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/Visitor.h
```diff
@@ -235,8 +230,79 @@ class SOFA_SIMULATION_CORE_API Visitor
...
+    for (const auto& element : list)
+    {
+        res = runVisitorTask(visitor, ctx, task, &*element, typeInfo);
+    }
+    return res;
+}
+
+} // namespace sofa::simulation
 
 #endif
```
fredroy: if you go for a clean, you may remove the ifdef/endif guard with a pragma

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/BaseMechanicalVisitor.cpp
alxbilger: In any case, the tag test will be performed twice

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/BaseMechanicalVisitor.h
alxbilger: constexpr is not allowed on strings. It is allowed from C++20

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/BaseMechanicalVisitor.cpp
fredroy: true that

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/BaseMechanicalVisitor.h
fredroy: right, I forgot that.

SofaKernel/modules/SofaSimulationCore/src/sofa/simulation/Visitor.h
alxbilger: done

fredroy (approved):

PR #2099, opened 2021-05-19
[SofaGeneralSimpleFem] Remove *Containers

fredroy:
(not the topological ones 😏)

{Length,Radius,Stiffness,Poisson}Container was created a **long time ago** to provide an "external location" where to set data per element, regardless of the component.
But: 
 - the design was flawed (hard deps, not generic at all)
 - the remaining files are just interfaces (virtual ... =0) and there is no implementation at all in the whole codebase (deleted files in the past? forgotten plugins?), so the code itself at the moment is totally useless. Actually 3 of them was even not including the necessary headers...

So the best course of action to do IMO is to delete them, and maybe find an other design later to provide means to get data per element (opposed to data for the whole set of elements)

A nice side effect is that SofaTopologicalMapping just depends on SofaBaseTopology now.
______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2126, opened 2021-05-27
[SofaTopologyMapping] Fix Edge2Quad condition

fredroy:
Fix failing test because the user is giving a focal radius of zero.
This became an error in #2099 .

But it could happen that a user gives both a radius (non zero) and a focal radius (with zero).
This should not be really a fatal error (acting like the focal radius has not be given), 
but more a "warning" (a "focal" radius at 0 is still wrong after all)

### DIFF
--- a/modules/SofaTopologyMapping/src/SofaTopologyMapping/Edge2QuadTopologicalMapping.cpp
+++ b/modules/SofaTopologyMapping/src/SofaTopologyMapping/Edge2QuadTopologicalMapping.cpp
@@ -81,16 +81,16 @@ void Edge2QuadTopologicalMapping::init()
         msg_error() << "Radius is zero or negative";
         return;
     }
-    else if (d_radiusFocal.isSet() && d_radiusFocal.getValue() < std::numeric_limits<double>::min())
+    if (d_radiusFocal.isSet() && d_radiusFocal.getValue() < std::numeric_limits<double>::min())
     {
-        msg_error() << "Focal Radius is zero or negative";
-        return;
+        msg_warning() << "Focal Radius is zero or negative";
     }
+
     double rho = d_radius.getValue();
 
     bool ellipse = false;
     double rhoFocal;
-    if (d_radiusFocal.isSet())
+    if (d_radiusFocal.isSet() && d_radiusFocal.getValue() >= std::numeric_limits<double>::min())
     {
         ellipse = true;
         rhoFocal = d_radiusFocal.getValue();


guparan (approved):

fredroy:
[ci-build][with-scene-tests]

modules/SofaTopologyMapping/src/SofaTopologyMapping/Edge2QuadTopologicalMapping.cpp
```diff
@@ -81,16 +81,16 @@ void Edge2QuadTopologicalMapping::init()
...
-        return;
+        msg_warning() << "Focal Radius is zero or negative";
     }
+
     double rho = d_radius.getValue();
 
     bool ellipse = false;
     double rhoFocal;
-    if (d_radiusFocal.isSet())
+    if (d_radiusFocal.isSet() && d_radiusFocal.getValue() >= std::numeric_limits<double>::min())
```
epernod, line 93: should not be positive value?

modules/SofaTopologyMapping/src/SofaTopologyMapping/Edge2QuadTopologicalMapping.cpp
```diff
@@ -81,16 +81,16 @@ void Edge2QuadTopologicalMapping::init()
         msg_error() << "Radius is zero or negative";
         return;
     }
-    else if (d_radiusFocal.isSet() && d_radiusFocal.getValue() < std::numeric_limits<double>::min())
+    if (d_radiusFocal.isSet() && d_radiusFocal.getValue() < std::numeric_limits<double>::min())
     {
-        msg_error() << "Focal Radius is zero or negative";
-        return;
+        msg_warning() << "Focal Radius is zero or negative";
```
epernod, line 86: Not sure to understand what happened now if you put a negative value. It will create a mapping?

modules/SofaTopologyMapping/src/SofaTopologyMapping/Edge2QuadTopologicalMapping.cpp
alxbilger: `d_radiusFocal.getValue() >= std::numeric_limits<double>::min()` is equivalent to `d_radiusFocal.getValue() > 0`, isn't it?

epernod: Ok I thought it was -2^...  But from what I understand you are right: https://en.cppreference.com/w/cpp/types/numeric_limits
type	lowest()	min()		max()
double	-1.79769e+308	2.22507e-308	1.79769e+308

fredroy: @alxbilger I suppose so 😗

modules/SofaTopologyMapping/src/SofaTopologyMapping/Edge2QuadTopologicalMapping.cpp
fredroy: later on, it tests if the focal radius is positive then do something with it (otherwise it will be treated as the focal radius has not been given)

epernod (approved):

PR #2134, opened 2021-05-27
[SofaMiscCollision] Contact response order

### LINKED ISSUE #2128: BilateralInteractionConstraint.scn regression test fails
The regression test fails after #2076.
See the difference in the following picture:
image
The screenshot is taken after 1200 iterations.

#2076 deals with the group manager. It's probably the source of the difference.
In the scene, there is a static object (no ODE solver, and collision models define `simulated="false"`). In #2076, there is a fix on object without any ODE solver, but where collision models don't define `simulated="false"`. We are not in this case. The source of the difference is elsewhere.

alxbilger:
In #2076, a `vector` was replaced in favour of a `map`. It has a consequence on the order of traversal, therefore on the contact response creation. With a `map`, the order is based on the contact pointer address, which can be random. The vector is ordered by insertion.
This PR restores the `vector` for reproducible simulations. It should fix #2128.


fredroy:
[ci-build][with-all-tests]

fredroy (approved):

PR #2163, opened 2021-06-04
[SofaSimulationCore] Add tons of details in Euler solver

alxbilger:
- Lots of comments
- Split into multiple functions (`solve` function easier to read)

Next steps:

- Create a visitor to detect automatically if the mass matrix is trivially invertible (diagonal matrix), in order to replace `d_optimizedForDiagonalMatrix`.
- Benchmark with/without `SOFA_NO_VMULTIOP`
- More tests, in particular for the standard Euler method (not symplectic)
- Update the doc

### DIFF
--- a/SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
+++ b/SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
@@ -75,7 +75,7 @@ void EulerExplicitSolver::solve(const core::ExecParams* params,
 
     acc.realloc(&vop, !d_threadSafeVisitor.getValue(), true);
 
-    addSeparateGravity(&mop, dt, core::VecDerivId::velocity());
+    addSeparateGravity(&mop, dt, vResult);
     computeForce(&mop, f);
     projectResponse(&mop, acc);
 
--- a/SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
+++ b/SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
@@ -75,7 +75,7 @@ void EulerExplicitSolver::solve(const core::ExecParams* params,
 
     acc.realloc(&vop, !d_threadSafeVisitor.getValue(), true);
 
-    addSeparateGravity(&mop, dt);
+    addSeparateGravity(&mop, dt, core::VecDerivId::velocity());
     computeForce(&mop, f);
     projectResponse(&mop, acc);
 
@@ -244,14 +244,14 @@ void EulerExplicitSolver::init()
     reinit();
 }
 
-void EulerExplicitSolver::addSeparateGravity(sofa::simulation::common::MechanicalOperations* mop, SReal dt)
+void EulerExplicitSolver::addSeparateGravity(sofa::simulation::common::MechanicalOperations* mop, SReal dt, core::MultiVecDerivId v)
 {
     sofa::helper::ScopedAdvancedTimer timer("addSeparateGravity");
 
     /// Calls the "addGravityToV" method of every BaseMass objects found in the current
     /// context tree, if the BaseMass object has the m_separateGravity flag set to true.
     /// The method "addGravityToV" usually performs v += dt * g
-    mop->addSeparateGravity(dt);
+    mop->addSeparateGravity(dt, v);
 }
 
 void EulerExplicitSolver::computeForce(sofa::simulation::common::MechanicalOperations* mop, core::MultiVecDerivId f)
--- a/SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
+++ b/SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
@@ -77,7 +77,6 @@ void EulerExplicitSolver::solve(const core::ExecParams* params,
 
     addSeparateGravity(&mop, dt, vResult);
     computeForce(&mop, f);
-    projectResponse(&mop, acc);
 
     // Mass matrix is diagonal, solution can thus be found by computing acc = f/m
     if(d_optimizedForDiagonalMatrix.getValue())
@@ -89,6 +88,8 @@ void EulerExplicitSolver::solve(const core::ExecParams* params,
     }
     else
     {
+        projectResponse(&mop, f);
+
         core::behavior::MultiMatrix<simulation::common::MechanicalOperations> matrix(&mop);
 
         // Build the global matrix. In this solver, it is the global mass matrix
@@ -276,15 +277,14 @@ void EulerExplicitSolver::computeAcceleration(sofa::simulation::common::Mechanic
     mop->accFromF(acc, f);
 }
 
-void EulerExplicitSolver::projectResponse(sofa::simulation::common::MechanicalOperations* mop, core::MultiVecDerivId acc)
+void EulerExplicitSolver::projectResponse(sofa::simulation::common::MechanicalOperations* mop, core::MultiVecDerivId vecId)
 {
     sofa::helper::ScopedAdvancedTimer timer("projectResponse");
 
     // Calls the "projectResponse" method of every BaseProjectiveConstraintSet objects found in the
     // current context tree. An example of such constraint set is the FixedConstraint. In this case,
-    // it will set to 0 every row (i, _) of the right-hand side (force) vector for the ith degree of
-    // freedom.
-    mop->projectResponse(acc);
+    // it will set to 0 every row (i, _) of the input vector for the ith degree of freedom.
+    mop->projectResponse(vecId);
 }
 
 void EulerExplicitSolver::solveConstraints(sofa::simulation::common::MechanicalOperations* mop, core::MultiVecDerivId acc)
--- a/SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
+++ b/SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
@@ -24,7 +24,6 @@
 #include <sofa/simulation/MechanicalOperations.h>
 #include <sofa/simulation/VectorOperations.h>
 #include <sofa/core/ObjectFactory.h>
-#include <sofa/core/behavior/MultiVec.h>
 #include <sofa/core/behavior/MultiMatrix.h>
 #include <sofa/helper/AdvancedTimer.h>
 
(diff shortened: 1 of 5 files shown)


SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
```diff
@@ -53,127 +52,190 @@ EulerExplicitSolver::EulerExplicitSolver()
...
+        // 1. Calls the "vMultiOp" method of every mapped BaseMechanicalState objects found in the
+        // current context tree. This method may be called with different parameters than for the non-mapped
+        // BaseMechanicalState objects.
+        // 2. Calls the "vMultiOp" method of every BaseMechanicalState objects found in the
+        // current context tree.
+        vop->v_multiop(ops);
+
+        // Calls "solveConstraint" on every ConstraintSolver objects found in the current context tree.
+        mop->solveConstraint(newVel,core::ConstraintParams::VEL);
+        mop->solveConstraint(newPos,core::ConstraintParams::POS);
```
jnbrunet, line 207: I know you didn't introduce this and it was like that before, but, don't you think this "single-operation optimization" introduces a different behavior with respect to constraints? i.e.:

Without optimization:
```
v = v + a*t
solve_constraints(v)
x = x + v*t
solve_constraints(x)
```
With optimization:
```
v = v + a*t
x = x + v*t
solve_constraints(v)
solve_constraints(x)
```

SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
```diff
@@ -182,5 +244,83 @@ void EulerExplicitSolver::init()
     reinit();
 }
 
+void EulerExplicitSolver::addSeparateGravity(sofa::simulation::common::MechanicalOperations* mop, SReal dt)
```
jnbrunet: To follow the other method signatures and to be more explicit. Note that you will need to update calls to this function with
```c++
addSeparateGravity(mop, dt, sofa::core::VecDerivId::velocity());
```

```suggestion
void EulerExplicitSolver::addSeparateGravity(sofa::simulation::common::MechanicalOperations* mop, SReal dt, sofa::core::MultiVecDerivId v)
```

SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
```diff
@@ -182,5 +244,83 @@ void EulerExplicitSolver::init()
...
+{
+    sofa::helper::ScopedAdvancedTimer timer("addSeparateGravity");
+
+    /// Calls the "addGravityToV" method of every BaseMass objects found in the current
+    /// context tree, if the BaseMass object has the m_separateGravity flag set to true.
+    /// The method "addGravityToV" usually performs v += dt * g
+    mop->addSeparateGravity(dt);
```
jnbrunet:
```suggestion
    mop->addSeparateGravity(dt, v);
```

jnbrunet (approved):
Hey @alxbilger ,

Awesome PR, thanks a lot. Reading through this class will now be much clearer and approachable.

Small general notes (not necessarily for this PR):
1. We could probably rename the "**EulerSolver.h**" and "**EulerSolver.cpp**" files to "**EulerExplicitSolver.h**" and "**EulerExplicitSolver.cpp**" to follow the name of the class.
2. Maybe we could factor the sympletic option as a new time integrator class? I think it would be much clearer to have, for example, 
```xml
<EulerExplicitSolver />
<EulerImplicitSolver />
<EulerSemiImplicitSolver />
```
and might speak to more people.

SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
alxbilger: I noticed that too, but I don't know the answer. @ChristianDuriez can you help us on this question?

SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
alxbilger: I'll follow your suggestion. But I noticed that, by default without specifying any parameter, `addSeparateGravity` acts on `core::VecDerivId::velocity()`. Is it still valid for a free motion animation loop, where the new velocity corresponds to `core::VecDerivId::freeVelocity()`?

alxbilger:
>     1. We could probably rename the "**EulerSolver.h**" and "**EulerSolver.cpp**" files to "**EulerExplicitSolver.h**" and "**EulerExplicitSolver.cpp**" to follow the name of the class.
> 
>     2. Maybe we could factor the sympletic option as a new time integrator class? I think it would be much clearer to have, for example,
> 
> 
> ```
> <EulerExplicitSolver />
> <EulerImplicitSolver />
> <EulerSemiImplicitSolver />
> ```
> 
> and might speak to more people.

I definitively agree with both your suggestions.
I also thought about your second suggestion, and it was my plan to talk about it during the next dev meeting. The problem I see is that `EulerExplicitSolver` is by default symplectic. So, applying your suggestion would change the ODE solver when the user write `<EulerExplicitSolver />`. Let's discuss it on Wednesday.

alxbilger:
One thing I changed compared to the original code is:
from
```cpp
ops[op_pos].second.push_back(std::make_pair(newVel.id(),dt));
```
to
```cpp
ops_pos.second.emplace_back(d_symplectic.getValue() ? newVel.id() : vel.id(), dt);
```
Instead of providing `newVel` as the first argument, I check `d_symplectic`, and if it's `false`, I provide `vel`. I do that because I think the non-symplectic variant works on the velocity of the previous time step. Also, I think `newVel` is not yet computed for the non-symplectic variant. I would like your opinion @jnbrunet

SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
jnbrunet: This is exactly why I prefer not using defaults on parameters. 

Anyway, I think the velocity should be the one passed as parameter to the `EulerExplicitSolver::solve` method. Therefore, this would introduce a change of behavior when the `FreeMotionAnimationLoop` is used (before it was the `core::VecDerivId::velocity()`, now it will be `core::VecDerivId::freeVelocity()`). I think the later is the correct one (you want to apply gravity on the velocity without constraints first).

jnbrunet:
> > ```
> > 1. We could probably rename the "**EulerSolver.h**" and "**EulerSolver.cpp**" files to "**EulerExplicitSolver.h**" and "**EulerExplicitSolver.cpp**" to follow the name of the class.
> > 
> > 2. Maybe we could factor the sympletic option as a new time integrator class? I think it would be much clearer to have, for example,
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > ```
> > <EulerExplicitSolver />
> > <EulerImplicitSolver />
> > <EulerSemiImplicitSolver />
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > and might speak to more people.
> 
> I definitively agree with both your suggestions.
> I also thought about your second suggestion, and it was my plan to talk about it during the next dev meeting. The problem I see is that `EulerExplicitSolver` is by default symplectic. So, applying your suggestion would change the ODE solver when the user write `<EulerExplicitSolver />`. Let's discuss it on Wednesday.

Unfortunately, I'm not sure I will be able to join this Wednesday and the next.

Could we do something like this?


1. If `EulerExplicitSolver` is used with default parameters or with explicit `sympletic=True`, then warn the user that he should use `EulerSemiImplicitSolver` instead, and create this new component automatically for him (this is the part where I'm not sure it's feasible..)
2. Else, the user manually specified `sympletic=False`, warn him that this option no longer exists and that he can remove it from its scene file.

alxbilger:
> Could we do something like this?
> 
>     1. If `EulerExplicitSolver` is used with default parameters or with explicit `sympletic=True`, then warn the user that he should use `EulerSemiImplicitSolver` instead, and create this new component automatically for him (this is the part where I'm not sure it's feasible..)
> 
>     2. Else, the user manually specified `sympletic=False`, warn him that this option no longer exists and that he can remove it from its scene file.

In your first suggestion, it would not be possible to create `EulerExplicitSolver`.

SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
alxbilger: Short answer: it depends on how the constraint is implemented.

In the non-optimized version, constraints are applied on the velocity, and then the position is updated. Therefore, the new position takes into account the changes due to the constraints on the velocity. Finally, constraints on the position are applied.
In the optimized version, the position is updated first, and then the constraints on the velocity are applied. Therefore, the new position does not take into account the constraints on the velocity. However, it may have the same effect, if  the constraint applied on the position has the same effect than the constraints applied on the velocity.
To be completely rigorous, I think the non-optimized version is the good one. However, in practice the optimized version works similarly to the non-optimized version, because constraints on velocity and position are consistent.
Let's see if it worths keeping the "optimized" version after a benchmark.

SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
alxbilger: So do you agree with 5c0e8585fbc42bcbb39e131c311c7e9695a1b51d?

jnbrunet:
> One thing I changed compared to the original code is:
> from
> 
> ```c++
> ops[op_pos].second.push_back(std::make_pair(newVel.id(),dt));
> ```
> 
> to
> 
> ```c++
> ops_pos.second.emplace_back(d_symplectic.getValue() ? newVel.id() : vel.id(), dt);
> ```
> 
> Instead of providing `newVel` as the first argument, I check `d_symplectic`, and if it's `false`, I provide `vel`. I do that because I think the non-symplectic variant works on the velocity of the previous time step. Also, I think `newVel` is not yet computed for the non-symplectic variant. I would like your opinion @jnbrunet

Well, your question took me some time haha. I'm not sure about this... This is what I have at the moment when I try to derive them manually:

First option
```
Begin time step
Solve a in : (M + hC) a = P -  C v - R(x)
x = x + h v
v = v + h a
solve_constraints(x)
solve_constraints(v)
End time step
```

Second option
```
Begin time step
x = x + h v + h^2 a
v = v + h a
solve_constraints(x)
solve_constraints(v)
Solve a in : M a = P -  C v - R(x)
End time step
```

I've put my maths from the middle of the night here: 

https://www.overleaf.com/read/mdbcmmnqxzkk

But I might have made some mistakes... I don't get how SOFA doesn't have the same thing...

Edit: To clarify, I didn't really answer your question because I just can't reproduce neither of those equations from the ones found in the theory...

SofaKernel/modules/SofaExplicitOdeSolver/src/SofaExplicitOdeSolver/EulerSolver.cpp
jnbrunet: Yes I do !

alxbilger:
@jnbrunet If I understand correctly, your problem with this solver is not the approximation equations on which I asked your opinion, but with the way `a_{n+1}` is computed. From your document, `a_{n+1}` should be computed differently depending on if it symplectic or not. In EulerExplicitSolver, we just compute `a` in `Ma=f` for both symplectic and non-symplectic, right?

alxbilger:
[ci-build][with-all-tests]

jnbrunet:
Hey @alxbilger ,

I'm very sorry for the late reply.

> If I understand correctly, your problem with this solver is not the approximation equations on which I asked your opinion

Well kinda. I am unable to answer your initial question, since when I try to derive the equation to validate your changes, I'm unable to get to the same formulation as you have. But, in addition, I'm also unable to derive the old one (before your change)....

Anyway, if you can validate that the multiop "optimized" version gives the same result as the "unoptimized" one for both explicit and semi-implicit, and for both before and after this PR, you have my +1 to merge!

alxbilger:
@jnbrunet To be clear, there are 2 optimizations in this solver: 1) optimization for diagonal masses, 2) multiop optimization.
I did not touch anything in the non-optimized multiop code. I changed the velocity input in the position update in the optimized multiop code.

# Non-optimized multiop code

I did not change this code

## Symplectic

```
newVel = vec + acc * dt
newPos = pos + newVel * dt
```

## Non-Symplectic

```
newPos = pos + vel * dt
newVel = vel + acc * dt
```

# Optimized multiop code

```cpp
ops_vel.second.emplace_back(vel.id(), 1.0);
ops_vel.second.emplace_back(acc.id(), dt);
ops_pos.second.emplace_back(pos.id(), 1.0);
ops_pos.second.emplace_back(d_symplectic.getValue() ? newVel.id() : vel.id(), dt);
```
`ops_vel` is executed first if symplectic. `ops_pos` is executed first if non-symplectic.

If I translate this code to formulas:

## Symplectic

```
newVel = vel + acc * dt
newPos = pos + newVel * dt
```

## Non-symplectic

```
newPos = pos + vel * dt
newVel = vel + acc * dt
```

This is exactly the same formulas than for the non-optimized code!
My change was on the line:
`ops_pos.second.emplace_back(d_symplectic.getValue() ? newVel.id() : vel.id(), dt);`
Before it was:
`ops_pos.second.emplace_back(vel.id(), dt);`

It would translate to:

## Symplectic

```
newVel = vel + acc * dt
newPos = pos + vel * dt  <-- difference here with the non-optimized code
```

## Non-symplectic

```
newPos = pos + vel * dt
newVel = vel + acc * dt
```

# Conclusion

I believe that I fixed the multiop optimized symplectic version. Both optimized and non-optimized codes are now consistent. Note that this makes a difference only for the FreeMotion animation loops.

@jnbrunet Regarding your math problem, I don't understand it. I don't see where you see a difference with SOFA.

If you agree, I think we can merge this PR because it does not change the behavior (except for the fix). If there are some maths problems, it can probably be tackled later. What do you think?

jnbrunet:
Hey @alxbilger , yes go ahead, you can absolutely merge this. The issue I think I see here isn't related to this PR.

jnbrunet (approved):

PR #2175, opened 2021-06-11
[CMake] FIX SofaFramework aliases

guparan:
Imported targets can be aliased only if they are global.

Otherwise, this error is raised:
```
CMake Error at /data/Softwares/sofa/build/build-master/install/lib/cmake/SofaFramework/SofaFrameworkConfig.cmake:45 (add_library):
  add_library cannot create ALIAS target "SofaHelper" because target
  "Sofa.Helper" is imported but not globally visible.
Call Stack (most recent call first):
  CMakeLists.txt:5 (find_package)
```

Thus, this PR adds the GLOBAL property to the imported targets we want to alias.
It also move all the alias definitions in their respective projects (not in SofaFramework anymore).

### DIFF
--- a/SofaKernel/SofaFramework/CMakeLists.txt
+++ b/SofaKernel/SofaFramework/CMakeLists.txt
@@ -54,10 +54,3 @@ sofa_create_package_with_targets(
     INCLUDE_SOURCE_DIR "src"
     INCLUDE_INSTALL_DIR "${PROJECT_NAME}"
     )
-
-# Create aliases to support compatibility, starting from v21.06
-# These aliases will be deleted in v21.12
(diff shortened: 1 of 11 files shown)


fredroy:
Did you try with an out-of-tree compilation of a plugin (e.g Python3) ? just in case

guparan:
> 
> 
> Did you try with an out-of-tree compilation of a plugin (e.g Python3) ? just in case

Yes but with a simpler plugin (MyAwesomeComponents). `find_package(SofaFramework)` passed 👌

fredroy:
Okay then lets go for it

fredroy (approved):

PR #2177, opened 2021-06-14
[MultiThreading] Parallel hexa fem

alxbilger:
Parallel implementation of `HexahedronFEMForceField`.

For details, see the comments of the class `ParallelHexahedronFEMForceField`, in `ParallelHexahedronFEMForceField.h`.

# Results

On a cantilever beam (`ParallelHexahedronFEMForceField.scn`) with 1911 hexahedra, after 1000 time steps:

## Single threaded
```
21.1615 s ( 47.2557 FPS)
   3       0.01    1      19.42   28.17   20.23    0.83   20.23   95.51 ...Mechanical
   4       0.02    1       0.86    1.54    0.90    0.06    0.90    4.23 ....ComputeForce
   4       0.91    1       0.70    1.38    0.74    0.05    0.74    3.51 ....ComputeRHTerm
   4       1.66    1       0       0.03    0       0       0       0.02 ....MBKBuild
   5       1.66    1       0       0.02    0       0       0       0.01 .....CG-setSystemMBKMatrix
   4       1.66    1      17.79   26.51   18.56    0.78   18.56   87.62 ....MBKSolve
   5       1.68    1      17.76   26.48   18.53    0.77   18.53   87.49 .....CG-Solve
   4      20.22    1       0.01    0.04    0.01    0       0.01    0.05 ....UpdateVAndX

```
## Multi threaded
```
8.40483 s ( 118.979 FPS)
   3       0.01    1       5.32   26.82    7.42    1.31    7.42   88.31 ...Mechanical
   4       0.02    1       0.22    0.88    0.32    0.07    0.32    3.80 ....ComputeForce
   4       0.34    1       0.19    0.70    0.28    0.06    0.28    3.29 ....ComputeRHTerm
   4       0.61    1       0       0.07    0.01    0       0.01    0.06 ....MBKBuild
   5       0.61    1       0       0.07    0       0       0       0.03 .....CG-setSystemMBKMatrix
   4       0.62    1       4.87   25.85    6.80    1.22    6.80   80.84 ....MBKSolve
   5       0.64    1       4.84   25.82    6.77    1.22    6.77   80.51 .....CG-Solve
   4       7.42    1       0.01    0.02    0.01    0       0.01    0.13 ....UpdateVAndX

```

# Comments

- Code is in plugin MultiThreading because of the large amount of boilerplate code for the multi threading management, that would pollute the initial implementation in `HexahedronFemForceField`. I have some ideas to reduce this code, but it requires work on the task scheduler.
- A small part of the code is duplicated from `HexahedronFemForceField`.
- Initially, I wanted to accelerate the execution of the Conjugate Gradient for `GraphScattered` types. It is not possible to propose a generic multi-threaded acceleration for all force fields: for FEM, the most time consuming step in the CG is `ForceField::addDForce`. This is the method to accelerate. The downside is that it has to be done for every force field...


alxbilger:
[ci-build][with-all-tests]

applications/plugins/MultiThreading/src/MultiThreading/ParallelHexahedronFEMForceField.h
```diff
@@ -0,0 +1,187 @@
...
+    typedef typename DataTypes::VecCoord VecCoord;
+    typedef typename DataTypes::VecDeriv VecDeriv;
+    typedef core::objectmodel::Data<VecCoord> DataVecCoord;
+    typedef core::objectmodel::Data<VecDeriv> DataVecDeriv;
+    typedef typename Coord::value_type Real;
+    typedef helper::ReadAccessor< Data< VecCoord > > RDataRefVecCoord;
+    typedef helper::WriteAccessor< Data< VecDeriv > > WDataRefVecDeriv;
+    typedef core::topology::BaseMeshTopology::Hexa Element;
+    typedef core::topology::BaseMeshTopology::SeqHexahedra VecElement;
+    typedef defaulttype::Mat<24, 24, Real> ElementStiffness;
```
fredroy: Please use the new namespace/header for the types (Vec, Mat, etc)

applications/plugins/MultiThreading/src/MultiThreading/ParallelHexahedronFEMForceField.inl
```diff
@@ -0,0 +1,356 @@
...
+template<class DataTypes>
+sofa::simulation::Task::MemoryAlloc AddDForceTask<DataTypes>::run()
+{
+    auto elementId = m_startingElementId;
+    auto it = m_first;
+
+    defaulttype::Vec<24, Real> X; //displacement
+    defaulttype::Vec<24, Real> F; //force
+
+    const auto& elementStiffnesses = m_ff->_elementStiffnesses.getValue();
```
alxbilger: `getValue` is not thread-safe

jnbrunet:
Thanks @alxbilger !

Why not parallelize the matrix assembly (addKToMatrix)? Especially for a hexahedral forcefield since it can be quite expensive. It would than be very interesting to compare the multithreaded CG solver (thanks to your PR) performence against a multithreaded direct solver such as Pardiso LLT solver.

alxbilger:
> Thanks @alxbilger !
> 
> Why not parallelize the matrix assembly (addKToMatrix)? Especially for a hexahedral forcefield since it can be quite expensive. It would than be very interesting to compare the multithreaded CG solver (thanks to your PR) performence against a multithreaded direct solver such as Pardiso LLT solver.

I did not measure the performances, but I thought `addKToMatrix` was quick. Most of the time being taken by the linear solver itself. But I may be wrong... I'll have a look.

alxbilger:
After some tests on a cantilever beam with a LDL solver and a scalar CSR matrix:

- `addKToMatrix` takes a large part of the time for low number of elements
- `addKToMatrix` takes less and less part of the time as the number of elements grows
- The time consuming operation (more than 90% of the time) in `addKToMatrix` is `r.matrix->add(i, j, v)`. In the test, `r.matrix` is of type `CompressedRowSparseMatrix<double>`, and I did not test a bloc-based matrix. 
- In `r.matrix->add`, `wbloc` is the time consuming function. My profiler did not tell me more information inside this function. I tried to pre-allocate `btemp` before all the `add` calls, but it does not help. I suspect the bad performances come from the algorithm itself which must find an element in sorted arrays at each `add` call.
- Since the bad performances come from `CompressedRowSparseMatrix<double>::add`, it would be interesting to parallelize the calls to `add`. However, I doubt `CompressedRowSparseMatrix` is thread-safe...
- My tests are in https://github.com/alxbilger/sofa/commit/41584bceaf5d5666c362cdfb28238318c750af9e. But I don't plan to merge it since it does not improve the performances

jnbrunet:
Hey @alxbilger ,

I just realize, this is because it is a linear force field. Hence the stiffness matrix is constant and doesn't have to be recomputed  at each Newton steps (time steps in this case). You are therefore probably only computing the time taken for the copy/addition from the internal stiffness buffer (pre-computed) + the rotation extraction per elements into the global matrix, which isn't too costly.

You are right, no need for parallel version of a matrix to matrix addition :-)

alxbilger:
[ci-build][with-all-tests]

PR #2185, opened 2021-06-16
[SofaGeneralSimpleFem] add vonMises method in TetrahedralCorotationalFEMForceField

Mrbunbun:
VonMisesStress_TetrahedralCorotationalFEMForceField


epernod:
[ci-build][with-all-tests]

fredroy:
Not that I want to be that guy, but the code is quite copy-pasted from TetrahedronFEMForcefield (even the data name starts with '_' ) 🧐 . Wont it be possible to factorize the code with TetrahedronFEMForcefield  or anything like that ?

epernod:
> 
> 
> Not that I want to that guy, but the code is quite copy-pasted from TetrahedronFEMForcefield (even the data name starts with '_' ) 🧐 . Wont it be possible to factorize the code with TetrahedronFEMForcefield or anything like that ?

of curse the answer is "Yes, but..." 
Like that is it not that simple because they have a totally different way to store the data. I see 2 methods that could be factorized.
- the computeVonMises which could take the data as parameter, but I don't know where it should be placed.
- the draw method, but it is too specific to be put in the drawTools.

is there a ForceField toolkit or helper somewhere?

alxbilger (changes requested):
I agree with @fredroy. Try to factorize what can be factorize. @fredroy can you give a hint on where to put the factorize code?
Also, follow the code style guidelines: prefix the Data with `d_`

fredroy:
For now, I would say we could introduce a file into the SofaSimpleFEM module (the first module introducing the FEM in SOFA), with  a name like FEMUtils.h (I am not good with names...). Inside you can put the functions (or a class with static functions) factorizing the code (I would say something like computeVonMisesFromElement() and drawVonMisesElements())

epernod:
> 
> 
> For now, I would say we could introduce a file into the SofaSimpleFEM module (the first module introducing the FEM in SOFA), with a name like FEMUtils.h (I am not good with names...). Inside you can put the functions (or a class with static functions) factorizing the code (I would say something like computeVonMisesFromElement() and drawVonMisesElements())

In fact TetrahedronFEMForceField and TetrahedralCorotationalFEMForceField should be merged. Historically the Tetrahedral is only a copy past with topological changes support.  Now that topological changes are done using callbacks, the merge could be done by simply adding an option that will or won't registered the call back.

PR #2187, opened 2021-06-17
[Sofa.Type] Move LCPsolver class into a utility function

fredroy:
LCPSolver is a class performing LCP on a matrix (actually on a raw pointer, assuming you did allocate the memory beforehand), but the fact this is a class is irrelevant.
This is only used in proximity.cpp (in the repo), which used for some proximity detections.

In the same way as other Matrix algo, I set this feature as a utility function in Mat_solve_LCP in Sofa.Type.
 - use Sofa.Type types instead of pointer (less prone to error)
 - rewrote some preproc def into constexpr values.
 - removed the print() function of the result, but you can just print the resulting type::matrix anyway

There are some comments in french but I did not bother translate them for now.

There is a compat layer for this function in Sofa.Compat so this PR should not be breaking.


fredroy:
[ci-build]

PR #2209, opened 2021-06-28
[SofaImplicitOdeSolver] Add a Euler implicit solver with Newton iterations

jjcasmar:
This is a port of the a Newton based Euler implicit solver I had in another project. In that project the solver return a ConvergenceStatus, which in the end allowed me to reduce the dt if the step failed to covnerged. In this case its not possible without modifying the AnimationLoop, but in any case, I am restoring the original position and velocities (or trying to, because it's not actually happening).

### DIFF
--- a/SofaKernel/modules/SofaCore/src/sofa/core/MechanicalParams.h
+++ b/SofaKernel/modules/SofaCore/src/sofa/core/MechanicalParams.h
@@ -238,7 +238,8 @@ public:
 
     MechanicalParams* operator= ( const MechanicalParams& mparams );
 
-
+    bool linearizeForces() const;
+    void setLinearizeForces(bool newLinearizeForces);
 protected:
 
     /// Time step
@@ -277,6 +278,9 @@ protected:
     /// True if a symmetric matrix is assumed in the left-hand term of the dynamics equations, for solvers specialized on symmetric matrices
     bool m_symmetricMatrix;
 
+    /// True if the call to addForce must store information for a latter call to addDForce
+    bool m_linearizeForces;
+
     /// @name Experimental compliance API
     /// @{
 protected:
@@ -289,9 +293,6 @@ public:
     void setImplicitPosition( SReal i ) { m_implicitPosition = i; }
     const SReal& implicitPosition() const { return m_implicitPosition; }
     /// @}
-
-
-
 
 };
 
--- a/SofaKernel/modules/SofaImplicitOdeSolver/src/SofaImplicitOdeSolver/NBESolver.cpp
+++ b/SofaKernel/modules/SofaImplicitOdeSolver/src/SofaImplicitOdeSolver/NBESolver.cpp
@@ -1,0 +1,205 @@
+#include "NBESolver.h"
+
+#include <sofa/core/ObjectFactory.h>
+#include <sofa/core/visual/VisualParams.h>
+#include <sofa/helper/AdvancedTimer.h>
+#include <sofa/simulation/MechanicalOperations.h>
+#include <sofa/simulation/VectorOperations.h>
+
+namespace sofa::component::odesolver {
+using sofa::core::VecId;
+using namespace sofa::defaulttype;
+using namespace sofa::core::behavior;
+
+NBESolver::NBESolver()
+    : sofa::core::behavior::OdeSolver(),
+      m_rayleighStiffness(0),
+      m_rayleighMass(0),
+      m_newtonThreshold(1e-5),
+      m_maxNewtonIterations(1),
+      m_lineSearchMaxIterations(1) {}
+
+void NBESolver::init() { sofa::core::behavior::OdeSolver::init(); }
+
+void NBESolver::cleanup() {}
+
+void NBESolver::solve(const sofa::core::ExecParams *params, SReal dt,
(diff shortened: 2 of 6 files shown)


SofaKernel/modules/SofaImplicitOdeSolver/src/SofaImplicitOdeSolver/NBESolver.h
```diff
@@ -0,0 +1,43 @@
+#ifndef VNCS_NBESOLVER_H
+#define VNCS_NBESOLVER_H
+
+#include <sofa/core/behavior/OdeSolver.h>
+
+namespace sofa::component::odesolver {
+class NBESolver : public sofa::core::behavior::OdeSolver {
```
alxbilger, line 7: Could you add a brief description of this class, what it does, what NBE stands for etc?

SofaKernel/modules/SofaImplicitOdeSolver/src/SofaImplicitOdeSolver/NBESolver.h
```diff
@@ -0,0 +1,43 @@
...
+ public:
+  SOFA_CLASS(NBESolver, sofa::core::behavior::OdeSolver);
+  NBESolver();
```
alxbilger, line 10: I think constructors are protected

SofaKernel/modules/SofaImplicitOdeSolver/src/SofaImplicitOdeSolver/NBESolver.cpp
```diff
@@ -0,0 +1,205 @@
+#include "NBESolver.h"
```
alxbilger, line 1: you need to include a header before anything in the file (see existing files)

SofaKernel/modules/SofaImplicitOdeSolver/src/SofaImplicitOdeSolver/NBESolver.h
```diff
@@ -0,0 +1,43 @@
+#ifndef VNCS_NBESOLVER_H
+#define VNCS_NBESOLVER_H
```
alxbilger, line 2:
```suggestion
#pragma once
```
and remove the `#endif` at the end of the file

guparan:
@jjcasmar Do you still plan to update NBESolver? You said on Gitter it was not up-to-date.

jjcasmar:
Yes, but I am doing some testing to check everything is worked as expected. I have some scenes using this NBESolver + a custom CG and they are failing, need to check if its the NBESolver or the CG.

fredroy:
Any news @jjcasmar ? 🧐

jjcasmar:
Sorry, I haven't done anything about this yet for SOFA. I am currently using the solver in my own project and right now I am testing it works correctly. I have some cases where the energy of the system seems to grow and I am not sure if its because of the solver or I have issues somewhere else, so I would say its not ready yet.

jjcasmar:
I have advance a bit on this and I have a robus implementation. But there is an issue regarding rayleigh damping. If we consider implicit damping forces, we would have something like f = - (a*M + b*K)*v_new. That M and K are constants (for example, they can be the M and K of the current point (M0 and K0 for instance)). They must remain constant, or we should take their derivatives. 

That means that during the linear solve, we have to compute M and K of the current linealization point and M0 and K0 of the point where we compute the damping forces. That is not efficient at all. For example, with a CG solver, we would need to change the state of the mechanical objects for each CG iteration

fredroy:
List of things to do:
 - fix conflicts
 - expose the private members as Data<> (for getters/setters, py2, xml etc)
 - register the component in the Factory (now it cannot be spawn like other BaseObject)
 - get an example, tests
 - some math people could review (@hugtalbot, @jnbrunet ) ?

SofaKernel/modules/SofaCore/src/sofa/core/MechanicalParams.h
```diff
@@ -277,6 +278,9 @@ class SOFA_CORE_API MechanicalParams : public sofa::core::ExecParams
     /// True if a symmetric matrix is assumed in the left-hand term of the dynamics equations, for solvers specialized on symmetric matrices
     bool m_symmetricMatrix;
 
+    /// True if the call to addForce must store information for a latter call to addDForce
+    bool m_linearizeForces;
```
alxbilger, line 282: I think it would be useful to have an example of a use of this bool. Is it required to use this variable in all force fields?

alxbilger (changes requested):

SofaKernel/modules/SofaCore/src/sofa/core/MechanicalParams.h
jjcasmar: Ideally yes. Rationale is that you don't want to linealize your forces each time you call addForce (or store linealization point info). But if you don't use it, then nothing bad happens, but you may be losing performance.

Actually, would be more useful to set the proper value of this variable from the solvers, so ForceFields can react to it. To be safe, it is initialize to true, so current behavior is preserved.

alxbilger:
I close because there is now a generic Newton solver

PR #2211, opened 2021-06-29
[SofaCore] Link: case where Data parent is invalid

alxbilger:
`parentData == nullptr` leads to a crash. I think it happens when the linked object could not be created for some reasons.

### DIFF
--- a/SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
+++ b/SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
@@ -484,7 +484,7 @@ bool Base::parseField( const std::string& attribute, const std::string& value)
                 }
                 else
                 {
-                    msg_warning() << "Link Data '" << attribute << "' to invalid Data '" << value << '\'';
+                    msg_error() << "Link Data '" << attribute << "' to invalid Data '" << value << '\'';
                     ok = false;
                     continue;
                 }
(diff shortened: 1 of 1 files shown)


alxbilger:
@damienmarchal I would like to know your opinion about what to do in the case `parentData==nullptr`, because I am not sure about what I did.

SofaKernel/modules/SofaCore/src/sofa/core/objectmodel/Base.cpp
```diff
@@ -475,8 +475,19 @@ bool Base::parseField( const std::string& attribute, const std::string& value)
...
             else
             {
-                BaseData* parentData = dataVec[d]->getParent();
-                msg_info() << "Link from parent Data " << value << " (" << parentData->getValueTypeInfo()->name() << ") to Data " << attribute << "(" << dataVec[d]->getValueTypeInfo()->name() << ") OK";
+                if (BaseData* parentData = dataVec[d]->getParent())
+                {
+                    msg_info() << "Link from parent Data " << value
+                                                         << " (" << parentData->getValueTypeInfo()->name()
+                                                         << ") to Data " << attribute
+                                                         << " (" << dataVec[d]->getValueTypeInfo()->name() << ") OK";
```
guparan:
```suggestion
                    msg_info() << "Link from parent Data "
                                    << value << " (" << parentData->getValueTypeInfo()->name() << ") "
                                    << "to Data "
                                    << attribute << " (" << dataVec[d]->getValueTypeInfo()->name() << ") "
                                    << "OK";
```

alxbilger:
I tested different scenarios:
If `ComponentA` is defined before `ComponentB`:

1. A Data of `ComponentA` referring to a Data of `ComponentB` is not linked properly: the numeric value is not the right one, and Data that depends on the first Data are not updated accordingly.
2. A Data of `ComponentB` referring to a Data that does not exist (and will not exist) does not warn the user. The link is created (pointing to nothing) with a default numeric value.

Therefore, I suggest to restore the `else` statement with a warning to the user that the link does not have a parent. It seems that the case when a Data points to a Data that will be initialized later is not supported.
What do you think @damienmarchal ?

PR #2220, opened 2021-07-01
[SofaBaseLinearSolver] Extract and uncomment CRSMultiMatrixAccessor

alxbilger:
The intention of `CRSMultiMatrixAccessor` was to use it as a replacement of `DefaultMultiMatrixAccessor` in `MatrixLinearSolver`. The switch is done at compilation time using a macro.
However, `CRSMultiMatrixAccessor` compilation was also controlled by the same macro which is not switched everyday. In other words, `CRSMultiMatrixAccessor` was almost dead code.

I extracted the `CRSMultiMatrixAccessor` class from the `DefaultMultiMatrixAccessor` file to put it in its own file.
I removed the condition on the compilation of `CRSMultiMatrixAccessor` so that this code is always compiled.

Currently, I don't know if `CRSMultiMatrixAccessor` is useful. It may be removed in the future.


alxbilger:
[ci-build][with-all-tests]

guparan (approved):
Thanks for the cleaning. Looks indeed like dead code to me. Added in 2011, hidden behind a macro, never touched since. Should be in a plugin I guess.

guparan:
Can't find a clean way to have both `[[deprecated]]` and `__attribute__((visibility ("default")))` on a class.
To be done in another PR.

PR #2235, opened 2021-07-08
[SofaCore] Fix RotationFinder return type for child classes

pedroperrusi:
Return type of `getRotations()` should use the `typedef` type to respect the floating point precision of the template `DataTypes`

As it is now, it causes compilation errors when  using template `defaulttype::Vec3fTypes`


fredroy (approved):
I guess it came from the uber PR setting the type:: namespace.

PR #2244, opened 2021-07-09
[SofaCore] Remove unused addSubMBKToMatrix in force fields

alxbilger:
I saw that `addSubMBKToMatrix` was not used in the SOFA code base, so I removed it savagely. I don't know the consequence of it. I would like feedback.

Of course, if it is decided to remove it, I'll add the deprecated flags in this PR.


fredroy:
It seems it was introduced by @courtecuisse long time ago: https://github.com/sofa-framework/sofa/commit/1340fd20218a601ed385d5c9c4d5bcfe8ec42ced
And indeed, the visitor calling this function seems not used anywhere in the (public) repository.
But maybe some (external) plugins may use it. 
@courtecuisse (or @pedroperrusi ) do you still need it ?

pedroperrusi:
I see no usage of addSubMBKToMatrix in our team's plugins, so it should be ok to remove it

pedroperrusi:
After discussion with @courtecuisse, the function is supposed to be used when cutting meshes.
In our plugins it is not used anymore, although it might still have relevance in the future.
Whether to keep it in SOFA sources or not... any option is ok for us.

pedroperrusi (approved):

alxbilger:
> After discussion with @courtecuisse, the function is supposed to be used when cutting meshes.
> In our plugins it is not used anymore, although it might still have relevance in the future.
> Whether to keep it in SOFA sources or not... any option is ok for us.

@epernod Would this function be useful for topological changes?

fredroy:
[ci-build][with-all-tests]

fredroy:
Waiting for the CI but ready in principle

courtecuisse:
Dear @alxbilger dear @epernod 
I'm wondering why did you "savagely" removed our work, and I'm writing to know what are the alternatives to this function? If there is no other solution to perform this critical operation, I'd like to know how many messages should we send to justify that a code is needed and should not be "savagely" removed?

To answer your question @epernod, yes this function was indeed written for topological changes, but on the mechanical side (not geometrical). Indeed, independently of the method that you use for topological modifications, you will need to recompute your mechanical matrices. Since you don't know which forcefield is used in your scene you cannot put that directly in a plugin. I guess that without this function you have to call reinit of your forcefield (assuming that the function is correctly implemented). However, there are many cases where these topological modifications only change some elements and not the entire mesh (there are many papers that describe this point on the web if you are interested...). In this case, the reinit method will be very expensive for nothing! I'm sure you are aware that sofa is designed for real-time, so we have to provide a solution to only compute the elements that are modified. 

In conclusion, if you combine the fact that you don't know the underlying forcefield and you want to recompute only subpart of Ke elements, you may think that it would be a great idea to have a method in BaseForcefield where you can specify which Ke you want to compute. This is precisely the goal of this function !

The question now is what is the alternative solution? I'm sure that the consortium does not "savagely" remove the work of other people without thinking of alternative solutions, especially if the same person has already discussed the need for this method in this PR :)

Waiting for answers, I have a student that is stuck on this problem for one week on this point. 

Hadrien.

alxbilger:
@courtecuisse 

I can't believe my eyes when reading your message. It is highly inappropriate, childish, unprofessional and overreacted.

Let's restart the chronology of this PR:
1) A suggestion has been made to remove some unused functions. The suggestion is made using the PR system.
2) Reviews have been requested from you and @pedroperrusi 
3) @pedroperrusi did not find any harm to remove the function, while you did not review the PR, nor answered to comments.
4) Therefore it has been decided to approve the PR. Nevertheless, the depreciation has been forgotten.

> I'd like to know how many messages should we send to justify that a code is needed and should not be "savagely" removed?

Only one! But I cannot see any.

> especially if the same person has already discussed the need for this method in this PR :)

Still no sign from you in this PR


If the removed function is really critical, a new PR can re-introduce it, associated with an example that the CI can execute.

Please, choose your words more carefully.

courtecuisse:
Hi guys,

Ok we'll have a look at the callback version to see if this operation (that I still believe may be of interest for several reasons i.e. cutting, domain decomposition,... ) can be replaced by the new callback solution. I have to admit that I don't know this new API. If it can be replaced easily, we'll modify our code to use it, if not we'll be forced to push our development a bit more on the side of sofa :(

PR #2259, opened 2021-07-20
[SofaHelper] Factory key type can be other than std::string

alxbilger:
Factories were used only with `std::string` as a key. That is why it was not noticeable that factories were not compatible with other types than `std::string` as the key.
The limitations come from:

- `logFactoryRegister` assumed `Key` is a string. It is now a template function.
- `Creator` inherits from `Key` and instances was used such as it was a string. Now, `Creator` no longer inherits from `Key` but just stores the key as a class member.

A unit test is added for enum type as a key.

Small downside: `std::string& getFactoryLog()` is now available outside the translation unit.

### DIFF
--- a/SofaKernel/modules/SofaCore/src/sofa/core/collision/Contact.h
+++ b/SofaKernel/modules/SofaCore/src/sofa/core/collision/Contact.h
@@ -83,10 +83,10 @@ public:
     virtual void setKeepAlive(bool /* val */) {}
 
     //Todo adding TPtr parameter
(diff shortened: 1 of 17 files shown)


alxbilger:
[ci-build][with-all-tests]

fredroy:
Lots of weird stuff happening with those *_API (dllimport/dllexport) thingies 🥴
TIL I learnt that you cannot put it after the type if it is a reference (I suppose the `&` bothers the compilation)
Seems better to put it always before the type.

fredroy (approved):

PR #2264, opened 2021-07-22
[SofaHelper] Remove boost::thread dependency

fredroy:
Follow-up on the quest to remove boost (at least the parts where the libraries are needed)

### 1
The only remaining part of SOFA using boost::thread is Sofa.Helper/Watchdog whose only purpose is to kill the current process if it exceeds a given time.
It is only used in the deprecated sofaInitTimer project, whose is litteraly a "temporary" project and was done to monitor the init() of the scenes... a kind of SofaVerification. Even the [commit ](https://github.com/sofa-framework/sofa/commit/0cbb991f6a7d6d1b1f32489d82c3267a6adc36a3) is saying "ADD: temporary project to profile the initialization of some scenes"

Anyway, I decided to not delete the file by pity, and just moved the Watchdog class into this project.
Frankly, this class and the sofaInitTimer could be deleted in my opinion

### 2
Sensable plugin is using optionnally `boost::thread::yield`, which:
 1. is deprecated
 2. can be replaced with std::this_thread::yield()

### 3
CircularQueue had its tests disabled since... forever. I saw that because it was using boost::thread.
I made these tests compiled again but 2 are failing. I made them disabled, if somebody wants to fix them...

### 4
Sofa.Helper does not need boost::thread anymore and thus is removed from the CMake files.


alxbilger (approved):

fredroy:
[ci-build][with-all-tests]

guparan (approved):

alxbilger (approved):

PR #2267, opened 2021-07-26
[SofaGraphComponent] Restore check of deprecated components 

alxbilger:
Introduction of a scene checker SceneCheckDeprecatedComponents. Its job is to check for deprecated components in the scene, which was done previously by SceneCheckAPIChange. But SceneCheckAPIChange is no longer enabled.

Note that SceneCheckAPIChange requires an APIVersion component in the scene, but does not use it to check for deprecated components. That is why I decided to move this check elsewhere.


alxbilger:
[ci-build][with-all-tests]

PR #2275, opened 2021-07-28
[SofaMiscFEM] Fix TriangleFEMForceField and TriangularFEMForceField to have the same behavior

epernod:
Several fixes:
- TriangleFEMForcefield : wrong init value for strainDisplacement
- TriangularFEMForceField: strainDisplacement not updated at each timestep and using rest Triangle area like in TriangleFEMForceField
- Enable the 2 tests which same value references for TriangleFEMForcefield and TriangularFEMForceField:
checkTriangularFEMForceField_init
checkTriangularFEMForceField_values

### DIFF
--- a/modules/SofaMiscFem/SofaMiscFem_test/TriangleFEMForceField_test.cpp
+++ b/modules/SofaMiscFem/SofaMiscFem_test/TriangleFEMForceField_test.cpp
@@ -731,12 +731,12 @@ TEST_F(TriangleFEMForceField3_test, checkTriangularFEMForceField_wrongAttributes
     this->checkWrongAttributes(1);
 }
 
-TEST_F(TriangleFEMForceField3_test, DISABLED_checkTriangularFEMForceField_init)
+TEST_F(TriangleFEMForceField3_test, checkTriangularFEMForceField_init)
 {
     this->checkInit(1);
 }
 
-TEST_F(TriangleFEMForceField3_test, DISABLED_checkTriangularFEMForceField_values)
+TEST_F(TriangleFEMForceField3_test, checkTriangularFEMForceField_values)
 {
     this->checkFEMValues(1);
 }
--- a/modules/SofaMiscFem/SofaMiscFem_test/TriangleFEMForceField_test.cpp
+++ b/modules/SofaMiscFem/SofaMiscFem_test/TriangleFEMForceField_test.cpp
@@ -411,7 +411,7 @@ public:
             {
                 typename TriangularFEM::TriangleInformation triangleInfo = triFEM->triangleInfo.getValue()[id];
                 const type::fixed_array <Coord, 3>& rotatedInitPos = triangleInfo.rotatedInitialElements;
-                const Mat33& rotMat = triangleInfo.initialTransformation;
+                const Mat33& rotMat = triangleInfo.rotation;
                 const Mat33& stiffnessMat = triangleInfo.materialMatrix;
                 const Mat63& strainDispl = triangleInfo.strainDisplacementMatrix;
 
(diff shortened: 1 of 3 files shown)


epernod:
[ci-build][with-all-tests]

epernod:
just for the record: it seems that test DISABLED_ are counted in the total number of tests displayed the dashboard.

fredroy:
[ci-build]

fredroy:
[ci-build][with-all-tests]

epernod:
I'm sorry mr @epernod . There seems to be a numerical issue in not reapplying the updated area (from deformed position).

epernod:
[ci-build][with-all-tests]

epernod:
There is a "big" divergence between ```TriangularFEMForceField``` and ```TriangleFEMForceField``` (which should have exactly the same behavior, the only difference is the c++ data structure to handle topological changes).

**In a short version**, when computing the Force, ```TriangularFEMForceField``` recompute the triangle area with the current positions whereas ```TriangleFEMForceField``` use the triangle area at rest position.


**In more details**:
In ```TriangleFEMForceField```:

- At init we compute the Material Stiffness Matrix **K** (function of young and poisson) and K[i] (i being the triangle number) is multiplied by the rest area of triangle i.
- At each timestep in computeForce, we compute:
 1. The Strain Displacement `J`  
 3. The Strain = Strain Displacement * Displacement: `JtD = Jt * Displacement`
 4. The Stress = Material Stiffness * Strain: `KJtD = K * JtD`   (here K has already the triangle area)
 5. Force : F = strain displacement * stress: `F = J * KJtD`


In ```TriangularFEMForceField``` the differences are:
- At init we compute the Material Stiffness Matrix **K** but no triangle area taken into account.
- At each time step:
 1. Each Triangle area is recomputed at the same time as the Strain Displacement
 2. same
 3. same
 4. Force: F = strain displacement * stress * triangleArea (recomputed in 1.)

According to the [paper](http://evasion.imag.fr/Publications/2005/NPF05/NPF05.pdf), the version of TriangleFEMForceField is the good one but I'm not expert enough in FEM to decide which solution is the good one.
@jnbrunet @Camille-K @TheGreatLudini @hugtalbot @ChristianDuriez @alxbilger  any idea?

Camille-K:
Indeed, in the paper they say that they use the initial volume of each element, instead of recomputing it. I'm not sure I understand it correctly, but it seems to me that this is not actually 'the good way', just something made possible by the fact that the volume can be directly multiplied to the Strain Displacement matrix, without creating artificial torque. So you can make the assumption that the volume of the element doesn't change much, and that using its initial volume during the whole deformation is a good approximation, which saves computational time.
It would be better if someone could comment on this, because I really don't know if I'm right.

That being said, when looking in the code of `TriangleFEMForceField`, the area of the triangles seems to be recomputed anyway in `computeStrainDisplacement`, with:
```
Real determinant = b[0] * c[1]; // Surface * 2
```
which is used to update the Strain Displacement matrix at each time step. By comparison, the difference with `computeStrainDisplacement` in `TriangularFEMForceField` is juste the next line
```
determinant = b[0] * c[1];
triangleInf[elementIndex].area = determinant*0.5f;
```
Following the paper, I think the computation of `determinant` at each time step should also be replaced by the triangle initial area. Or alternatively, follow the same version as in `TriangularFEMForceField` and update the triangle's area to make use of the computation.

Finally, but not related, the 'area' used in the material behaviour matrix of `TriangleFEMForceField` is actually a 'volume' with a default thickness of 1. I don't know if this feature is actually used, but if so it changes a bit the physical interpretation between `TriangleFEMForceField` and `TriangularFEMForceField`

jnbrunet:
It is probably using the rest area since it is working with a Lagrangian formulation. This is also what we use in other FEM code (tetra and hexa) where the integral is computed on the rest (undeformed) domain. For linear elements (triangle, tetra), it will multiply the integrand with the initial volume (or area for triangle) to get the integral result over an element. For trilinear or higher degree elements, it will use Gauss integration points where derivatives of the shape function, and the determinant of the jacobian of the transformation matrix are computed at the initial (undeformed) domain. Kinematics quantities from the current state are transformed into the initial domain using a transformation matrix. 

I would therefore think that the `TriangleFEMForceField ` is the "good" one in the sens that it is following the big majority of SOFA code with a Lagrangian formulation. 

You could always compare the result against another FE software such as FEBio and FEnIcs which are free. I can help you with that if you want.

epernod:
Thank you @Camille-K and @jnbrunet for your feedback and explanation. 

**As highlighted, we will keep the Lagrangian formulation and use the rest area to keep consistency with the others FEM implementation**. 

- Regarding the triangle Thickness, it is not used anymore and will be removed when TriangleFEMForceField and TriangularFEMForceField will be merged to keep the code as simple as possible.
- Comparing the results with FEBio or Fenics is a very good idea. Which one do you recommend ? Considering starting from 0 with not much time.

PR #2277, opened 2021-07-28
[SofaSimpleFem] Fix row/col indices in hexa fem for bloc-based matrices

alxbilger:
Indices were wrong in case of multiple objects under the same linear solver which uses an assembled 3x3 bloc-based matrix.
Bug introduced in https://github.com/sofa-framework/sofa/pull/2240


fredroy:

SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
```diff
@@ -1197,7 +1197,7 @@ void HexahedronFEMForceField<DataTypes>::addKToBlocMatrix(
                         Coord(Ke[3*n1+0][3*n2+0],Ke[3*n1+0][3*n2+1],Ke[3*n1+0][3*n2+2]),
                         Coord(Ke[3*n1+1][3*n2+0],Ke[3*n1+1][3*n2+1],Ke[3*n1+1][3*n2+2]),
                         Coord(Ke[3*n1+2][3*n2+0],Ke[3*n1+2][3*n2+1],Ke[3*n1+2][3*n2+2])) ) * Rot;
-                *crsmat->wbloc(offset + node1, offset + node2, true) += tmp * (-k);
+                *crsmat->wbloc(offset / 3 + node1, offset / 3 + node2, true) += tmp * (-k);
```
hugtalbot, line 1200: In the other cases of `addKToMatrix`, it appears that the same `offset` is used but it is never divided by 3:

`
r.matrix->add(r.offset + 3 * node1, r.offset + 3 * node2, tmp * (-kFactor));
`

I guess it comes from the difference between the `wbloc` and `add` functions, right?

alxbilger: Yes, they both accept indices, but they don't represent the same thing

hugtalbot (approved):

PR #2291, opened 2021-08-06
[Sofa.BaseTopology] Remove 1D template for 2D/3D primitives in topology algorithms

fredroy:
Because it does not make sense to me to use tetra algos in 1D for example 🤥

Maybe should we remove the 2D template for 3D primitives as well ?
I was considering tetra/hexa could be projected in 2D but it seems a bit far streched to me.


epernod:
[ci-build][with-all-tests]

fredroy:
A follow-up from the comment in the report (`PR is fine, usage of Vec1 should be checked with all teams`)
If a team really needs 1D for those, they can still do in their own code/plugin with their own explicit instanciation of {Hexa/Quand/Triangle/...}SetGeometryAlgorithms.

PR #2281, opened 2021-07-30
[SofaSimpleFem] Simplify bloc-based optimization

alxbilger:
PR based on #2280.
Template specialization for `void CompressedRowSparseMatrix<type::Mat<3,3,double> >::add(Index row, Index col, const type::Mat3x3d & _M)` in order to accelerate insertion.
This allows to avoid branching in force fields, based on the type of the system matrix (`dynamic_cast`). I removed it in `HexahedronFEMForceField`, but it could be removed in other places. It allows also to automatically optimize bloc insertion in force fields that did not have the branches.

# Benchmarks

## List of benchmarks

- `BM_CRS_Fixture<double>/Add3x3Bloc_CRSdouble`: insertion of 1000 3x3 blocs into a CRS made of double
- `BM_CRS_Fixture<sofa::type::Mat<3,3,double>>/Add3x3Bloc_CRS3x3d`: insertion of 1000 3x3 blocs into a CRS made of 3x3 blocs
- `BM_CRS_Fixture<sofa::type::Mat<3,3,double>>/Add3x3BlocShortcut_CRS3x3d`: insertion of 1000 3x3 blocs into a CRS made of 3x3 blocs, but insertion uses the fast function specialized for 3x3 CRS matrices. This is the fastest possible bloc insertion. It is actually used in the specialized function introduced by this PR, among other checks. Therefore, this speed is the goal to achieve for the specialized function.
- `BM_CRS_Fixture<double>/Add3x3BlocScalar_double`: insertion of 1000 3x3 blocs into a CRS made of double using 9 individual scalar insertion
- `BM_CRS_Fixture<sofa::type::Mat<3,3,double>>/Add3x3BlocScalar_CRS3x3d `: insertion of 1000 3x3 blocs into a CRS made of 3x3 blocs, using 9 individual scalar insertion. This is equivalent to what happens in BaseMatrix' bloc insertion, therefore it corresponds to the previous behavior of bloc insertion (before this PR).

## Before

```
-----------------------------------------------------------------------------------------------------------------
Benchmark                                                                       Time             CPU   Iterations
-----------------------------------------------------------------------------------------------------------------
BM_CRS_Fixture<double>/Add3x3Bloc_CRSdouble                                 75568 ns        75550 ns         9185
BM_CRS_Fixture<sofa::type::Mat<3,3,double>>/Add3x3Bloc_CRS3x3d              55533 ns        54699 ns        12798
[...]

PR #2306, opened 2021-08-25
[SofaBaseLinearSolver] CRS explicit instantiation

alxbilger:
Explicit instantiation of some common template for `CompressedRowSparseMatrix`.
This is to force the template specialization of some functions for specific templates (function `add` and `filterValues`). Before, template was specialized, but not in the right way. The specialization was not taken into account.
Also, `filterValues` has been factorized (1 function instead of 4 identical functions).

Now, `add` uses the correct specialization when it is used in a force field (badly introduced in #2281). It comes with the expected speed up.


alxbilger:
[ci-build][with-all-tests]

alxbilger:
[ci-build][with-all-tests]

fredroy:
Ready but waiting #2310

PR #2310, opened 2021-08-30
[SofaDenseSolver][SofaBaseLinearSolver] CLEAN macros *_CHECK and *_VERBOSE

guparan:
Should fix your infinite build issue @alxbilger :eyes: 

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2311, opened 2021-08-30
[SofaDenseSolver] Introduce NewMat unit tests

alxbilger:


fredroy:
Any interest to keep newmat ? I remember a discussion during a sofadevmeeting with @jnbrunet saying that we should remove the usage of newmat. Especially because it seems this library is really old and not maintained anymore. 
http://robertnz.net/nm_intro.htm
Latest update seems to be 2008 for newmat11 (beta)...

alxbilger:
> Any interest to keep newmat ? I remember a discussion during a sofadevmeeting with @jnbrunet saying that we should remove the usage of newmat. Especially because it seems this library is really old and not maintained anymore.
> http://robertnz.net/nm_intro.htm
> Latest update seems to be 2008 for newmat11 (beta)...

Good point. I just saw an opportunity to easily benefit from the type-parameterized tests that I introduced for `BaseMatrix`.

An option is to remove it, but it could also be moved into an external plugin. Let's discuss it.

jnbrunet:
Hey @alxbilger and @fredroy ,

Indeed, newmat is super old. I also have no idea on how it compares to modern implementations such as Eigen. 

I think however newmat is the only implementation in SOFA of dense solvers (LU and CG), but I may be wrong here. If you remove it, you might want to notify people that no dense solver are supported anymore in SOFA (if you do find people that use such solvers....). Or just implement an Eigen version of dense LU/CG solvers, it should not be too hard.

fredroy:
Ready when #2310 is merged

PR #2340, opened 2021-09-13
[image] Restore ability to select a subset of instanciations 

fredroy:
image plugin had the choice at one moment to compile just its classes with a subset of image type (Byte, etc)
The macro was always activated right now, so there was no choice but to compile always ALL the image types, and the image plugin was really long to compile.

This PR restores this behavior, with a third choice: to not compile any instanciations.
Some people would like to use some headers (especially CImgData.h) for other plugin, without dragging the whole compilation.

Anyway this PR introduces a new cmake option called PLUGIN_IMAGE_COMPILATION_SET restrained to 3 values (a combo box in cmake-gui actually)
- standard: a few selected types of image is chosen
- full: all the the instanciations will be compiled
- none: no instanciation, so you will only compile initImage.cpp effectively.


fredroy:

alxbilger:
Any reason CImgData is not in CImgPlugin ? (not a suggestion, just out of curiosity)

fredroy:
I did the move there #2307 (and explained why)

PR #2357, opened 2021-09-22
[Sofa.Type, Sofa.Topology] Fix testEdgeBuffer with clang (and add constexpr default constructors)

fredroy:
Happening on **clang ONLY** (linux or mac, same) 

Kind of 'mysterious' behavior happened here: https://github.com/sofa-framework/sofa/blob/65a54acb1d47c09acf294333ddfb5a5d57c1d0fe/SofaKernel/modules/SofaCore/src/sofa/core/topology/Topology.h#L86

core::topology::InvalidEdge was getting (0,0) instead of the expected (invalidID, invalidID).

_Assumption (really just guessing)_
the inline static initialization (with clang) is done through a first stage compilation where the default constructor of Edge has not been generated yet, hence calling a (really) default constructor initializing array with a zero-initialization. 
One hint came from the fact that if I switched the inline static init in the h, with a "classical" static declaration in h + initialization in the cpp, it was correctly initializing InvalidEdge with (invalidID, invalidID)...

Instead of just reverting to the classical static init, I have put all those inline static to constexpr members, leading to the fact that fixed_array's default constructor was not constexpr (forgotten 😗)
Anyway, now everything is constexpr and the inline static initialization is correct now....

Bear in mind that it seems that GCC and MSVC are different than CLANG in that regard.

### DIFF
--- a/SofaKernel/modules/Sofa.Topology/src/sofa/topology/Edge.h
+++ b/SofaKernel/modules/Sofa.Topology/src/sofa/topology/Edge.h
@@ -30,5 +30,5 @@ namespace sofa::topology
 {
     using Edge = sofa::topology::Element<sofa::geometry::Edge>;
 
-    constexpr static Edge InvalidEdge;
+    static constexpr Edge InvalidEdge;
 }
--- a/SofaKernel/modules/Sofa.Topology/src/sofa/topology/Edge.h
+++ b/SofaKernel/modules/Sofa.Topology/src/sofa/topology/Edge.h
@@ -30,5 +30,5 @@ namespace sofa::topology
 {
     using Edge = sofa::topology::Element<sofa::geometry::Edge>;
 
-    inline static const Edge InvalidEdge;
+    constexpr static Edge InvalidEdge;
 }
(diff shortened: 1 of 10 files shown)


fredroy:

alxbilger:
I suspect a Static Initialization Order Fiasco. I am not sure your changes really fixes the issue. It may be because your changes changed the initialization order. Luckily, it's the good one, but other changes could modify the initialization order later.
It's just a guess...

fredroy:
> 
> 
> I suspect a Static Initialization Order Fiasco. I am not sure your changes really fixes the issue. It may be because your changes changed the initialization order. Luckily, it's the good one, but other changes could modify the initialization order later. It's just a guess...

Thank you for the keywords `Static Initialization Order Fiasco` , I never really encountered such a problem before (lucky I guess).
It seems that the fact that the initialization of the static constexpr variable is done at compilation time (oh really?) solves the SIOF (e.g https://www.py4u.net/discuss/88472)

alxbilger:
@fredroy It's good to know that it does not apply in case of `constexpr`.
You just kept a static variable (`InvalidSet`), which is not `constexpr`. SIOF can apply. `InvalidSet` constructor can be `constexpr` only from C++20.

fredroy:
> @fredroy It's good to know that it does not apply in case of `constexpr`.
> You just kept a static variable (`InvalidSet`), which is not `constexpr`. SIOF can apply. `InvalidSet` constructor can be `constexpr` only from C++20.

You dont miss a thing 😆 Indeed std::vector constructor is constexpr in c++20 only.
But luckily in our case, InvalidSet is supposed to be an empty vector, so it will be zero-construct in any case. (so we will get an empty vector whatever the init order is)

PR #2361, opened 2021-09-27
[examples] Make scene resolution independent + doc

alxbilger:
The MechanicalMatrixMapper example scene deserved a bit of love.

- Converted to Python3
- Add documentation (the concept of rigidification deserves a more detailed doc somewhere else)
- It is possible to change the resolution of the grid while keeping the same behavior: 1) `indexPairs` is built algorithmically, 2) ConstantForceField is applied using a BoxROI

### DIFF
--- a/examples/Components/animationloop/MechanicalMatrixMapper.pyscn
+++ b/examples/Components/animationloop/MechanicalMatrixMapper.pyscn
@@ -1,55 +1,93 @@
+# This Python scene requires the SofaPython3 plugin
+#
+# This scene shows the use of MechanicalMatrixMapper. A beam made of hexahedrons is simulated. One part is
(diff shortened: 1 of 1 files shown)


fredroy:
As it is using SofaPython3 to run, should not it be ... in the SofaPython3 plugin 😶 ?

alxbilger:
For the record, the scene was already here before this PR, and it required Python...

There are pros and cons about putting Python scenes in SOFA. Let's discuss it.

For this scene in particular, Python is required as the scene has a Node with two parent Nodes. It also allows to compute `indexPairs`.

fredroy:
@alxbilger I understand your point, especially because it is the only scene showing an example on how to use MechanicalMatrixMapper...

PR #2364, opened 2021-09-28
[Sofa.Core] Use links instead of private std::list

alxbilger:
A `ConstraintCorrection` stores the list of constraint solvers using it. This is mainly to notice them to stop using them when they are destroyed (in the cleanup method).
This list was a `std::list`. I suggest a `MultiLink` instead.


SofaKernel/modules/SofaCore/src/sofa/core/behavior/ConstraintCorrection.inl
```diff
@@ -45,24 +37,24 @@ void ConstraintCorrection< DataTypes >::init()
...
     }
+    l_constraintsolvers.clear();
     sofa::core::behavior::BaseConstraintCorrection::cleanup();
 }
 
 template <class DataTypes>
 void ConstraintCorrection<DataTypes>::addConstraintSolver(core::behavior::ConstraintSolver *s)
 {
-    constraintsolvers.push_back(s);
+    l_constraintsolvers.add(s);
```
hugtalbot: add and remove are returning bool right? maybe we could use the returned value for warning/info?

alxbilger: done

fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):

fredroy:
One (new) scene is failing: `SceneTests.applications/plugins/LMConstraint/examples.DOFBlockerLMConstraint.scn`
The error message is: ` [ERROR] [UncoupledConstraintCorrection(default67)] Cannot remove the requested constraint solver` 
It is possible that the scene itself was wrong but it is highlighted here because you add a msg_error()

alxbilger:
[ci-build][with-all-tests]

alxbilger:
The error in the tests came from (applications/plugins/LMConstraint/src/LMConstraint/LMConstraintSolver.cpp) in `init`:
```cpp
listConstraintCorrection[i]->removeConstraintSolver(this);
listConstraintCorrection[i]->addConstraintSolver(this);
```
The first line removed the element from the list whereas the element was never added to the list (we are in the `init` method). It is added in the second line.

PR #2378, opened 2021-10-02
[All] Remove warnings

fredroy:
From Clang12 & gcc11

Mostly warnings about 
- signed int to unsigned int (and vice versa)
- invertMatrix return value was not retrieved (replace with an assert for the moment but exceptions would be a better way)
- deprecated things


fredroy:
[ci-build][with-all-tests]

guparan (approved):

SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/MechanicalObject.inl
```diff
@@ -783,7 +783,7 @@ void MechanicalObject<DataTypes>::applyScale(const SReal sx,const SReal sy,const
     const sofa::type::Vec<3,Real> s((Real)sx, (Real)sy, (Real)sz);
     for (unsigned int i=0; i<x_wA.size(); ++i)
     {
-        for (unsigned int j=0; j<std::min(3, (int)DataTypes::Coord::total_size); ++j)
+        for (unsigned int j=0; j< std::min(3u, (unsigned int)DataTypes::Coord::total_size); ++j)
```
alxbilger, line 786: `total_size` is an enum, implicitly cast to an `int`.
```suggestion
        for (int j=0; j< std::min(3, DataTypes::Coord::total_size); ++j)
```

alxbilger (approved):

SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/MechanicalObject.inl
fredroy: We ought to remove those enum to make a pesudo-constexpr field, and to use "true" constexpr.

PR #2380, opened 2021-10-04
[Sofa.Helper] More tests for NameDecoder + fix them

alxbilger:
There were only a few tests on NameDecoder in `SofaKernel/modules/SofaCore/SofaCore_test/objectmodel/BaseClass_test.cpp`. I created a dedicated file to test the main functions in NameDecoder.

It required changes in some of the algorithms to comply with the added test:

- Support of nested classes
- Spaces are processed in the first pass of the algorithm, instead of adding a second pass just for spaces. This is also to detect when spaces must be skipped or not (always skipped before). For example, it must not be skipped in `unsigned int`.
- Returned string get a balanced number of opening chevrons and closing chevrons.

Note: Anonymous namespaces are not supported (yet?).
Note 2: The algorithms in `NameDecoder` are complicated to read. It is hard to anticipate the result. IMO, it's due to the fact that it is written to be executed in a single pass. I think allowing multiple passes would help to read and debug these functions.

`decodeTemplateName` was never called. I call it now in `getOverridableTemplateName()` instead of assuming there is no template.


alxbilger:
[ci-build][with-all-tests]

PR #2425, opened 2021-10-15
[Sofa.Helper] PluginManager: Add optional way to check if a plugin is init at the load stage

fredroy:
No information is given when the initExternalModule entry point is called.
For 99.999% of the plugins, there is nothing special in it but some (SofaPython3, I am looking at you) is doing a lot of "failable" things when it is called.
This PR is a potential solution to this problem, by adding a (optional) entry point to check if the plugin reports by itself that it was loaded successfully.

This allows SOFA to not crash if a problem arises in SofaPython3 when calling PythonEnvironment::Init() 

It is not really necessary for the "ci depends on" as the two PR can compile without needing the other one; but it makes sense to merge them in the same time I think.

[ci-depends-on https://github.com/sofa-framework/SofaPython3/pull/210]

### DIFF
--- a/SofaKernel/modules/SofaHelper/src/sofa/helper/system/PluginManager.cpp
+++ b/SofaKernel/modules/SofaHelper/src/sofa/helper/system/PluginManager.cpp
@@ -192,7 +192,7 @@ bool PluginManager::loadPluginByPath(const std::string& pluginPath, std::ostream
     {
         if (!p.moduleIsInitialized())
         {
-            const std::string msg = pluginPath + " reported an error while trying to initialize. " + "This plugin will not be loaded.";
+            const std::string msg = pluginPath + " reported an error while trying to initialize. This plugin will not be loaded.";
             msg_error("PluginManager") << msg;
             if (errlog) (*errlog) << msg << std::endl;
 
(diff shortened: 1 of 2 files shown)


fredroy:
[ci-build][with-all-tests]

SofaKernel/modules/SofaHelper/src/sofa/helper/system/PluginManager.cpp
```diff
@@ -186,6 +187,20 @@ bool PluginManager::loadPluginByPath(const std::string& pluginPath, std::ostream
     m_pluginMap[pluginPath] = p;
     p.initExternalModule();
 
+    // check if the plugin is initialized (if it can report this information)
+    if (getPluginEntry(p.moduleIsInitialized, d))
+    {
+        if (!p.moduleIsInitialized())
+        {
+            const std::string msg = pluginPath + " reported an error while trying to initialize. " + "This plugin will not be loaded.";
```
alxbilger:
```suggestion
            const std::string msg = pluginPath + " reported an error while trying to initialize. This plugin will not be loaded.";
```

alxbilger (approved):

guparan (approved):

PR #2427, opened 2021-10-18
[SofaGeneralLinearSolver] Call super init() to check required data

### LINKED ISSUE #2426: Make sure data defined as required generate errors if not
Make sure the data defined as required (`d_myData.setRequired(true);`) generates an error if it is not indeed defined

To do so, the call to super init() must be done since this check is done in `BaseObject::init()`

Initiated in #2419

hugtalbot:
Fix #2426


alxbilger (changes requested):
If you don't override `init()`, its base class `init()` will be called. So, I don't think this PR is required.

Issue #2432, opened 2021-10-19
shadow variable


maddin200:
/sofa-master/applications/plugins/image/imagetoolbox/labelpointsbysection/labelpointsbysectionimagetoolbox.h line 111
Code:
     std::ostringstream values;

        values << d_axis.getValue();
        file << values.str() << "\n";

        for(int i=0;i<size;i++)
        {

            std::ostringstream values; // <-- shadow variable

hugtalbot:
as usual @maddin200 it would be great if you could PR all your fixes :+1:

maddin200:
fixed by #2437

PR #2437, opened 2021-10-20
[image] fix shadow variable #2432

maddin200:
rename variable


alxbilger (commented):
Thanks for the PR!
On my side, I'm not a fan of naming a variable with a numeric suffix, just because there is already another one with the same name.
For this specific case, I would prefer to surround lines 103-106 by curly brackets to limit the scope of `std::ostringstream values`.

maddin200:
Maybe only one declaration is sufficient.

Issue #2392, opened 2021-10-09
Update of texture coordinates in OglModel


hugtalbot:
Further to #2300 the coordinate textures do no update anymore when using a TextureInterpolation as input of texcoord

Due [this change](https://github.com/sofa-framework/sofa/pull/2300/files#diff-ebb9cb0da6bb4d64a21079624cb5082e2dc4d807d5d711a148d3ef756661418bR994):

``` cpp
m_vtexcoords.setParent(nullptr); // manually break the data link to follow topological changes
```

deleting this line solves the issue, but I would need your insight @epernod on what was exactly intended here.

epernod:
example scene, log and backtrace, please :)

hugtalbot:
@epernod the _examples/Components/forcefield/TetrahedronDiffusionForceField.scn_

hugtalbot:
See #2412 
explanation : topology comes from the IdentityMapping which maps a topology possibly with topological changes while the textcoords are coming from the TextureInterpolation (here no able to handle topo changes). To avoid crash the link is cut. To avoid this, only use: `handleDynamicTopology="0"`

epernod:
As mentioned above, this has been fixed by PR #2415 and #2445

PR #2300, opened 2021-08-18
[SofaBaseVisual] Fix VisualModelImpl to use topologyData and callback to handle topological changes

epernod:
This PR depends on PR #2299 (should be rebased)

1. Use callback to update the list of triangles/quads and vertices when topological changes happened.
2. Add also callback on texcoord to update texture coordinate when mesh changed.

- Update of texture coordinates was buggy when removing points

- Adding new points where created with texCoord(0, 0)

-> Using the callback on texcoords, use ancestors and coefficient of added point to apply the good texCoords

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2445, opened 2021-10-21
[examples] Add handleDynamicTopology in OglModel with dynamic texcoords

hugtalbot:
Further to #2300 and issue was noticed in #2392 : texture coordinates of the OglModel were not updating anymore from parent (InterpolationEngine recomputing the tex coord) since the link was broken.

It has been decided that if dynamic tex coordinates are managed by the parent component the option `handleDynamicTopology="0"` could be used.

The two corresponding scene are changed here


epernod (approved):
It would be interesting to see what happened if TextureInterpolation were supporting topologicalChanges and then all Data are updated inside the OglModel at next texCoords.GetValue after a topological change... 
I would say 50% chance of crash... but this is another story!

Issue #2066, opened 2021-05-06
[SofaGuiQt] QTreeWidget is not correctly updated with a DefaultCollisionGroupManager


alxbilger:
The component `DefaultCollisionGroupManager` manipulates the scene graph structure, by creating nodes, and moving nodes or objects. It happens when there are contacts.
The scene graph is correctly updated, but the tree in the runSofa Qt panel is not correctly updated according to the new graph structure.
Here is an example:
![Capture d’écran du 2021-05-06 23-25-18](https://user-images.githubusercontent.com/10572752/117368218-80efff80-aec3-11eb-9169-44da8211f75c.png)
In the screenshot, we can see that the root node has only one child node in the Qt panel (`SaladBowl`), but it has two child nodes in its Data `child` (`SaladBowl` and `collision1`). `collision1` is created by `DefaultCollisionGroupManager` after a contact and should contain other nodes and objects.
The issue arises in `examples/Tutorials/Collision/MultipleObjectsDynamicCollisionGroups.scn` where I took the screenshot, but also in other scenes with a `DefaultCollisionGroupManager`.

Note: `DefaultCollisionGroupManager` is optional in a scene.

fredroy:
Not really related to your issue itself (GUI) but are you sure that `DefaultCollisionGroupManager ` is working well? last time I tried, if you let the simulation running, you can see in `MechanicalObjects` keeping allocating vectors (without deleting them)... leading to a big slowdown and memory consumption.

alxbilger:
No I am not sure. And I strongly suspect that it is related to the crash revealed in #2047. If what you say is true, it is a critical bug: I count 156 scenes (.scn) in Sofa where there is a `DefaultCollisionGroupManager`

alxbilger:
As a workaround, the Qt tree view can be refreshed by switching to another tab (for example 'View'), and going back into the 'Graph' tab.

PR #2458, opened 2021-11-05
[SofaGuiQt] Fix graph update on startup

fredroy:
(should) Fix issue #2066

The GraphListener for the visual Graph is frozen when runSOFA is started (and never unfrozen)
The temp fix to swap tabs was unfreezing it actually.
But I think letting it unfrozen is also a not-so-good idea as the listener will update a lot (how many? I dont know) the graph for nothing.

Now this PR unfreeze/freeze the visual graph at each timestep of the simulation.

### DIFF
--- a/modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
+++ b/modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
@@ -2147,6 +2147,7 @@ void RealGUI::interactionGUI ( bool )
 //called at each step of the rendering
 void RealGUI::step()
 {
+    simulationGraph->Unfreeze();
     sofa::helper::AdvancedTimer::begin("Animate");
 
     Node* root = currentSimulation();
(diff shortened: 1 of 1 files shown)


epernod (approved):
works for me

alxbilger:
@fredroy Why does it have to be frozen at all?

fredroy:
> 
> 
> @fredroy Why does it have to be frozen at all?

I assumed that when unfrozen, unnecessary code would be called (I did not really dig as the gui code is quite a mess)

alxbilger (approved):

alxbilger:
Note that the graph is not updated if a slave is added during a time step. The tab switch is still a workaround.

alxbilger:
@fredroy This change seems to have a bad impact on the performances. I noticed that on the caduceus. Could you test it on your end?

alxbilger:
@fredroy a kind reminder to verify that this PR has a bad impact on the performances.

fredroy:
I am vsync-limited on my setup so cannot really test.
Lets simply revert this PR, I will see later

PR #2459, opened 2021-11-07
[SofaCore] Fix potential memory failure on TopologyData Add method 

epernod:
Error due to passing const & to temporary vector created inside the method. Use ref to static vector instead.


alxbilger (commented):
Does it work for multiple TopologyData in the scene? Is it possible that more than one component use the same static variable, leading to a bug?

epernod:
I guess yes because this is an empty variable. It is only used in the callbacks code to check: .empty() and not using it.

epernod:
[ci-build][with-all-tests]

PR #2462, opened 2021-11-08
[examples] Add another way to link two independent objects

alxbilger:
A mapping-based scene design is introduced in `examples/Components/forcefield/StiffSpringForceField_simple.scn`.

The force field is defined on a mechanical state which is the fusion of two independent sets of points.

### DIFF
--- a/examples/Components/forcefield/StiffSpringForceField_simple.scn
+++ b/examples/Components/forcefield/StiffSpringForceField_simple.scn
@@ -4,6 +4,7 @@
     <RequiredPlugin name="SofaDeformable"/>
     <RequiredPlugin name="SofaImplicitOdeSolver"/>
     <RequiredPlugin name="SofaSparseSolver"/>
+    <RequiredPlugin name="SofaMiscMapping"/>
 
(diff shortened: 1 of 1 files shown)


alxbilger:
[ci-build][with-all-tests]

examples/Components/forcefield/StiffSpringForceField_simple.scn
```diff
@@ -48,4 +49,42 @@
...
+                4 1 50 1 1
+                4 2 50 1 1
+                4 3 50 1 1
+                4 5 50 1 1
+                5 1 50 1 1
+                5 2 50 1 1
+            " showArrowSize="0.05" drawMode="2"/>
+        </Node>
+    </Node>
+
```
hugtalbot, line 89: would it not deserve a separated example scene?

alxbilger: It's possible. For now, I decided not to merge this PR because the matrices are not mapped in this example. We need to have a conversation.

PR #2470, opened 2021-11-09
[examples] add a warning comment in the scene header

alxbilger:
The scene `examples/Components/animationloop/FreeMotionAnimationLoop.scn` takes a lot of time to initialize. I would like to warn the user that it is normal, that SOFA did not freeze.

### DIFF
--- a/examples/Components/animationloop/FreeMotionAnimationLoop.scn
+++ b/examples/Components/animationloop/FreeMotionAnimationLoop.scn
@@ -1,4 +1,10 @@
 <?xml version="1.0" ?>
+<!--
+WARNING: this scene uses a PrecomputedConstraintCorrection which has a heavy initialization step. It may take some time
+to load the scene. To cache the result, set the recompute Data of PrecomputedConstraintCorrection to false.
+To speed up the collision detection, replace BVHNarrowPhase by ParallelBVHNarrowPhase located in the MultiThreading plugin.
+-->
+
 <Node name="root" dt="0.01" gravity="0 981 0">
     <RequiredPlugin pluginName="SofaOpenglVisual"/>
     <RequiredPlugin pluginName='SofaConstraint'/> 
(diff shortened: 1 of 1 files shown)


examples/Components/animationloop/FreeMotionAnimationLoop.scn
```diff
@@ -9,7 +15,7 @@
     <RequiredPlugin pluginName='SofaSimpleFem'/> 
 
     <VisualStyle displayFlags="showBehaviorModels"/>
-    <FreeMotionAnimationLoop/>
+    <FreeMotionAnimationLoop parallelCollisionDetectionAndFreeMotion="true"/>
```
hugtalbot, line 18: this means that now MT codes are accepted to be active by default
I am fine with it, I just wanted to point this out 👉

hugtalbot (approved):

examples/Components/animationloop/FreeMotionAnimationLoop.scn
alxbilger: What do you mean? The default value for `parallelCollisionDetectionAndFreeMotion` is still false.

hugtalbot: ❓ `parallelCollisionDetectionAndFreeMotion="true"`

epernod: I think the full question is: what is the behavior if this option is set to true but the multiThreading plugin is not compiled.
Is there by default a TaskScheduler?

alxbilger: `parallelCollisionDetectionAndFreeMotion` is a Data in `FreeMotionAnimationLoop`. This Data is `false` by default, if you don't overwrite it in your scene. This is what I do in the caduceus, but the other scenes keep their value to false.

The multithreading capabilities are in the SOFA core, **NOT** in the MultiThreading plugin. There are some components exploiting the multithreading capabilities in SOFA. `FreeMotionAnimationLoop` is an example. Therefore, your don't need to compile the MultiThreading plugin in order to compute the collision detection and the free motion in parallel.

PR #2478, opened 2021-11-12
[SofaSimpleFem] Check topology in draw

alxbilger:
In `HexahedronFEMForceField<DataTypes>::draw`, the topology pointer is checked first before being used. It avoids a crash if the topology cannot be found in the initialization.

### DIFF
--- a/SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
+++ b/SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
@@ -127,7 +127,7 @@ void HexahedronFEMForceField<DataTypes>::init()
     }
 
 
-    if ( m_topology->getNbHexahedra() <= 0 )
+    if ( m_topology->getTopologyType() != sofa::core::topology::TopologyElementType::HEXAHEDRON )
     {
         msg_error() << "Object must have a hexahedric MeshTopology." << msgendl
                     << " name: " << m_topology->getName() << msgendl
--- a/SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
+++ b/SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
@@ -102,8 +102,11 @@ HexahedronFEMForceField<DataTypes>::HexahedronFEMForceField()
 template <class DataTypes>
 void HexahedronFEMForceField<DataTypes>::init()
 {
-    if(_alreadyInit)return;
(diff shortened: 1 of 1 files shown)


SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
```diff
@@ -1191,6 +1191,7 @@ void HexahedronFEMForceField<DataTypes>::draw(const core::visual::VisualParams*
     if (!vparams->displayFlags().getShowForceFields()) return;
     if (!this->mstate) return;
     if (!f_drawing.getValue()) return;
+    if (!m_topology) return;
```
fredroy, line 1200: Wont it be more consistent to rely on the componentState ? I saw that it is set to invalid if the topology is not found at init()

alxbilger: I agree that we must rely on the component state. This is added in my last commit. But I also prefer an extra security and keep the pointer check.

SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
```diff
@@ -124,14 +127,16 @@ void HexahedronFEMForceField<DataTypes>::init()
     }
 
 
-    if( m_topology->getNbHexahedra()<=0 )
+    if ( m_topology->getNbHexahedra() <= 0 )
```
epernod, line 130: same as a recent PR. It is better to check the type of element in tricky case when you would like to start the simulation with empty mesh.

```suggestion
    if ( m_topology->getTopologyType() != sofa::core::topology::TopologyElementType::HEXAHEDRON )
```

epernod (changes requested):

SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
alxbilger: Does it mean topologies with multiple types of elements cannot be supported?

alxbilger:
[ci-build][with-all-tests]

SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/HexahedronFEMForceField.inl
epernod: Ok @fredroy told me this is not working. So just revert my suggestion and we will see this point later.

PR #2501, opened 2021-11-23
[SofaSparseSolver] Introduce FillReducingOrdering

alxbilger:
A new DataEngine is introduced: `FillReducingOrdering`

A DataEngine to reorder the degrees of freedom in a mesh in order to reduce fill-in in sparse matrix factorization.

In other terms, the algorithm minimizes the number of non-zeros entries in the factorization of the sparse matrix of a FEM problem by reordering the degrees of freedom.

The implementation is based on METIS.

Note: some of the direct linear solvers embed such a reordering internally (e.g. SparseLDLSolver), but not all of them (e.g SparseLUSolver, SparseCholeskySolver etc).

An example is also introduced (`examples/Components/engine/FillReducingOrdering.scn`).
Here is the comparison of the two system matrices:

Initial matrix:
initial_matrix00001

Reordered matrix:
reorder_matrix00001

It improves a lot the performances when the linear solver does not embed a fill-in reducing reordering. This is the case of the `SparseLUSolver` in the example scene.
Rough framerate without reordering: **2.5 FPS**
Rough framerate with reordering: **25 FPS**


alxbilger:
[ci-build][with-all-tests]

fredroy:
By curiosity, do you think using mt-metis (metis with openmp in a nutshell) could be useful ? 
I am quite not sure because I suppose the speedup would come for a mesh with a huge number of nodes but still... 🧐

alxbilger:
> By curiosity, do you think using mt-metis (metis with openmp in a nutshell) could be useful ? I am quite not sure because I suppose the speedup would come for a mesh with a huge number of nodes but still... 🧐

@fredroy I don't think it's a priority. The reordering is computed only at initialization. And you can always export your new mesh for later use.

guparan:
LGTM but the new example `FillReducingOrdering.scn` crashes (timeout) on all platforms :cry:

alxbilger:
[ci-build][with-all-tests]

fredroy (approved):

PR #2518, opened 2021-11-30
[SofaBaseMechanics] Use directly clear() when resetting force in MechanicalObject

fredroy:
...instead of assign operator + defaultconstructor (which in fine do the same clear() function)

Micro optimization but still 😗
(nice french idiom for that: "ça ne mange pas de pain")

(on examples/Components/DiagonalMass.scn)
Before:
Capture1
After:
Capture2

_______________________________________


alxbilger (commented):
Interesting.
I wonder if adding a move assignment operator in `fixed_array` would produce the same result...

fredroy:
[ci-build][with-all-tests]

alxbilger:
@fredroy https://github.com/alxbilger/SofaBenchmark/pull/6 See the difference between `std::array` and `sofatypefixedarray3f` and `sofa::type::Vec3f`.
The changes you suggest should not make a difference if `sofa::type::Vec3f` was implemented like `std::array`.

epernod (approved):

alxbilger:
Ah! C-style arrays are not movable. I think it explains the lower performances with the operator=

PR #2522, opened 2021-12-01
[all] Update collision response names

hugtalbot:
This PR aims at cleaning and improving the response method available in SOFA.
Here are the changes on the user standpoint:

- default :arrow_right: PenalityContactForceField
- FrictionContact :arrow_right: FrictionContactConstraint
- ray :arrow_right: RayContact
- registration :arrow_right: RegistrationContactForceField
- stick :arrow_right: StickContactForceField
- StickContactConstraint (remains unchanged)
- distanceLMConstraint (remains unchanged)

Personally, I love it :heart: 

- Fix #1966
- NOTE: a documentation page should soon join the PR if it pleases the reviewers

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2523, opened 2021-12-01
[All] Add warnings if components are missing to solve a specific response

hugtalbot:
Based on #2522  this PR adds a sceneCheck to guide users when using collision responses.
For some, a specific animation loop / constraint solver is needed. This is added in the SceneCheckContactResponse !


alxbilger:
I recommend to review this PR with this comparison link: https://github.com/hugtalbot/sofa/compare/change_collision_response_names_202111...add_checks_warnings_response_202111
since it is polluted by https://github.com/sofa-framework/sofa/pull/2522

modules/SofaGraphComponent/src/SofaGraphComponent/SceneCheckCollisionResponse.cpp
```diff
@@ -0,0 +1,113 @@
...
+using sofa::simulation::Node;
+
+const std::string SceneCheckCollisionResponse::getName()
+{
+    return "SceneCheckCollisionResponse";
+}
+
+const std::string SceneCheckCollisionResponse::getDesc()
+{
+    return "Check there is not duplicated name in the scenegraph";
```
alxbilger: This is a copy/paste. Please adapt the description

modules/SofaGraphComponent/src/SofaGraphComponent/SceneCheckCollisionResponse.cpp
```diff
@@ -0,0 +1,113 @@
...
+    else
+    {
+        m_checkDone=true;
+        if( nbContactManager!= 1 )
+        {
+            m_message << "Only one DefaultContactManager is needed."<< msgendl;
+        }
+        else
+        {
+            std::string response = contactManager[0]->response.getValue().getSelectedItem();
```
alxbilger:
```suggestion
            const std::string response = contactManager[0]->response.getValue().getSelectedItem();
```

modules/SofaGraphComponent/src/SofaGraphComponent/SceneCheckCollisionResponse.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+            if ( response == "StickContactConstraint" )
+            {
+                sofa::core::behavior::BaseAnimationLoop* animationLoop;
+                root->get(animationLoop, sofa::core::objectmodel::BaseContext::SearchRoot);
+                if (!animationLoop || ( animationLoop && ( animationLoop->getClassName() != "FreeMotionAnimationLoop" )) )
+                {
+                    m_message <<"A FreeMotionAnimationLoop must be in the scene to solve StickContactConstraint" << msgendl;
+                }
```
alxbilger, line 77: What if there is an animation loop inheriting from `FreeMotionAnimationLoop`? Maybe you could use a `dynamic_cast` or `BaseClass::hasParent`.

alxbilger: The comment is also valid for your other searches

modules/SofaGraphComponent/src/SofaGraphComponent/SceneCheckCollisionResponse.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+                sofa::core::behavior::ConstraintSolver* constraintSolver;
+                root->get(constraintSolver, sofa::core::objectmodel::BaseContext::SearchRoot);
+                if (!constraintSolver || ( constraintSolver && ( constraintSolver->getClassName() != "GenericConstraintSolver" )) )
+                {
+                        m_message <<"A GenericConstraintSolver must be in the scene to solve StickContactConstraint" << msgendl;
```
alxbilger: Indentation

modules/SofaGraphComponent/src/SofaGraphComponent/SceneCheckCollisionResponse.cpp
```diff
@@ -0,0 +1,113 @@
...
+    std::vector<sofa::component::collision::DefaultContactManager*> contactManager;
+    root->get<sofa::component::collision::DefaultContactManager>(&contactManager, core::objectmodel::BaseContext::SearchDown);
+    Size nbContactManager = contactManager.size();
+    if( nbContactManager == 0 )
+    {
+        m_checkDone=true;
+    }
+    else
+    {
+        m_checkDone=true;
```
alxbilger:
```suggestion
    m_checkDone=true;
    const Size nbContactManager = contactManager.size();
    if( nbContactManager  > 0 )
    {
```
`m_checkDone` is set to `true` in any case

alxbilger (changes requested):
I just find it weird that a `SceneChecker` is required for this. I would have imagined this check in the contact manager or directly in the contact response.
A scene checker is independent from the contact manager. If the behavior of a contact changes (e.g. a new constraint solver becomes compatible with StickContactConstraint), the scene checker must be updated. And this is a problem if the developer is not aware of this scene checker or if the new constraint solver is in a plugin.
In any case, adding those warnings are a very good initiative!

modules/SofaGraphComponent/src/SofaGraphComponent/SceneCheckCollisionResponse.cpp
hugtalbot: no ... :angel: never copy paste code! it's bad ..

modules/SofaGraphComponent/src/SofaGraphComponent/SceneCheckCollisionResponse.cpp
hugtalbot: do you mean instead of the `getClassName` @alxbilger ?

I am not sure to get the point.
In `SceneCheckCollisionResponse` you are in SofaGraphComponent which cannot link against SofaConstraint (FreeMotionAnimationLoop, GenericConstraintSolver etc) since it would create a circular CMake dependency. This is the simplest way I found to implement the feature

alxbilger: I understand. Maybe (in a future PR) we can think about adding a SceneChecker from any module/plugin. If we were able to do that, we could put the scene checker in SofaConstraint

alxbilger (approved):

guparan:
@hugtalbot Please put a suitable title for v21.12 changelog

PR #2543, opened 2021-12-15
[SofaSparseSolver] Fix compilation error due to namespace change

bakpaul:
Fix compilation error due to namespace change

### DIFF
--- a/modules/SofaSparseSolver/src/SofaSparseSolver/PrecomputedLinearSolver.h
+++ b/modules/SofaSparseSolver/src/SofaSparseSolver/PrecomputedLinearSolver.h
@@ -134,7 +134,7 @@ private :
 };
 
 #if !defined(SOFA_COMPONENT_LINEARSOLVER_PRECOMPUTEDLINEARSOLVER_CPP)
-extern template class SOFA_SOFASPARSESOLVER_API PrecomputedLinearSolver< CompressedRowSparseMatrix<double> , FullVector<double> >;
+extern template class SOFA_SOFASPARSESOLVER_API PrecomputedLinearSolver< linearalgebra::CompressedRowSparseMatrix<double> , FullVector<double> >;
 #endif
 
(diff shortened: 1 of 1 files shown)


modules/SofaSparseSolver/src/SofaSparseSolver/PrecomputedLinearSolver.h
```diff
@@ -134,7 +134,7 @@ private :
 };
 
 #if !defined(SOFA_COMPONENT_LINEARSOLVER_PRECOMPUTEDLINEARSOLVER_CPP)
-extern template class SOFA_SOFASPARSESOLVER_API PrecomputedLinearSolver< CompressedRowSparseMatrix<double> , FullVector<double> >;
+extern template class SOFA_SOFASPARSESOLVER_API PrecomputedLinearSolver< linearalgebra::CompressedRowSparseMatrix<double> , FullVector<double> >;
 #endif
```
fredroy:
```suggestion
extern template class SOFA_SOFASPARSESOLVER_API PrecomputedLinearSolver< linearalgebra::CompressedRowSparseMatrix<double> , linearalgebra::FullVector<double> >;
#endif
```
Well spotted, as it could be only detected if somebody else than PrecomputedLinearSolver.cpp is including this header.
But shoud FullVector not be prefixed with linearalgebra namespace as well?

bakpaul: Yes you are right. I don't know why but only modifying the first one worked for me. It was certainly not the reason why it didn't compile in the first place. But well, at least I've spotted half of the problem...

fredroy: could be something before this head doing a typedef or something like that maybe.
I am adding the suggestion and will merge afterwards

PR #2550, opened 2021-12-17
[SofaSimpleFem] Refactor TetrahedronFEMForceField for better readability

alxbilger:
This is a tentative to refactor `TetrahedronFEMForceField` for better readability.

All the FEM methods have been dispatched into dedicated files.

The methods are now classes sharing the same interface (polymorphism), called by the component.

Advantages:
- `TetrahedronFEMForceField` is smaller
- `TetrahedronFEMForceField` had many similar codes (often identical). This is now avoided as all common code are in the base classes.
- It is easier to read the differences between the different FEM methods.
- We could imagine having a factory of FEM methods for `TetrahedronFEMForceField`. Then, the component could be extended in plugins by adding more methods into the factory.

To do:

- [ ] Clean
- [ ] Move some class members from `TetrahedronFEMForceField` to `TetrahedronFEMForceFieldImpl`
- [ ] Restore initial performances
- [ ] Fix broken features such as drawing rotations
- [ ] If everyone agrees, remove the assembled methods. It adds complexity while the resulted matrix is not used.
- [ ] Adapt (if necessary) the derived classes (multithreading, CUDA etc)

### DIFF
--- a/SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/TetrahedronFEMForceFieldImplPolar.h
+++ b/SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/TetrahedronFEMForceFieldImplPolar.h
@@ -1,0 +1,62 @@
+﻿/******************************************************************************
+*                 SOFA, Simulation Open-Framework Architecture                *
+*                    (c) 2006 INRIA, USTL, UJF, CNRS, MGH                     *
+*                                                                             *
+* This program is free software; you can redistribute it and/or modify it     *
+* under the terms of the GNU Lesser General Public License as published by    *
+* the Free Software Foundation; either version 2.1 of the License, or (at     *
+* your option) any later version.                                             *
+*                                                                             *
+* This program is distributed in the hope that it will be useful, but WITHOUT *
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or       *
+* FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License *
+* for more details.                                                           *
+*                                                                             *
+* You should have received a copy of the GNU Lesser General Public License    *
+* along with this program. If not, see <http://www.gnu.org/licenses/>.        *
+*******************************************************************************
+* Authors: The SOFA Team and external contributors (see Authors.txt)          *
+*                                                                             *
+* Contact information: contact@sofa-framework.org                             *
+******************************************************************************/
+#pragma once
+
+#include <SofaSimpleFem/TetrahedronFEMForceFieldImplCorotational.h>
+
+namespace sofa::component::forcefield
+{
+
+template<class DataTypes>
+class TetrahedronFEMForceFieldImplPolar : public TetrahedronFEMForceFieldImplCorotational<DataTypes>
+{
+protected:
+
+    using Inherit = TetrahedronFEMForceFieldImplCorotational<DataTypes>;
+    using Inherit::Transformation;
+    using Inherit::Coord;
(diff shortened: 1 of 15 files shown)


epernod:
arg... good input but it is not going to help my work on merging back this class with TetrahedralCorotationalFEMForceField...
I started the other way, by moving the "generic" methods into a TetrahedronFEMUtils

SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/TetrahedronFEMForceFieldImplPolar.h
```diff
@@ -0,0 +1,62 @@
...
+
+    void addForceAssembled(const FiniteElementArrays& finiteElementArrays) override;
+
+};
+
+
+
+}
+
+#include <SofaSimpleFem/TetrahedronFEMForceFieldImplPolar.inl>
```
damienmarchal, line 62: This is an un-usual pattern. 

In general the .inl is included where some templates instanciation happens and not in the .h. 
See... general comment in the end.

damienmarchal:
There is a lot of code in the .h. 
Would it be possible to have the same refactoring while keeping the headers clean.

alxbilger:
@damienmarchal @epernod This PR is a WIP. I did not finish it on purpose. The goal was to show a new organisation of the code and discuss about it. Thanks for having reviewed it. It means we can discuss about it on Wednesday ;)

SofaKernel/modules/SofaSimpleFem/src/SofaSimpleFem/TetrahedronFEMForceFieldImplPolar.h
alxbilger: I agree with your comment but only in the scope of SOFA. It's not so unusual to include .inl files in the header outside of SOFA. See also my other comment.

fredroy:
Some bench to see the perf loss, on examples/Components/forcefield/TetrahedronFEMForcefield (with only one beam with small method):
master: `TetrahedronFEMForceFieldScene>/1024 4504 ms`
PR:        `TetrahedronFEMForceFieldScene>/1024 5735 ms`
+~25% slower... 😥

alxbilger:
> Some bench to see the perf loss, on examples/Components/forcefield/TetrahedronFEMForcefield (with only one beam with small method): master: `TetrahedronFEMForceFieldScene>/1024 4504 ms` PR: `TetrahedronFEMForceFieldScene>/1024 5735 ms` +~25% slower... 😥

@fredroy I don't remember such a high difference. I don't think I worked on small deformation. Have you tried large def?

fredroy:
> > Some bench to see the perf loss, on examples/Components/forcefield/TetrahedronFEMForcefield (with only one beam with small method): master: `TetrahedronFEMForceFieldScene>/1024 4504 ms` PR: `TetrahedronFEMForceFieldScene>/1024 5735 ms` +~25% slower... 😥
> 
> @fredroy I don't remember such a high difference. I don't think I worked on small deformation. Have you tried large def?

For big deformation:
master: `TetrahedronFEMForceFieldScene>/1024       6717 ms`
PR: `TetrahedronFEMForceFieldScene>/1024       8100 ms`
so ~20% slower

Additional note, these benches are done using clang(13) on linux (and in a VM)

fredroy:
Update (as bench on scenes is unreliable)
- Linux/clang: `15.5sec` (master) vs `16.5sec` (pr)
- Windows/cl: `17.5sec` (master) vs `18.2sec` (pr)

Much less difference after all

PR #2554, opened 2021-12-21
[SofaDeformable] Make SpringForceField a ForceField (not an InteractionForceField)

alxbilger:
Contains https://github.com/sofa-framework/sofa/pull/2557

This PR is the first step toward the conversion of the interaction force fields into simple force fields. This PR concerns the class `SpringForceField` and all its derived classes, including `StiffSpringForceField` which is used extensively to create springs.

The changes maintain the backward compatibility with current scenes. However, the signature of some functions changed. That's why it's breaking. It is now impossible to construct a `StiffSpringForceField` between 2 objects in C++. Instead a utility function is provided.

To maintain the backward compatibility and to facilitate the design of a simulation scene, `StiffSpringForceField` continues to accept two Data attributes ('object1' and 'object2') corresponding to two different objects. However, they don't correspond to any Data or Link in the component. Actually, during its creation, the component will instantiate other components in order to link both objects into one. In the end, the component is able to work on a single mechanical state while keeping the same physical behavior.

The component now instantiates:

1. A new Node
2. A new mechanical state which will be the fusion of the provided objects, 
3. A `SubsetMultiMapping` that will make the link between the two provided objects and the new mechanical state
4. The spring force field

Since `SubsetMultiMapping` was initially in SofaMiscMapping, it has been moved into SofaBaseMechanics. A compatibility layer is available.

Other components using `StiffSpringForceField` had to be adapted to create also the extra components. An example is `AttachBodyPerformer` which creates the spring between an object and the mouse (shift + clic).

WARNING:
Since `StiffSpringForceField` is now necessarily located in a mapped node, its stiffness matrix is not mapped in case of a built linear system. The forces are then computed explicitly only. This issue will be addressed in the new design of the matrix assembly.

TODO:

- [x] A description for this PR
- [ ] Handle topological changes in `SpringForceField` -> see https://github.com/sofa-framework/sofa/pull/2653 and https://github.com/sofa-framework/sofa/pull/2729
- [x] Deal with `BoxStiffSpringForceField` -> see https://github.com/sofa-framework/sofa/pull/2621
- [x] Compilation and tests issues on CI (CUDA)
- [x] In `AttachBodyPerformer`, a Node remains in the scene when `AttachBodyPerformer` is destroyed

[ci-depends-on https://github.com/sofa-framework/SofaPython3/pull/224]


alxbilger:
[ci-build][with-all-tests]

alxbilger:
[ci-build][with-all-tests]

alxbilger:
[ci-build][with-all-tests]

fredroy:

alxbilger:

alxbilger:

alxbilger:
[ci-build][with-all-tests]

alxbilger:
[ci-build][with-all-tests]

SofaKernel/modules/SofaDeformable/src/SofaDeformable/SpringForceField.inl
```diff
@@ -177,6 +146,27 @@ SReal SpringForceField<DataTypes>::getPotentialEnergy(const core::MechanicalPara
...
 }
 
+template<class DataTypes>
+void SpringForceField<DataTypes>::addSpringForce(Real& ener, VecDeriv& f1, const VecCoord& p1, const VecDeriv& v1, VecDeriv& f2, const VecCoord& p2, const VecDeriv& v2, sofa::Index /*i*/, const Spring& spring)
+{
+    sofa::Index a = spring.m1;
+    sofa::Index b = spring.m2;
+    typename DataTypes::CPos u = DataTypes::getCPos(p2[b])-DataTypes::getCPos(p1[a]);
+    Real d = u.norm();
+    if( spring.enabled && d<1.0e-4 ) // null length => no force
```
fredroy, line 270: I know it was in the code before, but we could replace this magic number ? 😗

fredroy:
Just a note for the reviewers: SubsetMultiMapping was moved to SofaBaseMechanics for dependency reasons (but the problem will be solved in the new sofang architecture a priori)

SofaKernel/modules/SofaDeformable/src/SofaDeformable/SpringForceField.inl
alxbilger: What's the good practice in this case? Is there a global constant value somewhere?

alxbilger: Should I use
1. https://github.com/fredroy/sofa/blob/master/SofaKernel/modules/Sofa.Type/src/sofa/type/Vec.h#L34
2. https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon
3. Another
?

alxbilger: @fredroy any thoughts?

alxbilger:
This PR was considered to simplify the dev of #2777 but it is no longer required.

PR #2559, opened 2021-12-22
[SofaBoundaryCondition] Sanitize

alxbilger:
- Hiding local declaration 
- Parameter names do not match 
- Uninitialized class members 
- Lambda capture is never used 
- Local variable can be made const 
- Result of a postfix operator is discarded 
- add explicit specifier to make constructor non-converting 
- Member function can be made const 
- Variable can be moved to if-statement 
- Declaration and assignment can be joined 
- Return by const value

### DIFF
--- a/modules/SofaBoundaryCondition/src/SofaBoundaryCondition/FixedRotationConstraint.inl
+++ b/modules/SofaBoundaryCondition/src/SofaBoundaryCondition/FixedRotationConstraint.inl
@@ -87,9 +87,9 @@ void FixedRotationConstraint<DataTypes>::projectPosition(const core::MechanicalP
     for (unsigned int i = 0; i < x.size(); ++i)
     {
         // Current orientations
-        const sofa::type::Quat<SReal> Q = x[i].getOrientation();
+        const sofa::type::Quat<SReal>& Q = x[i].getOrientation();
         // Previous orientations
-        const sofa::type::Quat<SReal> Q_prev = previousOrientation[i];
+        const sofa::type::Quat<SReal>& Q_prev = previousOrientation[i];
 
         auto project = [](const Vec3 a, const Vec3 b) -> Vec3 {
             return (a * b) * b;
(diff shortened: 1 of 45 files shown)


fredroy:
[ci-build][with-all-tests]

modules/SofaBoundaryCondition/src/SofaBoundaryCondition/FixedRotationConstraint.inl
```diff
@@ -87,9 +87,9 @@ void FixedRotationConstraint<DataTypes>::projectPosition(const core::MechanicalP
     for (unsigned int i = 0; i < x.size(); ++i)
     {
         // Current orientations
-        sofa::type::Quat<SReal> Q = x[i].getOrientation();
+        const sofa::type::Quat<SReal> Q = x[i].getOrientation();
```
fredroy:
```suggestion
        const sofa::type::Quat<SReal>& Q = x[i].getOrientation();
```
In this case, you might be able to use a ref

modules/SofaBoundaryCondition/src/SofaBoundaryCondition/FixedRotationConstraint.inl
```diff
@@ -87,9 +87,9 @@ void FixedRotationConstraint<DataTypes>::projectPosition(const core::MechanicalP
...
         // Previous orientations
-        sofa::type::Quat<SReal> Q_prev = previousOrientation[i];
+        const sofa::type::Quat<SReal> Q_prev = previousOrientation[i];
```
fredroy:
```suggestion
        const sofa::type::Quat<SReal>& Q_prev = previousOrientation[i];
```
same (could not do a multi-line suggestion 😕)

fredroy (changes requested):
Overall a nice cleaning PR 👍

modules/SofaBoundaryCondition/src/SofaBoundaryCondition/FixedRotationConstraint.inl
alxbilger: Actually, you can select multiple lines before making the suggestion

PR #2562, opened 2021-12-23
[SofaMiscTopology] Introduce EdgeTopologyFrom2PointSets

alxbilger:
`EdgeTopologyFrom2PointSets` is an edge topology built from 2 point sets. It computes pairs of points based on the nearest distance from one point to the other point set. It is very useful to be used in combination with springs and SubsetMultiMapping.

This component may change depending on the work in #2554 

image

### DIFF
--- a/examples/Components/mapping/EdgeTopologyFrom2PointSets.scn
+++ b/examples/Components/mapping/EdgeTopologyFrom2PointSets.scn
@@ -1,0 +1,91 @@
+<?xml version="1.0"?>
+<Node name="root" dt="0.1" gravity="0 -9.81 0">
+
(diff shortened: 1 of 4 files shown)


epernod:
We should definitely have a talk at beginning of January. We are working on the same features. I usually use NearestPointROI engine + StiffSpringForceField
I also created a long time ago, SubsetTopology engine, which is doing the opposite (dividing a topology in 2, not using mapping), we could easily create the opposite like a mergeTopology engine.

alxbilger:
[ci-build][with-all-tests]

alxbilger:
@epernod Let me know your opinion on #2595.

PR #2567, opened 2021-12-31
[SofaGeneralSimpleFEM] Some cleaning in TriangularFEMForceFieldOptim

epernod:
- Remove unsued topologyData: d_vertexInfo, d_edgeInfo
- remove corresponding inner classes
- small cleanings in headers

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2570, opened 2022-01-03
[Sofa.Type] Simplify template rebinding

alxbilger:
When reading https://github.com/sofa-framework/sofa/pull/2567, I did not understand immediately how to read:

```cpp
typedef typename VecCoord::template rebind<TriangleInfo>::other VecTriangleInfo;
```

I did not know this technique to rebind templates, as since C++11, it is possible to template an alias.

So, instead of writing

```cpp
template<class T2> struct rebind
{
    typedef vector< T2,CPUMemoryManager<T2> > other;
};
```
I propose to write:
```cpp
template<class T2>
using rebind_to = vector< T2, CPUMemoryManager<T2> >;
```

To define `VecTriangleInfo`, we would write:

```cpp
using VecTriangleInfo = typename VecCoord::template rebind_to<TriangleInfo>;
```

IMO, it is a bit more readable. But we can do better.

I introduced convenience aliases in Rebind.h. With that, I can define `VecTriangleInfo` with:
```cpp
using VecTriangleInfo = sofa::type::rebind_to<VecCoord, TriangleInfo>;
```

`sofa::type::rebind_to` works with types having a `rebind_to` typedef (such that the one I introduced in vector_T.h), or on types having one template parameter.

Useful links:

- https://stackoverflow.com/a/58789407
- https://rvarago.github.io/2020/10/28/rebinding-template-template-parameter.html


alxbilger:
[ci-build][with-all-tests]

PR #2577, opened 2022-01-05
[SofaConstraint] Sanitize

alxbilger:
Cleaning!!!

### DIFF
--- a/modules/SofaConstraint/src/SofaConstraint/LCPConstraintSolver.cpp
+++ b/modules/SofaConstraint/src/SofaConstraint/LCPConstraintSolver.cpp
@@ -469,7 +469,7 @@ void LCPConstraintSolver::MultigridConstraintsMerge_Compliance()
 void LCPConstraintSolver::MultigridConstraintsMerge_Spatial()
 {
     const int merge_spatial_step = this->merge_spatial_step.getValue();
-    const int merge_spatial_shift = 0; // merge_spatial_step/2
+    constexpr int merge_spatial_shift = 0; // merge_spatial_step/2
     const int merge_local_levels = this->merge_local_levels.getValue();
     int numConstraints = _numConstraints;
     int numContacts = numConstraints/3;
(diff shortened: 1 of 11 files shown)


modules/SofaConstraint/src/SofaConstraint/LCPConstraintSolver.cpp
```diff
@@ -469,7 +469,7 @@ void LCPConstraintSolver::MultigridConstraintsMerge_Compliance()
 void LCPConstraintSolver::MultigridConstraintsMerge_Spatial()
 {
     const int merge_spatial_step = this->merge_spatial_step.getValue();
-    const int merge_spatial_shift = 0; // merge_spatial_step/2
+    constexpr int merge_spatial_shift = 0; // merge_spatial_step/2
```
fredroy, line 472: 🤔 should we not just remove this variable ?

fredroy:
[ci-build][with-all-tests]

modules/SofaConstraint/src/SofaConstraint/LCPConstraintSolver.cpp
alxbilger: It's an option, but from the comment, we can see that an intention is to set this variable to `merge_spatial_step/2`.
Based on the poor code quality in this module, I would keep it. It's not like it is the only code quality concern in this module. There are many other things to clean in this module.

fredroy:
2 scenes are crashing:
 - examples/Components/mapping.DeformableOnRigidFrameMappingConstraints.scn
 - examples/Components/animationloop.FreeMotionAnimationLoop.scn

Seens to be related to PrecomputedConstraintCorrection (which is modified in this PR)

modules/SofaConstraint/src/SofaConstraint/PrecomputedConstraintCorrection.inl
```diff
@@ -739,7 +734,7 @@ void PrecomputedConstraintCorrection<DataTypes>::applyContactForce(const lineara
                 {
                     DXbuf += appCompliance[ offset2 + i ] * Fbuf[i];
                 }
```
fredroy, line 736: I guess the problem is here (i is used in two loops)

alxbilger: Thanks for pointing to this direction

fredroy:
👍

fredroy (approved):

PR #2590, opened 2022-01-12
[Sofa.Helper] Introduce narrow_cast

alxbilger:
This is a way to explicit the narrowing conversions in SOFA. Instead of using the classic `static_cast`, this function can check if the value changed after the cast. It can happen in case of narrowing conversion, when the target type is smaller than the source type.
Since the check must be performed at run-time (affects the performances), it is enabled only in DEBUG mode.

The idea is to use this cast everywhere the compiler warns about possible data loss and it's not possible to avoid a narrow conversion (e.g. by adapting both target and source types). An example is provided in this PR in `MeshGmsh`.

### DIFF
--- a/SofaKernel/modules/SofaHelper/SofaHelper_test/narrow_cast_test.cpp
+++ b/SofaKernel/modules/SofaHelper/SofaHelper_test/narrow_cast_test.cpp
@@ -58,57 +58,45 @@ namespace sofa
         //size_t -> int: out of range
         {
             constexpr size_t a = static_cast<size_t>(std::numeric_limits<int>::max()) + 1;
-            if constexpr (sofa::helper::forceNarrowCastChecking)
-            {
-                EXPECT_THROW(sofa::helper::narrow_cast<int>(a), sofa::helper::narrowing_error);
-            }
-            else
-            {
-                constexpr int b = std::numeric_limits<int>::min(); // overflow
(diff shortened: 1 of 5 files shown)


SofaKernel/modules/SofaHelper/SofaHelper_test/narrow_cast_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+            {
+                EXPECT_THROW(sofa::helper::narrow_cast<int>(a), sofa::helper::narrowing_error);
+            }
+            else
+            {
+                const int b = std::numeric_limits<int>::min(); // overflow
+                EXPECT_EQ(sofa::helper::narrow_cast<int>(a), b);
+            }
+        }
+    }
```
fredroy, line 101: maybe add 2 tests on casting : 
 - a unsigned int to a int ? 
 - a int to a short or char ?
  (and see throwing error and stuff)

alxbilger: done

alxbilger:
[ci-build][with-all-tests]

alxbilger:

fredroy:

fredroy (approved):

PR #2620, opened 2022-01-20
[SofaCUDA] Add support for NearestPointROI

alxbilger:
`NearestPointROI` is now compatible with 3d Cuda types. It is still executed on the CPU though.

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2595, opened 2022-01-13
[SofaGeneralEngine] Extend features of NearestPointROI

alxbilger:
1. a clean of the `NearestPointROI` class.
2. More features are added to NearestPointROI:
- Generates more output format so they can be used in other components accepting only those formats. Now, it can be used with SubsetMultiMapping and EdgeSetTopologyContainer.
- Add a filter on the input dofs.

The example scene has been reworked. It now uses constraints. And springs have been added to show how to use the new features from this PR.

This PR replaces https://github.com/sofa-framework/sofa/pull/2562

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2621, opened 2022-01-20
[SofaGeneralObjectInteraction] BoxStiffSpringForceField is now a pack of components

alxbilger:
`BoxStiffSpringForceField` was a force field, deriving from `StiffSpringForceField`. Its only difference is in the `init()` step, where there is a matching of pairs of points, based on the minimal distances.

This behavior could be achieved with a `StiffSpringForceField` along with components which the goal is to compute the matching of pairs of points. This is done with 2 `BoxROI` and a `NearestPointROI`.
Therefore, `BoxStiffSpringForceField` is now a simple `BaseObject` that instantiates a pack of components that mimics the previous behavior.

Based on:
- https://github.com/sofa-framework/sofa/pull/2595
- https://github.com/sofa-framework/sofa/pull/2620


alxbilger:
[ci-build][with-all-tests]

examples/Components/interactionforcefield/BoxStiffSpringForceField.scn
```diff
@@ -33,14 +31,14 @@
             <BarycentricMapping input="@.." output="@VisualModel" />
         </Node>
         <Node name="Surf">
-	    <SphereLoader filename="mesh/liver.sph" />
+	        <SphereLoader filename="mesh/liver.sph" />
```
hugtalbot: indent issue

alxbilger: you are right! It is a mix between spaces and tabs

PR #2643, opened 2022-01-27
[SofaGuiQt][WindowProfiler] Add root tree + show overhead

alxbilger:
The root tree is now displayed in the tree. Previously, only the root children were displayed. This root item provides more information about the time step: the time that has not been measured by timers (children). It's in the columns 'Self'. In the example, 3% of the step is not measured, corresponding to 1 ms.

The second addition is the overhead due to timers processing. It is displayed in the left panel. Here, it takes 10 ms to process the incoming timers and update the chart.

image

### DIFF
--- a/modules/SofaGuiQt/src/sofa/gui/qt/SofaWindowProfiler.cpp
+++ b/modules/SofaGuiQt/src/sofa/gui/qt/SofaWindowProfiler.cpp
@@ -198,6 +198,8 @@ bool SofaWindowProfiler::AnimationStepData::processData(const std::string& idStr
 {
     type::vector<Record> _records = sofa::helper::AdvancedTimer::getRecords(idString);
 
+    m_totalTimers = 0;
+
     //AnimationSubStepData* currentSubStep = nullptr;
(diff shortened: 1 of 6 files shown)


guparan (approved):

epernod (approved):

modules/SofaGuiQt/src/sofa/gui/qt/SofaWindowProfiler.cpp
```diff
@@ -603,6 +606,7 @@ void SofaWindowProfiler::updateSummaryLabels(int step)
     label_stepValue->setText(QString::number(stepData->m_stepIteration));
     label_timeValue->setText(QString::number(stepData->m_totalMs));
     label_overheadValue->setText(QString::number(stepData->m_overheadMs));
+    label_timersCounterValue->setText(QString::number(stepData->m_totalTimers));
```
epernod, line 609: I can't see this number on the screenshot? It has been added after?

alxbilger: Indeed

PR #2644, opened 2022-01-28
[SofaBaseMechanics][SofaMiscForcefield] Auto-detect MassType

fredroy:
The 3 usual masses are templated on DataTypes and a TMassType, which is usually float/double if DataTypes is VecTypes and a special type RigidMass<X,Real> if DataTypes is a RigidTypes.

It appears to me that it could be automatic and we dont need to explicitely cite the RigidMass/Real for MassType as it seems to depend solely on the DataTypes.
Moreover it is confusing for writting scene, if you want to not rely of the automatic detection of templates (canCreate and stuff()) from the associated MechanicalObject. For example in `/Components/mapping/ArticulatedSystemMapping.scn`
```xml
<UniformMass template="Rigid3d,RigidMass<3u,double>" ..... >
```
This is so weird to explicitly write the masstype.
(Actually, the xml line is even not working, as it needs to be `Rigid3d,RigidMass<3,double>` but I suppose it was this way when the file was written)

~~In this PR, [std::conditional_t](https://en.cppreference.com/w/cpp/types/conditional) is used to specify MassType from DataTypes (if rigidtypes then rigidmass, otherwise Real)
One other solution would be to specify a MassType in VecTypes/RigidTypes but seems invasive to me.
I let the double template thing, as I dont know if it is possible than a developer would like to specify special instanciation, like RigidType with Real for example. But this does not solve the double template thing in xml.~~

~~If it is not the case, then we could remove the second template, and just use a `typedef` (so we remove the need to specify the double template in xml....)~~

Template has been removed directly, and the MassType is SFINAE'ed from the DataTypes.
parse() in the 3 differents masses warns the user of the modification, and what to do.


alxbilger:
I fully agree with your approach to remove this template parameter. However, I can suggest you another strategy. Instead of adding a MassType directly into the DataTypes, you can define it outside.

```cpp
template<class DataTypes>
struct MassType{}
```

and using SFINAE, you define MassType for each DataType:

```cpp
template<class T>
struct MassType<StdRigidTypes<T> >
{
   using type = RigidMass<blabla>;
}
```
(not sure about the syntax)
And you do that for the other types also.

Then you could define `template<class TDataTypes, class TMassType = MassType<TDataTypes> > UniformMass`.

Like that you don't constraint the mass component to accept only 2 types of DataTypes. You are more generic.
What do you think?

SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/DiagonalMass.h
```diff
@@ -49,7 +50,8 @@ public :
     typedef sofa::defaulttype::StdVectorTypes< Vec3, Vec3, Real > GeometricalTypes ; /// assumes the geometry object type is 3D
 };
 
-template <class DataTypes, class TMassType>
+template <class DataTypes,
+          class TMassType = std::conditional_t<std::is_same_v<DataTypes, sofa::defaulttype::StdRigidTypes<DataTypes::spatial_dimensions, typename DataTypes::Real> >, defaulttype::RigidMass<DataTypes::spatial_dimensions, typename DataTypes::Real>, typename DataTypes::Real> >
```
hugtalbot: Fully agree with this very nice PR @fredroy 
Further to discussion together, we would suggest instead we would avoid having a second template and simply declare `TMassType` in the class:

``` cpp
using TMassType = std::conditional_t<std::is_same_v<DataTypes, sofa::defaulttype::StdRigidTypes<DataTypes::spatial_dimensions, typename DataTypes::Real> >, defaulttype::RigidMass<DataTypes::spatial_dimensions, typename DataTypes::Real>, typename DataTypes::Real> >
```

`TMassType` always depends on the `DataTypes`. It would make no sense according to me to keep this redundant template.

hugtalbot (changes requested):

fredroy:
> Like that you don't constraint the mass component to accept only 2 types of DataTypes. You are more generic.
> What do you think?

This is more elegant indeed, but I am just wondering if there will be more than VecTypes and RigidTypes ?

But in any case, I think like @hugtalbot Masstype must derive from DataTypes and that's it (aka removing the second template and use a typedef/using

alxbilger:
I also agree that MassType should by an alias inside the class.

> This is more elegant indeed, but I am just wondering if there will be more than VecTypes and RigidTypes ?

Cuda types?

fredroy:
> I also agree that MassType should by an alias inside the class.
> 
> > This is more elegant indeed, but I am just wondering if there will be more than VecTypes and RigidTypes ?
> 
> Cuda types?

Indeed, I thought CudaVectorTypes was inheriting VecTypes but seems not...

alxbilger:
Also, if you remove the second template parameter, it will break scenes. You can always fix them in the SOFA code base, but what about the users?

fredroy:
> Also, if you remove the second template parameter, it will break scenes. You can always fix them in the SOFA code base, but what about the users?

I was thinking tweaking the parse()/canCreate() methods to detect when the user has set the two templates (and then warn, etc)

alxbilger (changes requested):
The SFINAE mechanism is not very explicit (but I believe this is the best strategy in this situation). A user searching for `MassType` would end up on an empty struct. I suggest to add a description on the default `MassType` definition, saying that there are other definitions in other files.

I also propose a simplification of the `MassType` definition in a PR.

I also wonder if your message in the `parse` function is a good strategy. `parse` means that `canCreate` returns true. Is it possible if both templates are provided?

fredroy:
> I also wonder if your message in the `parse` function is a good strategy. `parse` means that `canCreate` returns true. Is it possible if both templates are provided?

Not 100% sure (ObjectFactory stuff is a mess) but canCreate() is this case will try all the possible templates and is testing against the template of the present MechanicalObject (ForceField::canCreate). So the template string of the mass itself is not really read, and it essentially depends on the MechanicalObject.
I already tried on a example with rigid and it worked well. (ArticulationSystemMapping.scn)

fredroy:
[ci-build][with-all-tests]

alxbilger (changes requested):
Just a reminder that I would appreciate a description of the `MassType` struct and its mechanism in `SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/MassType.h`. Indeed, if you follow the `MassType` definition and you end up in an empty struct, it can bring some questions.

fredroy:
> Just a reminder that I would appreciate a description of the `MassType` struct and its mechanism in `SofaKernel/modules/SofaBaseMechanics/src/SofaBaseMechanics/MassType.h`. Indeed, if you follow the `MassType` definition and you end up in an empty struct, it can bring some questions.
> 

Thats a good note, I wanted to wait for the end of the design, but yes a documentation is necessary

fredroy:
[ci-build][with-all-tests]

PR #2585, opened 2022-01-11
[SofaCUDA] Add first version of CudaFastTetrahedralCorotationalForceField

epernod:
Add first version of the code compatible with CudaVec3f and CudaVec3d.
Performances are lower than CUDATetrahedronFEMForceField for the moment. Some profiling work will be done in future work. 

Add cpu-gpu comparison scene and benchmarks scenes on a beam.

this PR depends on changes done in FastTetrahedralCorotationalForceField suggested in PR #2569 (will need to rebase)

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2675, opened 2022-02-08
[SofaCUDA] Add matrix3 transpose method on device

epernod:
transpose method was missing for matrix3 on device.
Not sure it is the best implementation but it does the job.

Will fix PR #2585 

### DIFF
--- a/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaMath.h
+++ b/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaMath.h
@@ -626,6 +626,15 @@ public:
                      x.z * v.x.z + y.z * v.y.z + z.z * v.z.z );
     }
 
+    __device__ matrix3<real> transpose(matrix3<real> v)
+    {
+        matrix3<real> M;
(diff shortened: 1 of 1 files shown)


applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaMath.h
```diff
@@ -626,6 +626,15 @@ class /*__align__(4)*/ matrix3
                      x.z * v.x.z + y.z * v.y.z + z.z * v.z.z );
     }
 
+    __device__ matrix3<real> transpose(matrix3<real> v)
```
alxbilger: I would have passed `v` by const reference. Same rules apply on CUDA than in c++. I am surprised it is not done for the other functions.
I am sure it does not impact the performances though.

alxbilger (changes requested):
It seems you can use `matrix3::make` taking 9 parameters. I think it would avoid an instantiation and some copies.

alxbilger (approved):

alxbilger:
[ci-build]

PR #2678, opened 2022-02-08
[examples] An example showing a skybox

alxbilger:
I think skyboxes are a great tool for demos with better visual quality. That's why I wanted a demo showing how to make one.
I had more ambitions than that but it's rather difficult to work with OpenGL from the 90's in SOFA.
The limitation is that the skybox reacts to translations and zoom.

A scene is introduced and here is a capture:

https://user-images.githubusercontent.com/10572752/153029239-2b5fd241-173d-4314-9ba8-a4870d21f31e.mp4

In this PR, I also added a build "project" containing all the shaders.


CMakeLists.txt
```diff
@@ -259,6 +259,12 @@ if (SOFA_BUILD_ADD_SCENES)
     add_subdirectory(examples)
 endif()
 
+## SOFA shaders
+option(SOFA_BUILD_ADD_SHADERS "Add SOFA shaders as a project of the build." OFF)
+if (SOFA_BUILD_ADD_SHADERS)
+    add_subdirectory(share/shaders)
+endif()
+
```
guparan, line 267: Why the subproject?

alxbilger: Similar to `SOFA_BUILD_ADD_SCENES`, it makes the shader files available in the solution, so they can be found and indexed, in order to ease edition of those files.

PR #2706, opened 2022-02-18
[SofaMiscFem] FIX missing Strain Displacement matrix update in TriangularFEMForceField

### LINKED ISSUE #2704: No new strainDisplacementMatrix computation in TriangularFEMForceField
### Problem

**Description**
In [TriangularFEMForceField.inl](https://github.com/sofa-framework/sofa/blob/v21.12/modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl#L1022), in `computeForce()`, the `strainDisplacementMatrix` is only updated for the new displacement when the flag `_anisotropicMaterial` is true. Otherwise, the strainDisplacementMatrix that was computed during the initialization in the initLarge() is taken.
That leads to the issue that the following computations of strain and stress use an old version of the strainDisplacementMatrix that does not correspond to the current displacements
I think that in the computeForce-method, the method `computeStrainDisplacement(J, elementIndex, A, B, C)` should be called, no matter if the material is anisotropic or not.

**Steps to reproduce**
The `_anisotropicMaterial` flag must be set to false (which is the default). Then, if I print the strainDisplacementMatrix `J` in the computeForce-method, afterwards call `computeStrainDisplacement(J, elementIndex, A, B, C)` and print `J` again, the two outputs differ. That shows that an old, not updated version of the strainDisplacementMatrix is used in the following computations.

**Expected behavior**
The strainDisplacementMatrix should always be updated.

---------------------------------------------

### Environment

**Context**

- System: Windows 10
- Version of SOFA: v21.12 branch at commit [baa0c04] (latest commit)

TheGreatLudini:
In TriangularFEMForceField, the strainDisplacementMatrix was not re-computed for a new simulation step in the computeForce-method. Because of that, the strain computation used an old version of the strainDisplacementMatrix

EDIT: this will fix issue: 
fix #2704

### DIFF
--- a/modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
+++ b/modules/SofaMiscFem/src/SofaMiscFem/TriangularFEMForceField.inl
@@ -1008,10 +1008,7 @@ void TriangularFEMForceField<DataTypes>::computeForce(Displacement &F, Index ele
         Coord B = R_0_2 * (p[b]-p[a]);
         Coord C = R_0_2 * (p[c]-p[a]);
 
-        if (_anisotropicMaterial)
-            computeStrainDisplacement(J, elementIndex, A, B, C);
-        else
-            J = triangleInf[elementIndex].strainDisplacementMatrix;
+        computeStrainDisplacement(J, elementIndex, A, B, C);
         computeStrain(strain, J, D);
         computeStress(stress, triangleInf[elementIndex].materialMatrix, strain);
         computeStiffness(J,K,triangleInf[elementIndex].materialMatrix);


alxbilger (changes requested):
`triangleInf[elementIndex].strainDisplacementMatrix` is computed at init in `TriangularFEMForceField::initLarge` and `TriangularFEMForceField::initSmall`. An hypothesis is that it is an optimization based on the fact it does not change from a time step to another. Are there any case where it is the case? Small method?

I see a similar code in `computeStress`, where the use of the pre-computed strain displacement is in `(method == SMALL)`, in opposite to the `computeForce` method, where it is in the corotational method. I guess those two functions must be consistent. It is my requested change.

Camille-K:
Hello,

I just had a look in the code and the associated paper ([link](https://hal.inria.fr/file/index/docid/394480/filename/NPF05.pdf)).

@alxbilger, there are two `computeStress` methods in TriangularFEMForceField, with the following signatures :
```
void computeStress(type::Vec<3,Real> &stress, Index elementIndex);
```
and
```
void computeStress(type::Vec<3,Real> &stress, const MaterialStiffness &K, const type::Vec<3,Real> &strain);
```
Only the first one seems to use the same pattern as in `computeForce` (with the call to `computeStrainDisplacement` conditioned by `_anisotropicMaterial`), in case of small deformation, which is - as you pointed out - incoherent with what is in `computeForce`.
However this `computeStress` method with this signature is apparently used only in 
```
void computeStressAlongDirection(Real &stress_along_dir, Index elementIndex, const Coord &direction);
```
which itself is used only in both `computeStressAcrossDirection` methods, and (finally) these are apparently not called anywhere.
So, _bottomline_: even though the code in `computeStress` is incoherent with the code in `computeForce`, it's actually not used and - I guess - should be cleaned rather than made coherent.

Regarding the initial issue, what I understand from the paper is that the strain-displacement matrix **should be updated at each time step** in case of large deformations. It seems to be clearly stated in the introduction, and demonstrated later in the text (section 2.3). I have a doubt if this is necessary in case of small displacement (for me the paper focuses on large deformation), but even in this case, the current implementation is not coherent (the update should be made in case of large displacement and could be saved in case of small displacement).
I think it's worth highlighting that it was actually the case at first, but it seems to have been changed in this [commit](https://github.com/sofa-framework/sofa/commit/b7d68ef394f703957e7a00037f88d7abeed82d24).

alxbilger:
I confirm `computeStressAcrossDirection` is dead code.
@epernod do you want to remove this code in your cleaning PR? Or do I make a PR?

epernod:
Sorry I think  I inverted Small and Large approach in my explanation at the SOFA meeting. 
I checked my wip PRs on TriangularFEMForcefield (#2275) and I have the same fix awaiting: https://github.com/sofa-framework/sofa/pull/2275/files#diff-33034b71c7c012b3d436dd149856f2ea3c991e6ebcb78511599d465128b6f559L1019-R1019
So let's merge this one and I will rebase my PR.

@alxbilger I will remove ```computeStressAcrossDirection ``` (if not already done), in #2287 
@hugtalbot I'm not. Tearing mechanism should be computed outside of the FEM to not add complexity in the code.

epernod (approved):

TheGreatLudini:
Hi everyone, sorry for taking so long to further explain my PR. I fully agree with @Camille-K, the referenced paper states it clearly that the matrix should be updated at each time step.
Thanks for merging my PR.

PR #2742, opened 2022-03-02
[SofaGuiQt] Remove sec unit from GUI

hugtalbot:
Unit should not be specified in the GUI, nothing tell us that seconds are used.
This PR just remove the secs from the GUI.

### DIFF
--- a/modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
+++ b/modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
@@ -1639,7 +1639,7 @@ void RealGUI::eventNewTime()
     {
         double time = root->getTime();
         char buf[100];
-        sprintf ( buf, "Time: %.3g", time );
+        sprintf ( buf, "Time: %.3g,   Steps:  %o", time, frameCounter );
         timeLabel->setText ( buf );
     }
 #endif
@@ -2132,7 +2132,6 @@ void RealGUI::playpauseGUI ( bool startSimulation )
         SimulationStopEvent startEvt;
         root->propagateEvent(core::execparams::defaultInstance(), &startEvt);
         m_clockBeforeLastStep = 0;
-        frameCounter=0;
         timerStep->start(0);
         return;
     }
--- a/modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
+++ b/modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
@@ -429,7 +429,7 @@ RealGUI::RealGUI ( const char* viewername)
     fpsLabel->setMinimumSize ( fpsLabel->sizeHint() );
     fpsLabel->clear();
 
-    timeLabel = new QLabel ( "Time: 999.9999 s", statusBar() );
+    timeLabel = new QLabel ( "Time: 999.9999", statusBar() );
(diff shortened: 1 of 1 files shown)


modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
```diff
@@ -429,7 +429,7 @@ RealGUI::RealGUI ( const char* viewername)
     fpsLabel->setMinimumSize ( fpsLabel->sizeHint() );
     fpsLabel->clear();
 
-    timeLabel = new QLabel ( "Time: 999.9999 s", statusBar() );
+    timeLabel = new QLabel ( "Time: 999.9999", statusBar() );
```
epernod, line 432: Everywhere in SOFA Time and Timestep are without unit?
I understand but I find weird/confusing to have a value shown without unit. Maybe we should also add the number of iteration side to side. Or another info to tell user that Time shown ==  dt * nbstep.  (except if dt has been modified during the simu ...)

alxbilger: I agree that time is not necessarily in seconds. Simulations in geophysics have a dt in thousands of years. It's the same of others physical values: they don't have units, as long as all of them are consistent.
I also agree to show the number of iterations. I already needed it

hugtalbot: agreed thanks for the feedback

alxbilger:
@hugtalbot could you add a screenshot and relaunch the CI?

alxbilger (commented):
1. Is `iterationCounter` reset when you press the reset button?
2. What's the difference between `iterationCounter` and `frameCounter`?

epernod:
> 1. Is `iterationCounter` reset when you press the reset button?
> 
It is reset to 0 in method: void RealGUI::resetScene()

>     2. What's the difference between `iterationCounter` and `frameCounter`?

the only diff I see is that frameCounter is reset to 0 when you pause/play the scene. I assume it is just used as counter for the FPS calculation. You could remove this reset to 0 and use it instead of the new iterationCounter.

modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
```diff
@@ -1639,7 +1639,7 @@ void RealGUI::eventNewTime()
     {
         double time = root->getTime();
         char buf[100];
-        sprintf ( buf, "Time: %.3g s", time );
+        sprintf ( buf, "Time: %.3g,   Steps:  %o", time, frameCounter );
```
epernod: why in octal: %o  and not %i ?

epernod:
Good result: 

image

modules/SofaGuiQt/src/sofa/gui/qt/RealGUI.cpp
epernod: in fact it is wrong to use octal. steps 9 and 10 are missing.

epernod:
```suggestion
        sprintf ( buf, "Time: %.3g,   Steps:  %i", time, frameCounter );
```
in fact it is wrong to use octal. steps 9 and 10 are missing in the count.

epernod (changes requested):

PR #2717, opened 2022-02-22
[SofaNG] Setup LinearSolver

fredroy:
and lots of cleaning.

#1527

Based on master

Deprecates SofaBaseLinearSolver, SofaGeneralLinearSolver, SofaPreconditioner, SofaDenseSolver, SofaSparseSolver.

Re-organise Linear Solvers into 3 categories:
 - Direct
 - Iterative
 - Preconditioner

Move Implementations of MatrixAccessors along its API in SofaCore (could be the occasion to definitely ditch CRSMultiMatrixAccessor), as they are not components + the utility class BlocMatrixWriter
And cleaning (namespaces, pragma, explicite instanciations, symbol export for windows, SReal instead of double, etc)

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2783, opened 2022-03-14
[SofaNG] Setup MechanicalLoad

fredroy:
#1527 

based on:
- #2717  (for one file moved in core)

Deprecates partially _SofaBoundaryCondition_

Gathers all other forcefields which applies **external** force on mechanical objects. (torsion, compression, etc). https://mechanicalengineering.blog/types-of-loads/

I chose `MechanicalLoad` because of (https://en.wikipedia.org/wiki/Mechanical_load) and it seems an official term but I was hesitating also with `ExternalForce` (because these components set external forces on objects) 
Could have been simply `Load` as well, but there are other type of Load: electrical load for example.

EDIT: create a new sub module for FEM, called **FEM.Damping**, which gathers DampingFF


fredroy:
[ci-build][with-all-tests]

PR #2787, opened 2022-03-15
[Sofa.Type] Mat: Correct tensorProduct

olivier-goury:
The current implementation of the tensor product of two vectors seems incorrect since it assumes the resulting square matrix is always symmetric.
This is not true in general, for example: 
```
tensorProduct([0 1],[1 2]) = [0 1]'*[1 2] 
= [[0,0]
   [1 2]]
```

### DIFF
--- a/SofaKernel/modules/Sofa.Type/Sofa.Type_test/MatTypes_test.cpp
+++ b/SofaKernel/modules/Sofa.Type/Sofa.Type_test/MatTypes_test.cpp
@@ -203,3 +203,13 @@ TEST(MatTypesTest, invert55)
         EXPECT_EQ(M, Mtest);
     }
 }
+
+TEST(MatTypesTest, tensorProduct)
+{
+    Vec<2,SReal> v1(0.,1.), v2(1.,2.);
+    Mat<2, 2, SReal> Mtest = tensorProduct(v1,v2);
+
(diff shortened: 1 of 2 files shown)


fredroy:
Would it be possible to write a unit test to show this ? thanks ! 👍
(there are already some in Sofa.Type_test/MatTypes_test.cpp so it would be easy to add tensorproduct tests here)

alxbilger (changes requested):
Let's use this PR as an opportunity to improve this function:

I suggest to not initialize the matrix at construction:
```cpp
typedef MatNoInit<L,L,Real> Mat;
```

and take the vec parameters by const reference:

```cpp
template<sofa::Size L,class Real>
constexpr Mat<L,L,Real> tensorProduct(const Vec<L,Real>& a, const Vec<L,Real>& b ) noexcept
```
(the character '&' was missing)

olivier-goury:
Thanks for the feedback

olivier-goury:
I did the changes as requested. Let me know if that's OK!

alxbilger (approved):

epernod:
[ci-build][with-all-tests]

epernod (approved):

fredroy (approved):

PR #2762, opened 2022-03-07
[SofaSparseSolver] Apply fill in permutation in SparseLDLSolver

dessteph:
Based on https://github.com/sofa-framework/sofa/pull/2756

I added a boolean data, applyPermutation, to the SparseLDLSolver. It gives the choice to the user to apply or not a fill-in reducing permutation on the matrix of the system. By default the boolean is true in order to keep the previous behavior.

FEMBAR-SparseLDLSolver.scn average computation time (to solve the system) with permutation :  3.31 ms
FEMBAR-SparseLDLSolver.scn average computation time (to solve the system) without permutation :  2.05 ms
Benchmarck MatrixAssembly_direct_blocs.scn (with permutation) :  7,45 ms
Benchmarck MatrixAssembly_direct_blocs.scn (without permutation) : 114,39 ms

As the fill-in reducing permutation is computed following heuristic instructions, there may be some situation where it is not worth applying this permutation.

Asynchronous solver still works, not applying the permutation seems to slow it down.
______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2788, opened 2022-03-15
[SofaSparseSolver] SparseLUSolver and SparseCholeskySolver support fill reducing permutation

dessteph:
Based on https://github.com/sofa-framework/sofa/pull/2762

Based on SuiteSparse, SparseLUSolver and SparseCholeskySolver can now apply a fill reducing permutation via SuiteSpare or METIS.
The aim of this is to reduce the number of non null values in the factorization.
I added a data d_typePermutation that allows the choice of the type of permutation used, by default it is set to None.

SparseCholeskySolver is a little faster 1.59ms in avarage now against 1.84ms before (13.5% faster) on FEMBAR with the default behavior (no permutation applied).
The performances of SparseLUSolver seems to stay at the same level, or loose a little ( 0.07ms , 1% slower) on FEMBAR.
But with an adequate topology there is a significant rise of performances.
For MatrixAssembly_direct_blocs.scn :
-SparseLUSolver : 
     -METIS : 26.43ms on average
     -SuiteSparse : 30.76 ms on average.
against 469.64ms previously (without any fill reducing permutation). We have a speedup of 18 for METIS and of 16 for SuiteSparse.

-SparseCholskySolver:
   -METIS :  6.85ms
   -SuiteSparse : 5.71ms
against 107.48ms previously. There is a speedup of 15 for METIS and of 19 for SuiteSparse.

|   _   | without FRP | with FRP | Speed  |
|:-------------|:-------------:|:-------------:|:-------------:|
| SparseLUSolver<br/>+ METIS | 469.64 ms | 26.43 ms | x18
| SparseLUSolver<br/>+ SuiteSparse | 469.64 ms | 30.76 ms | x16
| SparseCholeskySolver<br/>+ METIS | 107.48 ms | 6.85 ms | x15
| SparseCholeskySolver<br/>+ SuiteSparse | 107.48 ms | 5.71 ms | x19

Related doc: https://github.com/sofa-framework/doc/pull/62


dessteph:
[ci-build][with-all-tests]

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
```diff
@@ -108,10 +110,117 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
     A.nz = -1;							// # of entries in triplet matrix, -1 for compressed-col
     cs_dropzeros( &A );
     tmp.resize(A.n);
-    S = cs_schol (&A, order) ;		/* ordering and symbolic analysis */
+    
+    perm = new int[A.n ];
+    iperm = new int[A.n ];
```
alxbilger: You need to free this allocated memory otherwise there is a memory leak.

I suggest your to use `sofa::type::vector` instead of a C-array.

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
```diff
@@ -108,10 +110,117 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+
+    fill_reducing_perm(A , perm, iperm); // compute the fill reducing permutation
```
alxbilger: It means you compute the permutation at each time step. I am not sure that it is cheap, so I suggest you to add a mechanism to do it only once.

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
```diff
@@ -108,10 +110,117 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+    sofa::helper::AdvancedTimer::stepEnd("factorization");
+}
+
+template<class TMatrix, class TVector>
+void SparseCholeskySolver<TMatrix,TVector>::fill_reducing_perm(cs A,int * perm,int * invperm)
+{
+    int n = A.n;
+    if(d_applyPermutation.getValue() )
+    {
+        int *M_colptr=A.p, *M_rowind=A.i ;
```
alxbilger: If this is the same algorithm in LDL, LU, and Cholesky, I suggest to factorize this code. Put this code into a free function, in a file that you would include in LDL, LU and Cholesky.

modules/SofaSparseSolver/src/SofaSparseSolver/SparseLUSolver.inl
```diff
@@ -85,12 +90,133 @@ void SparseLUSolver<TMatrix,TVector,TThreadManager>::invert(Matrix& M)
...
+            xadj[j] = M_colptr[j] - j ;
+
+            for(int ip = M_colptr[j]; ip<M_colptr[j+1];ip++)
+            {
+               int i = M_rowind[ip];
+               if(i !=j ) adj[it++] = i;
+            }
+        }
+        xadj[n] = M_colptr[n] - n;
+*/
```
alxbilger:
```suggestion
```
Please remove commented code

modules/SofaSparseSolver/src/SofaSparseSolver/SparseLUSolver.inl
```diff
@@ -85,12 +90,133 @@ void SparseLUSolver<TMatrix,TVector,TThreadManager>::invert(Matrix& M)
...
+
+ //Compute transpose in tran_colptr, tran_rowind, tran_values, tran_D
+            tran_countvec.clear();
```
alxbilger: Same remark than for Cholesky: if it's the same algorithm, factorize it

alxbilger (changes requested):
Good job! I have important suggestions.

Also, make sure that you did not slow down the solvers with your changes: benchmark the same behavior, before your changes, and after your changes.

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
dessteph: I will commit a new version handling permutations with vector<int> instead of int*.
The default destructor should free memory.

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
dessteph: It is not that expensive but still cost some computation. It was easier for SparseLDL that already used symbolic decomposition. It may need some work.

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
dessteph: I tough about doing exactly that but didn't know where in the inheritance I should put it.

modules/SofaSparseSolver/src/SofaSparseSolver/CSR_to_adj.h
```diff
@@ -0,0 +1,15 @@
...
+#include <SofaSparseSolver/config.h>
+#include <SofaBaseLinearSolver/MatrixLinearSolver.h>
+
+extern "C" {
+#include <metis.h>
+}
+
+namespace sofa::component::linearsolver
+{
+void CSR_to_adj(int n,int * M_colptr,int * M_rowind,type::vector<int> adj,type::vector<int> xadj); // compute the adjency matrix in CSR format from the matrix given in CSR format
```
alxbilger: Excellent! you have isolated this piece of code, now use it in the solvers.

```suggestion
void CSR_to_adj(int n, int * M_colptr, int * M_rowind, type::vector<int>& adj, type::vector<int>& xadj); // compute the adjency matrix in CSR format from the matrix given in CSR format
```

examples/Benchmark/Performance/MatrixAssembly/MatrixAssembly_direct_blocs.scn
```diff
@@ -27,7 +27,7 @@ The differences are in the way the global system matrix is built and solved:
             and provided to the linear solver.
             The template CompressedRowSparseMatrixMat3x3d indicates that the matrix is defined by 3x3 blocs.
          -->
-        <SparseLDLSolver template="CompressedRowSparseMatrixMat3x3d"/>
+        <SparseLUSolver template="CompressedRowSparseMatrixMat3x3d"/>
```
alxbilger: I am not sure this is change you wanted to push

modules/SofaSparseSolver/src/SofaSparseSolver/CSR_to_adj.h
```diff
@@ -0,0 +1,16 @@
...
+#include <SofaBaseLinearSolver/MatrixLinearSolver.h>
+
+extern "C" {
+#include <metis.h>
+}
+
+namespace sofa::component::linearsolver
+{
+void CSR_to_adj(int n,int * M_colptr,int * M_rowind,type::vector<int>& adj,type::vector<int>& xadj,type::vector<int>& t_adj, type::vector<int>& t_xadj, type::vector<int>& tran_countvec ); // compute the adjency matrix in CSR format from the matrix given in CSR format
+void CSR_to_adj(int n,int * M_colptr,int * M_rowind,type::vector<int>& adj,type::vector<int>& xadj);
```
alxbilger: Both functions content is very similar. Any way to factorize?

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
```diff
@@ -108,10 +109,67 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+void SparseCholeskySolver<TMatrix,TVector>::fill_reducing_perm(cs A,int * perm,int * invperm)
+{
+    int n = A.n;
+    if(d_applyPermutation.getValue() )
+    {   
+        sofa::type::vector<int> adj,xadj;
+
+        CSR_to_adj( A.n, A.p , A.i , adj, xadj);
+        
+        METIS_NodeND(&n, xadj.data(), adj.data(), NULL, NULL, perm,invperm);
```
alxbilger:
```suggestion
        METIS_NodeND(&n, xadj.data(), adj.data(), nullptr, nullptr, perm,invperm);
```
Prefer `nullptr` over `NULL`

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.h
```diff
@@ -31,6 +31,10 @@
 #include <sofa/helper/map.h>
 #include <cmath>
 #include <csparse.h>
+#include"CSR_to_adj.h"
```
alxbilger:
```suggestion
#include <SofaSparseSolver/CSR_to_adj.h>
```

modules/SofaSparseSolver/src/SofaSparseSolver/SparseLDLSolverImpl.h
```diff
@@ -198,64 +201,80 @@ protected :
...
-    void LDL_ordering(int n,int * M_colptr,int * M_rowind,int * perm,int * invperm) {
-        //Compute transpose in tran_colptr, tran_rowind, tran_values, tran_D
-        tran_countvec.clear();
-        tran_countvec.resize(n);
+    void LDL_ordering(int n,int * M_colptr,int * M_rowind,int * perm,int * invperm)
+    {
+        if( d_applyPermutation.getValue() )
+        {
+            CSR_to_adj( n, M_colptr, M_rowind, adj, xadj, t_adj, t_xadj, tran_countvec );
+            /*
```
alxbilger: Please remove commented code

modules/SofaSparseSolver/src/SofaSparseSolver/SparseLUSolver.h
```diff
@@ -68,13 +74,19 @@ class SparseLUSolver : public sofa::component::linearsolver::MatrixLinearSolver<
...
     void solve (Matrix& M, Vector& x, Vector& b) override;
     void invert(Matrix& M) override;
 
+    SparseLUSolver();
+
 protected :
 
+    Data<bool> d_applyPermutation ;
+  
+    void fill_reducing_perm(cs A,int * perm,int * invperm);
```
alxbilger:
```suggestion
    void fill_reducing_perm(cs A, int * perm, int * invperm);
```
cosmetic: just put a space after a coma

examples/Benchmark/Performance/MatrixAssembly/MatrixAssembly_direct_blocs.scn
dessteph: It was for a test, i forgot to leave it out

alxbilger (changes requested):

modules/SofaSparseSolver/src/SofaSparseSolver/CSR_to_adj.h
dessteph: By the way the LDL solver is implemented, it seems it needs to save the change on the variables used inside the fonction. Maybe by adding a dummy variable that we free after the execution for LU and Cholesky.

modules/SofaSparseSolver/src/SofaSparseSolver/CSR_to_adj.h
```diff
@@ -0,0 +1,15 @@
+#ifndef SOFA_CSR_TO_ADJ_H
```
alxbilger: you must add the header that is in every file of SOFA.

modules/SofaSparseSolver/src/SofaSparseSolver/CSR_to_adj.cpp
```diff
@@ -0,0 +1,59 @@
+#include <SofaSparseSolver/CSR_to_adj.h>
```
alxbilger: you must add the header that is in every file of SOFA.

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
```diff
@@ -108,10 +108,71 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
     A.nz = -1;							// # of entries in triplet matrix, -1 for compressed-col
     cs_dropzeros( &A );
     tmp.resize(A.n);
-    S = cs_schol (&A, order) ;		/* ordering and symbolic analysis */
+    
+    perm.resize(A.n);
+    iperm.resize(A.n);
+
+    if(computePermutation)
```
alxbilger: Are there any cases where the permutation must be updated? For example, when the matrix size changes, when the matrix structure changes. If yes, it needs to be checked.

modules/SofaSparseSolver/src/SofaSparseSolver/SparseCholeskySolver.cpp
```diff
@@ -108,10 +108,71 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+    }
+    S = symbolic_Chol( &A ); // symbolic analysis
+    permuted_A = cs_permute( &(A), iperm.data() , perm.data() , 1);
+
+    sofa::helper::ScopedAdvancedTimer factorizationTimer("factorization");
     N = cs_chol (&A, S) ;		/* numeric Cholesky factorization */
 }
 
+template<class TMatrix, class TVector>
+void SparseCholeskySolver<TMatrix,TVector>::fill_reducing_perm(cs A,int * perm,int * invperm)
```
alxbilger:
```suggestion
void SparseCholeskySolver<TMatrix,TVector>::fill_reducing_perm(const cs& A,int * perm,int * invperm)
```
pass by reference to avoid a copy

modules/SofaSparseSolver/src/SofaSparseSolver/SparseLUSolver.inl
```diff
@@ -85,12 +91,66 @@ void SparseLUSolver<TMatrix,TVector,TThreadManager>::invert(Matrix& M)
...
+    invertData->perm.resize(invertData->A.n);
+    invertData->iperm.resize(invertData->A.n);
+
     invertData->tmp = (Real *) cs_malloc (invertData->A.n, sizeof (Real)) ;
-    invertData->S = cs_sqr (&invertData->A, order, 0) ;		/* ordering and symbolic analysis */
-    invertData->N = cs_lu (&invertData->A, invertData->S, f_tol.getValue()) ;		/* numeric LU factorization */
+
+    if( invertData->computePermutation){
+        fill_reducing_perm(invertData->A, invertData->perm.data(), invertData->iperm.data() ); // compute the fill reducing permutation
+        invertData->computePermutation = false;
```
alxbilger: Same remark than for Cholesky. Cases where permutation must be recomputed must be supported

modules/SofaSparseSolver/src/SofaSparseSolver/CSR_to_adj.h
alxbilger: Also, we tend to use the `#pragma once` directive instead of `#ifndef` guards

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/CSR_to_adj.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+    t_adj.resize(t_xadj[n]);
+    for (int j=0;j<n;j++) {
+        for (int i=M_colptr[j];i<M_colptr[j+1];i++) {
+            int line = M_rowind[i];
+            if (line>j) {
+                t_adj[t_xadj[line] + tran_countvec[line]] = j;
+                tran_countvec[line]++;
+        }
+    }
```
alxbilger: There is an indentation issue here

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -100,8 +140,78 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
     tmp.resize(A.n);
-    S = cs_schol (&A, order) ;		/* ordering and symbolic analysis */
-    N = cs_chol (&A, S) ;		/* numeric Cholesky factorization */
+
+    switch (type_perm.getValue().getSelectedId() )
+    {
+        case 0:
+        default://None->identity
+            {
+                sofa::helper::ScopedAdvancedTimer factorization_permTimer("factorization_perm");
```
alxbilger: Put this timer out to avoid repeating it

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -100,8 +140,78 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+            break;
+        case 2://METIS
+            perm.resize(A.n);
+            iperm.resize(A.n);
+            { 
+                sofa::helper::ScopedAdvancedTimer factorization_permTimer("factorization_perm");
+                if(computePermutation)
+                {
+                    fill_reducing_perm(A , perm.data(), iperm.data() ); // compute the fill reducing permutation
+                    computePermutation = false;
```
alxbilger: Same remark than earlier: you need to check if the structure changed to detect if you need to recompute the permutation.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -41,34 +47,69 @@ SparseCholeskySolver<TMatrix,TVector>::~SparseCholeskySolver()
...
+            break;
+
+        case 2://METIS
+
+            cs_ipvec (n, perm.data(),  b , tmp.data() );	//x = P*b , permutation on rowsl;
+            cs_lsolve (N->L, tmp.data() );			//x = L\x
+            cs_ltsolve (N->L, tmp.data() );			//x = L'\x/
+            cs_pvec (n, perm.data() , tmp.data() , x );	 //b = P'*x , permutation on columns
+            break;
+
```
alxbilger: Why SuiteSparse and METIS cases cannot be combined? You cannot use a common variable to store the permutation?

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLUSolver.h
```diff
@@ -67,13 +76,19 @@ class SparseLUSolver : public sofa::component::linearsolver::MatrixLinearSolver<
...
-    SparseLUSolver();
+    
     void solve (Matrix& M, Vector& x, Vector& b) override;
     void invert(Matrix& M) override;
 
+    SparseLUSolver();
+
 protected :
 
+    Data<sofa::helper::OptionsGroup> type_perm;
```
alxbilger: All Data must start with `d_`. See https://www.sofa-framework.org/community/doc/programming-with-sofa/guidelines/.
Please rename to `d_typePermutation`.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -30,7 +31,12 @@ template<class TMatrix, class TVector>
 SparseCholeskySolver<TMatrix,TVector>::SparseCholeskySolver()
     : f_verbose( initData(&f_verbose,false,"verbose","Dump system state at each iteration") )
     , S(nullptr), N(nullptr)
+    , type_perm(initData(&type_perm, "permutation", "Type of fill reducing permutation"))
 {
+    computePermutation = true;
```
alxbilger: Put the initialization of `type_perm` into the constructor initializer list (https://en.cppreference.com/w/cpp/language/constructor).

alxbilger (changes requested):

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
dessteph: Pinv is int* and perm is vector<int> , there are some memory issues.
Doable if I manage to resolve them.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLUSolver.h
alxbilger: Thanks for adding the `d_` prefix. You also must use lower camel case for variables names (see again the guidelines). In your case, `d_type_perm` must be written `d_typePerm`.  But I find clearer to have full names (permutation instead of perm). So I suggest `d_typePermutation`.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.h
```diff
@@ -47,22 +52,28 @@ class SparseCholeskySolver : public sofa::component::linearsolver::MatrixLinearS
...
+
+    SparseCholeskySolver();
+    ~SparseCholeskySolver();
+    void solve (Matrix& M, Vector& x, Vector& b) override;
+    void invert(Matrix& M) override;
+
     void solveT(double * z, double * r);
     void solveT(float * z, float * r);
+
+    void fill_reducing_perm(cs A,int * perm,int * invperm);
```
alxbilger:
```suggestion
    void fill_reducing_perm(const cs& A,int * perm,int * invperm);
```

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -30,7 +31,11 @@ template<class TMatrix, class TVector>
 SparseCholeskySolver<TMatrix,TVector>::SparseCholeskySolver()
     : f_verbose( initData(&f_verbose,false,"verbose","Dump system state at each iteration") )
     , S(nullptr), N(nullptr)
+    , d_type_perm(initData(&d_type_perm, "permutation", "Type of fill reducing permutation"))
+    , d_type_permOptions(3,"None", "SuiteSparse", "METIS")
```
alxbilger: There is no need to have 2 variables for the same thing here: d_type_perm contains the options group and the selected value, so there is no need for `d_type_permOptions`

alxbilger (changes requested):

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/CSR_to_adj.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
```
alxbilger: Don't  forget the headers that are in every file in SOFA

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/CSR_to_adj.h
```diff
@@ -0,0 +1,9 @@
...
+
+#include <sofa/component/linearsolver/iterative/MatrixLinearSolver.h>
+
+namespace sofa::component::linearsolver
+{
+void CSR_to_adj(int n,int * M_colptr,int * M_rowind,type::vector<int>& adj,type::vector<int>& xadj,type::vector<int>& t_adj, type::vector<int>& t_xadj, type::vector<int>& tran_countvec ); // compute the adjency matrix in CSR format from the matrix given in CSR format
```
alxbilger: Could you add a detailed comment to document this function?

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -41,34 +46,78 @@ SparseCholeskySolver<TMatrix,TVector>::~SparseCholeskySolver()
...
+    sofa::helper::ScopedAdvancedTimer solveTimer("solve");
+
+    switch( d_type_perm.getValue().getSelectedId() )
+    {
+        case 0://None->identity
+        default:
+            {  
+                cs_lsolve (N->L, (double*) b );			//x = L\x
+                cs_ltsolve (N->L, (double*) b );			//x = L'\x/
+                cs_pvec (n, S->Pinv, (double*) b , (double*) x );	 //used here to copy, transopse(Pinv) = Id
```
alxbilger: Why this copy here since `Pinv` is the identity?

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -100,8 +148,75 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+    int n = A.n;
+    sofa::type::vector<int> adj, xadj, t_adj, t_xadj, tran_countvec;
+    CSR_to_adj( A.n, A.p , A.i , adj, xadj, t_adj, t_xadj, tran_countvec );
+    METIS_NodeND(&n, xadj.data(), adj.data(), nullptr, nullptr, perm, invperm);
+
+}
+
+template<class TMatrix, class TVector>
+css* SparseCholeskySolver<TMatrix,TVector>::symbolic_Chol(cs *A)
+{ //based on cs_chol
```
alxbilger:
```suggestion
{ //based on cs_schol
```
I think you meant `cs_schol`

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -100,8 +148,75 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+            break;
+
+        case 2://METIS
+            perm.resize(A.n);
+            iperm.resize(A.n);
+            
+            fill_reducing_perm(A , iperm.data(), perm.data() ); // compute the fill reducing permutation
+    
+            permuted_A = cs_permute( &A , perm.data() , iperm.data() , 1);
+            S = symbolic_Chol( permuted_A ); // symbolic analysis
```
alxbilger: Can it be avoided in case of constant matrix structure?

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
dessteph: Pinv is in fact nullptr, the SuiteSparse function considers it as identity (no permutation applied).

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
dessteph: We can, but then it will be locked on that permutation it for all the simulation.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
dessteph: I did.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
alxbilger: Even if `Pinv` is `nullptr`, `cs_pvec` still performs a copy.
I think you can avoid this copy by providing directly `x` to `cs_ltsolve`

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
alxbilger: This is what we do in `SparseLDLSolver` right? Along with a mechanism to detect some changes in the matrix structure

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
dessteph: I'll try, x =b after solving the issue with the 2 OptionsGroup objects. I don't know how to call the constructor of the OptionsGroup after calling the constructor of the data.

alxbilger: See an example here: https://github.com/sofa-framework/sofa/blob/9c5e4b320b2eaaedcecf805e79e20598f72f769f/SofaKernel/modules/Sofa.GL/Component/Rendering3D/src/sofa/gl/component/rendering3d/OglSceneFrame.cpp#L42

dessteph: That is what I did. We build a new object. But then we have one data<OptionsGroup> and one OptionsGroup object.

alxbilger: Yes, but you defined a temporary object storing the options group as a class member. In the example I provided, it's just a local variable.

dessteph: As x and b are pointers given as parameters, i cannot change their value inside the function. I need to copy the pointed values or change the function to take pointers on pointers (which is not compatible with the 2 other cases).

alxbilger: Ok I got it: `b` is both input and output. So a copy must be performed either at the beginning in the solution vector, or as it is right now after the solve.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/CSR_to_adj.h
```diff
@@ -0,0 +1,13 @@
+#pragma once
+
+#include <sofa/component/linearsolver/iterative/MatrixLinearSolver.h>
+
+namespace sofa::component::linearsolver
+{
+void CSR_to_adj(int n,int * M_colptr,int * M_rowind,type::vector<int>& adj,type::vector<int>& xadj,type::vector<int>& t_adj, type::vector<int>& t_xadj, type::vector<int>& tran_countvec ); 
+/* compute the adjency matrix in CSR format from the matrix given in CSR format
```
alxbilger: Please put the comment above the function signature.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/CSR_to_adj.h
alxbilger: Another reminder

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/CSR_to_adj.h
alxbilger: Another reminder

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.h
```diff
@@ -47,24 +52,46 @@ class SparseCholeskySolver : public sofa::component::linearsolver::MatrixLinearS
...
+    void solveT(Vector& z, Vector& r);
 
-    void solveT(double * z, double * r);
-    void solveT(float * z, float * r);
+    void fill_reducing_perm(const cs &A,int * perm,int * invperm);
+    css* symbolic_Chol(cs *A);
 };
 
+
+inline bool need_symbolic_factorization(int s_M, int * M_colptr,int * M_rowind, int s_P, int * P_colptr,int * P_rowind) {
```
alxbilger: Must be factorized among all the direct sparse solvers

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -41,48 +47,72 @@ SparseCholeskySolver<TMatrix,TVector>::~SparseCholeskySolver()
...
-    cs_ipvec (n, S->Pinv, (double*) &(r_tmp[0]), (double*) &(tmp[0]));	//x = P*b
+    Previous_rowind.clear();
+    Previous_colptr.resize(A.n +1);
 
-    cs_lsolve (N->L, (double*) &(tmp[0]));			//x = L\x
-    cs_ltsolve (N->L, (double*) &(tmp[0]));			//x = L'\x/
 
-    cs_pvec (n, S->Pinv, (double*) &(tmp[0]), (double*) &(z_tmp[0]));	 //b = P'*x
+    // store the shape of the matrix
+    if ( need_factorization )
```
alxbilger: This step must be done in the factorization, not in the solve

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -100,9 +130,90 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
     A.nz = -1;							// # of entries in triplet matrix, -1 for compressed-col
     cs_dropzeros( &A );
     tmp.resize(A.n);
-    S = cs_schol (&A, order) ;		/* ordering and symbolic analysis */
-    N = cs_chol (&A, S) ;		/* numeric Cholesky factorization */
+    int order = -1;
+
+    sofa::helper::ScopedAdvancedTimer factorization_permTimer("factorization_perm");
+
+    need_factorization = need_symbolic_factorization( A.n , A.p , A.i, Previous_colptr.size()-1 , Previous_colptr.data() , Previous_rowind.data() );
```
alxbilger:
```suggestion
    need_factorization = need_symbolic_factorization( A.n , A.p , A.i, Previous_colptr.size() , Previous_colptr.data() , Previous_rowind.data() );
```

I removed the -1

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -100,9 +130,90 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+                fill_reducing_perm( A , iperm.data(), perm.data() ); // compute the fill reducing permutation
+            }
+            
+            permuted_A = cs_permute( &A , perm.data() , iperm.data() , 1);
+            
+            if (need_factorization) 
+            { 
+                if (S) cs_sfree(S);
+                S = symbolic_Chol( permuted_A ); 
+                } // symbolic analysis
```
alxbilger, line 164: Indentation issue

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLUSolver.h
```diff
@@ -67,19 +76,41 @@ class SparseLUSolver : public sofa::component::linearsolver::MatrixLinearSolver<
...
+    css* symbolic_LU(cs *A);
+
     MatrixInvertData * createInvertData() override {
         return new SparseLUInvertData<Real>();
     }
 
 };
 
+
+inline bool need_symbolic_factorization(int s_M, int * M_colptr,int * M_rowind, int s_P, int * P_colptr,int * P_rowind) {
```
alxbilger: Code factorization

alxbilger (changes requested):
Please make sure that everything is consistent between all 3 solvers, and factorize the code as much as possible.

I suggest to also have the 3 permutation options in SparseLDLSolver. But it could be done in a second PR.

Also, be consistent with the way you name your variables. Use always the same style. In SOFA, the guideline is to use camel lower case. The same guideline apply to functions.

Try to name the variable with the most meaningful name.

Don't forget to add the headers on the files you added.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
dessteph: We want to compare the size if the two matrix, Previous_colptr has a size of n+1. We need to keep that -1.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCommon.h
```diff
@@ -10,9 +32,8 @@ extern "C" {
 
 namespace sofa::component::linearsolver
 {
-void CSR_to_adj(int n,int * M_colptr,int * M_rowind,type::vector<int>& adj,type::vector<int>& xadj,type::vector<int>& t_adj, type::vector<int>& t_xadj, type::vector<int>& tran_countvec ); 
 /** 
-compute the adjency matrix in CSR format from the matrix given in CSR format
+compute the adjency matrix in CSR format from the matrix given in CSR format, we assume that the given matrix is syùetric
```
alxbilger:
```suggestion
compute the adjency matrix in CSR format from the matrix given in CSR format, we assume that the given matrix is symmetric
```

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCommon.h
```diff
@@ -0,0 +1,68 @@
...
+xadj[i+1]-xadj[i] is the number of neighbors of the i-th node
+adj[xadj[i]] is the first neighbor of the i-th node
+
+**/
+void csrToAdj(int n,int * M_colptr,int * M_rowind,type::vector<int>& adj,type::vector<int>& xadj,type::vector<int>& t_adj, type::vector<int>& t_xadj, type::vector<int>& tran_countvec ); 
+
+/// compute the fill reducing permutation via METIS
+void fillReducingPermutation(const cs &A,int * perm,int * invperm); 
+
+inline bool compareMatrixShape(int s_M, int * M_colptr,int * M_rowind, int s_P, int * P_colptr,int * P_rowind) {
```
alxbilger, line 51: The function definition can be moved into the cpp file

alxbilger: And a description of the function is appreciate. In particular, a description of the return value.

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolverImpl.h
```diff
@@ -386,13 +325,13 @@ protected :
     }
 
     type::vector<Real> Tmp;
-protected : //the folowing variables are used during the factorization they cannot be used in the main thread !
+protected : ///the folowing variables are used during the factorization they cannot be used in the main thread !
```
alxbilger:
```suggestion
protected : ///the following variables are used during the factorization they cannot be used in the main thread !
```

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLUSolver.inl
```diff
@@ -39,49 +40,173 @@ template<class TMatrix, class TVector,class TThreadManager>
...
+                cs_ipvec (n, invertData->N->Pinv, b.ptr(), invertData->tmp) ;	/// x = P*b 
+                cs_lsolve (invertData->N->L, invertData->tmp) ;		/// x = L\x 
+                cs_usolve (invertData->N->U, invertData->tmp) ;		/// x = U\x 
+                cs_ipvec (n, invertData->S->Q, invertData->tmp, x.ptr()) ;	/// x = Q*x fill reducing permutation on columns only
+                break;
+            }
+            
+            case 2:///METIS
+            {
+                ///cs_ipvec (n, invertData->N->Pinv , invertData->tmp, invertData->tmp); /// partial pivot, not used
```
alxbilger:
```suggestion
```
Remove commented code

alxbilger (changes requested):

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCommon.h
dessteph: I added a description but it is an inline function, i cannot move it in the .ccp and converting it slow down the execution

alxbilger:
[ci-build][with-all-tests]

dessteph:
[ci-build][with-all-tests]

fredroy:

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.inl
```diff
@@ -100,9 +112,99 @@ void SparseCholeskySolver<TMatrix,TVector>::invert(Matrix& M)
...
+        switch (d_typePermutation.getValue().getSelectedId() )
+        {
+            case 0:
+            default:// None->identity 
+                if ( notSameShape)
+                {
+                    if (S) cs_sfree(S);
+                    S = cs_schol (&A, -1) ;// SuiteSparse does not compute permutation 
+                }
+                    N = cs_chol (&A, S) ;		// numeric Cholesky factorization
```
fredroy: indentation problem

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCommon.cpp
```diff
@@ -0,0 +1,99 @@
...
+void fillReducingPermutation(const cs &A,int * perm,int * invperm)
+{
+    int n = A.n;
+    sofa::type::vector<int> adj, xadj, t_adj, t_xadj, tran_countvec;
+    csrToAdj( A.n, A.p , A.i , adj, xadj, t_adj, t_xadj, tran_countvec );
+    METIS_NodeND(&n, xadj.data(), adj.data(), nullptr, nullptr, perm, invperm);
+
+}
+
+}
```
fredroy, line 99:
```suggestion
} // namespace sofa::component::linearsolver::direct
```

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCommon.h
```diff
@@ -0,0 +1,67 @@
...
+compute the adjency matrix in CSR format from the matrix given in CSR format, we assume that the given matrix is symmetric
+
+M_colptr[i+1]-M_colptr[i] is the number of non null values on the i-th line of the matrix
+M_rowind[M_colptr[i]] to M_rowind[M_colptr[i+1]] is the list of the indices of the columns containing a non null value on the i-th line
+
+xadj[i+1]-xadj[i] is the number of neighbors of the i-th node
+adj[xadj[i]] is the first neighbor of the i-th node
+
+**/
+void csrToAdj(int n,int * M_colptr,int * M_rowind,type::vector<int>& adj,type::vector<int>& xadj,type::vector<int>& t_adj, type::vector<int>& t_xadj, type::vector<int>& tran_countvec );
```
alxbilger: It is recommended to put a space just after a comma

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCommon.h
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    for (int i=0;i<M_colptr[s_M];i++) {
+        if (M_rowind[i]!=P_rowind[i]) return true;
+    }
+
+    return false;
+}
+
+}
```
fredroy:
```suggestion
} // namespace sofa::component::linearsolver::direct
```

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCommon.h
```diff
@@ -0,0 +1,67 @@
...
+
+#include <sofa/component/linearsolver/direct/config.h>
+#include <sofa/component/linearsolver/iterative/MatrixLinearSolver.h>
+#include <csparse.h>
+
+extern "C" {
+#include <metis.h>
+}
+
+namespace sofa::component::linearsolver
```
fredroy:
```suggestion
namespace sofa::component::linearsolver::direct
```

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCommon.cpp
```diff
@@ -0,0 +1,99 @@
...
+*******************************************************************************
+* Authors: The SOFA Team and external contributors (see Authors.txt)          *
+*                                                                             *
+* Contact information: contact@sofa-framework.org                             *
+******************************************************************************/
+
+
+#include <sofa/component/linearsolver/direct/SparseCommon.h>
+
+namespace sofa::component::linearsolver
```
fredroy:
```suggestion
namespace sofa::component::linearsolver::::direct
```

fredroy:
My changes have been done, so I put a OK to this PR

fredroy (approved):

PR #2790, opened 2022-03-15
[SofaNG] Setup Constraint

fredroy:
#1527 

based on:
- #2717 

contains:
- #2752 (UncoupledConstraintCorrection needs UniformMass)

Deprecates partially SofaGeneralObjectInteraction, SofaConstraint and SofaBoundaryCondition.

Module containing two types of Constraint in SOFA:
 - **Projective** constraint
 - **Lagrangian** constraint ; this submodule is itself subdived in three (four?) categories: **Model**, **Correction**, and **Solver**. (still debate on a fourth one with lagrangian-related constraint animation loop)

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2635, opened 2022-01-25
[SofaNG] Setup Mapping 

fredroy:
2nd big guy

#1527
based on 
- ~~#2612 ~~
- #2717 (for a moved file in core)

Deprecates **SofaMiscMapping**
Partially deprecates **SofaBaseMechanics**, **SofaRigid**, **SofaGeneralRigid**

This module gathers all "mechanical" mappings (inhering BaseMapping) and __not__ topological mapping.
~~I am a bit concerned by the size of this module but I cannot really find any good sub-modules to sort these components into ?~~

Following a discussion with @ChristianDuriez @Younesssss @alxbilger , Mappings are organized between:
 - NonLinear: mappings implementing getK() and applyDJT()
 - Linear: the rest of mappings
 - MappedMatrix: components which are not mapping? but helps to map matrices (stiffness and stuff) 
      -> Ad-hoc module which is only used to patch up matrices in SOFA, will disappear once matrix assembly will be pushed

~~TODO (in an other PR), set Articulated-related things into a dedicated plugin.~~ Done in #2684 

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2797, opened 2022-03-21
[SofaNG] Setup AnimationLoop

fredroy:
#1527 

based on:
- #2790

Deprecates partially **SofaConstraint**
Deprecates **SofaGeneralAnimationLoop**

Module with all the non-default animation loop.
DefaultAL stays in Core to allow scene to run even if one did not build this module (and its deps).

### DIFF
--- a/Component/CMakeLists.txt
+++ b/Component/CMakeLists.txt
@@ -8,7 +8,7 @@ endif()
 
 set(SOFACOMPONENT_SOURCE_DIR "src/sofa/component")
 
-set(SOFACOMPONENT_DIRS ODESolver IO Playback SceneUtility Topology Visual LinearSolver Mass Diffusion Mapping SolidMechanics StateContainer Constraint) # AnimationLoop Mapping Collision Engine UI Controller) # Device
+set(SOFACOMPONENT_DIRS ODESolver IO Playback SceneUtility Topology Visual LinearSolver Mass Diffusion Mapping SolidMechanics StateContainer Constraint AnimationLoop) # Collision Engine UI Controller) # Device
 set(SOFACOMPONENT_TARGETS)
 foreach(component_dir ${SOFACOMPONENT_DIRS})
 	sofang_add_component_subdirectory(${component_dir} ${PROJECT_NAME}.${component_dir})
--- a/Component/AnimationLoop/CMakeLists.txt
+++ b/Component/AnimationLoop/CMakeLists.txt
@@ -1,0 +1,37 @@
+cmake_minimum_required(VERSION 3.12)
+project(Sofa.Component.AnimationLoop LANGUAGES CXX)
+
+set(SOFACOMPONENTANIMATIONLOOP_SOURCE_DIR "src/sofa/component/animationloop")
+
+set(HEADER_FILES
+    ${SOFACOMPONENTANIMATIONLOOP_SOURCE_DIR}/config.h.in
+    ${SOFACOMPONENTANIMATIONLOOP_SOURCE_DIR}/FreeMotionAnimationLoop.h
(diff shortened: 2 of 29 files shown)


fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):

fredroy:
> Just a question @fredroy I guess the DefaultAL will be brought back later on? I thought it could be done here as well
> 
> Else 👍

This is an interesting question; DefaultAnimationLoop is in SofaSimulationCore (in the framework) and if it moves, it will be mandatory to compile S.C.AnimationLoop then.
In my mind, there should be always an AnimationLoop available (even with the skinniest set of cmake options) but I understand why you would move it in this module.

Component/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ endif()
 
 set(SOFACOMPONENT_SOURCE_DIR "src/sofa/component")
 
-set(SOFACOMPONENT_DIRS ODESolver IO Playback SceneUtility Topology Visual LinearSolver Mass Diffusion Mapping SolidMechanics StateContainer Constraint) # AnimationLoop Mapping Collision Engine UI Controller) # Device
+set(SOFACOMPONENT_DIRS ODESolver IO Playback SceneUtility Topology Visual LinearSolver Mass Diffusion Mapping SolidMechanics StateContainer Constraint AnimationLoop) # Collision Engine UI Controller) # Device
```
epernod, line 11: is the comment needed?

modules/SofaGeneralAnimationLoop/src/SofaGeneralAnimationLoop/initSofaGeneralAnimationLoop.cpp
```diff
@@ -34,9 +34,10 @@ void initSofaGeneralAnimationLoop()
     static bool first = true;
     if (first)
     {
-        // msg_deprecated("SofaGeneralAnimationLoop") << "SofaGeneralAnimationLoop is deprecated. It will be removed at v23.06. You may use Sofa.Component.Mapping.MappedMatrix instead.";
+        // msg_deprecated("SofaGeneralAnimationLoop") << "SofaGeneralAnimationLoop is deprecated. It will be removed at v23.06. You may use Sofa.Component.Mapping.MappedMatrix and Sofa.Component.AnimationLoop instead.";
```
epernod, line 37: MappedMatrix?

Component/CMakeLists.txt
fredroy: on the right ? it is just a kind of roadmap; the goal is to have nothing at the end

modules/SofaGeneralAnimationLoop/src/SofaGeneralAnimationLoop/initSofaGeneralAnimationLoop.cpp
fredroy: https://github.com/sofa-framework/sofa/pull/2635

```
MappedMatrix: components which are not mapping? but helps to map matrices (stiffness and stuff)
-> Ad-hoc module which is only used to patch up matrices in SOFA, will disappear once matrix assembly will be pushed
```

epernod: just to be sure I got it. This msg_deprecated will be uncommented for the 22.12, and the compat layer will be removed in 23.06 ?

fredroy: I will uncomment the msg_warning before the v22.06 release

epernod (approved):

PR #2763, opened 2022-03-07
[Sofa.Geometry][SofaMeshCollision] Deprecate RayTriangleIntersection class

fredroy:
RayTriangleIntersection is a weird class, used to only compute the intersection of a ray and a triangle (with even a version with TriangleCollisionModel?).
So ;
 - I moved the useful function directly in Sofa.Geometry::Triangle (+a version w/o barycoeff )
 - replace the usage where RayTriangleIntersection  was used
 - inline it (no compilation unit anymore for it)
 - and deprecate it for people still using it 
 - write a small unittest as well

I took the liberty to generalize the normal() function in geometry::Triangle, as it would provoke a compilation error if used with something else than Vec3 (and extend the unittest with array type as well)

Diff: https://github.com/fredroy/sofa/compare/remove_triplane_intersection_file...move_triangleoctree

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2805, opened 2022-03-22
[SofaGeneralCollisionMesh] Move TriangleOctree utility class to Helper

fredroy:
TriangleOctree is used in some case where people wants to do ray tracing on an octree structure (filled with triangles)
It makes more sense to move it in helper, as it is not a component but a shared utility class potentially used by something else than "GeneralCollisionMesh" things (whatever that means)

Based on
 - #2763 
 
Breaking 'cause I removed dependency on VisualTool (Sofa.Core) to use directly Drawtool. Still, it has a compat file to redirect deprecated usages.
Removed more useless headers...
This class would need a quite heavy facelift BTW

PS: There was a weird include of defaulttype/fwd.h in DrawTool which was undetected until now, 🤥 ??


fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):

PR #2811, opened 2022-03-25
[SofaCore] Name dynamically allocated vec ids

alxbilger:
In the scene `examples/Benchmark/Performance/TorusFall.scn`, vectors are dynamically allocated by EulerImplicitSolver and GenericConstraintSolver. They now have a name instead of a meaning less number.

image

The names are not associated to the VecIds, but to the Data associated to the VecIds.

Note: CGLinearSolver allocates dynamically temporary vectors. I did not add any complexity for having names for them, as they are "freed" after each solve. But they are actually not freed, just unset. I suspect its a source of the bug we already met when there are many vecids.


fredroy:
[ci-build][with-all-tests]

damienmarchal:
I suggest pecifying the "group" of the data so thay are not in the default group  "vector" but appears in their own panel.  
Some suggestion for naming (i have no real idea on what is the best one): 
 - "Constraint Vectors"
 - "Free Vectors"
 - "Free States" 
 - "Solver's properties"
 
 Actually, the vectors should also be grouped into something like "States"

alxbilger:
@damienmarchal Excellent suggestion, thanks

alxbilger:
@damienmarchal following your suggestion:

image
image
image

PR #2822, opened 2022-03-31
[SofaCore] Remove assert in case of multiple init

alxbilger:
The assert triggers in the following case:
A SphereROI is defined in a Python script. In order for the outputs to be used right away, it must be initialize by the Python script, just after its definition.
But when the scene is completely defined, a visitor calls again the init method.

So the assert triggers in case the addInput method is called multiple times (like in the example I gave).


fredroy:
Bold of you to set "fast merge" on a PR about DDGNode (and its update mechanism) 🤠

alxbilger:
@fredroy That was to scare @damienmarchal :satisfied:

alxbilger:
Closed in favor of https://github.com/sofa-framework/sofa/pull/2825/files

PR #2377, opened 2021-10-01
[SofaBaseMechanics] Fix UniformMass topology changes handling mode.

epernod:
This PR depends on #2374 #2375 and #2376 and should be rebased before review.

- Remove old method ```handleTopologyChange```
- Turn d_indices into TopologySubsetIndices, this automatically fix the topological changes update. 
- Fix the use of subset indices combined with topology changes.
- Add callback on endingEvent to update the totalMass/vertexMass
- Fix the center of gravity display when using subset of indices

Here is an example of scenes RemovingTrianglesProcess.scn using a subset of indices (yellow points)
```<UniformMass template="Vec3d,double" name="mass" totalMass="1.0" handleTopologicalChanges="1" indices="0 1 2 3 4 5 6 7 8 9"/>```

RemovingTrianglesProcess_00000001

RemovingTrianglesProcess_00000002

______________________________
________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2853, opened 2022-04-08
[Mass] Fix point topological changes for UniformMass

hugtalbot:
Based on #2870 

The topological callback was actually not working for the UniformMass.
Now POINTSADDED and POINTSREMOVED are handled

in addition:
- add override to avoid massive warnings
- add box in scene to allow seeing the mesh

Thanks to @oparras for rising this in #2668
A regression could be added on this scene

### DIFF
--- a/Sofa/Component/Mass/src/sofa/component/mass/UniformMass.inl
+++ b/Sofa/Component/Mass/src/sofa/component/mass/UniformMass.inl
@@ -179,7 +179,28 @@ void UniformMass<DataTypes>::initDefaultImpl()
         msg_info() << "Topology path used: '" << l_topology.getLinkedPath() << "'";
 
         d_indices.createTopologyHandler(meshTopology);
-        d_indices.addTopologyEventCallBack(sofa::core::topology::TopologyChangeType::ENDING_EVENT, [this](const core::topology::TopologyChange* eventTopo) {
+        /// Computations to be done when points are added
+        d_indices.addTopologyEventCallBack(sofa::core::topology::TopologyChangeType::POINTSADDED, [this](const core::topology::TopologyChange* eventTopo) {
+            WriteAccessor<Data<SetIndexArray > > indices = d_indices;
+            const core::topology::PointsAdded* pointsAdded = static_cast<const core::topology::PointsAdded*>(eventTopo);
+            auto& addedIndices = pointsAdded->getIndexArray();
+
+            for (unsigned i = 0; i<addedIndices.size(); i++)
+            {
+                indices.push_back(addedIndices[i]);
+            }
+            updateMassOnResize(d_indices.getValue().size());
+        });
+        /// Computations to be done when points are removed
+        d_indices.addTopologyEventCallBack(sofa::core::topology::TopologyChangeType::POINTSREMOVED, [this](const core::topology::TopologyChange* eventTopo) {
+            WriteAccessor<Data<SetIndexArray > > indices = d_indices;
+            const core::topology::PointsRemoved* pointsRemoved = static_cast<const core::topology::PointsRemoved*>(eventTopo);
+            auto& removedIndices = pointsRemoved->getArray();
+
+            for (unsigned i = 0; i<removedIndices.size(); i++)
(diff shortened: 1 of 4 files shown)


Component/Mass/src/sofa/component/mass/UniformMass.inl
```diff
@@ -179,7 +179,28 @@ void UniformMass<DataTypes>::initDefaultImpl()
...
+        /// Computations to be done when points are added
+        d_indices.addTopologyEventCallBack(sofa::core::topology::TopologyChangeType::POINTSADDED, [this](const core::topology::TopologyChange* eventTopo) {
+            WriteAccessor<Data<SetIndexArray > > indices = d_indices;
+            const core::topology::PointsAdded* pointsAdded = static_cast<const core::topology::PointsAdded*>(eventTopo);
+            auto& addedIndices = pointsAdded->getIndexArray();
+
+            for (unsigned i = 0; i<addedIndices.size(); i++)
+            {
+                indices.push_back(addedIndices[i]);
+            }
```
epernod: d_indices is already a subsetIndices, its resize on adding/removing event is already internally handled before calling the callback. So you are duplicating the work.
I'm quite surprised no scene-tests are crashing. Perhaps because all tests on topological changes are using Diagonal or MeshMatrixMass.
Normally UniformMass behavior was already fixe din PR: #2377
Do you confirm you experienced a bug of mass value? 

Looking at the issue pointed in the comment, it seems the scene is using only a MechanicalObject + ParticleSource (which create new dofs). There is no topology in the node so all those callback are not created.

epernod (changes requested):

Component/Mass/src/sofa/component/mass/UniformMass.inl
fredroy: I can confirm the bug (as I was with @hugtalbot for the investigation 🕵️‍♂️), and the test scene, SPHParticleSource.scn does have a PointSetTopologyContainer (and a modifier)

epernod:
Thank you for this PR, it highlights a more serious bug when adding elements in TopologySubsetData. A PR to fix the TopologySubsetData is coming, then we will update this PR to adapt UniformMass with the good callbacks.

Right now the callback on PointRemoved event proposed in this PR will crash as the TopologySubsetData already has an internal mechanism to update its size. So you perform the deletion twice.

epernod:
Update the FIX but now this PR depends on change done in PR #2870

Component/Mass/src/sofa/component/mass/UniformMass.inl
```diff
@@ -179,7 +179,22 @@ void UniformMass<DataTypes>::initDefaultImpl()
         msg_info() << "Topology path used: '" << l_topology.getLinkedPath() << "'";
 
         d_indices.createTopologyHandler(meshTopology);
-        d_indices.addTopologyEventCallBack(sofa::core::topology::TopologyChangeType::ENDING_EVENT, [this](const core::topology::TopologyChange* eventTopo) {
+        d_indices.supportNewElements(true);
+
+        // Need to create a call back to assign index of new point into the topologySubsetData. Deletion is automatically handle.
+        d_indices.setCreationCallback([this](Index dataIndex, Index& valueIndex,
```
hugtalbot: @epernod what about the destruction?

epernod:
> any update on the #2870 @epernod ?

@hugtalbot now that #2869 has been merged, you can review the #2870, then this one will be the following...

epernod:
[ci-build][with-all-tests]

epernod:
[ci-build][with-all-tests]

Component/Mass/src/sofa/component/mass/UniformMass.inl
hugtalbot: Deletion is automatically handle: creating a callback is possible though but to add an additional behavior on the default one

epernod (approved):

epernod:
@hugtalbot this one is ready for me. Could you double check and merge?

PR #2865, opened 2022-04-12
[SofaCUDA] Resurrect CudaTetrahedronTLEDForceField

alxbilger:
It was commented in CMakeLists.txt, so never compiled.

* cylinder_TLEDTetraSolution.scn: must not be in SOFA examples because it contains CudaTetrahedronTLEDForceField. Deleted
* CudaTetrahedronTLEDForceField.scn: it mentioned meshes that are not in the repository. Therefore the content of cylinder_TLEDTetraSolution.scn has been taken instead. However, the simulation compared the TLED solution to a reference solution and they are not matching. The reference solution is removed. It may indicate a problem, or not...

Note that there is no CPU version of this force field.


applications/plugins/SofaCUDA/CMakeLists.txt
```diff
@@ -64,8 +64,7 @@ set(HEADER_FILES
     sofa/gpu/cuda/CudaTetrahedralTensorMassForceField.inl
     sofa/gpu/cuda/CudaTetrahedronFEMForceField.h
     sofa/gpu/cuda/CudaTetrahedronFEMForceField.inl
-    #sofa/gpu/cuda/CudaTetrahedronTLEDForceField.h
-    #sofa/gpu/cuda/CudaTetrahedronTLEDForceField.inl
+    sofa/gpu/cuda/CudaTetrahedronTLEDForceField.h
```
epernod, line 67: any idea why it was commented? something in the git log maybe?

I'm quite sure I saw the cpu version somewhere, I'll give a look.

Could you add the same benchmark scenes as for the others tetraFEM:
SofaCUDA\scenes\benchmarks\TetrahedronFEMForceField_beam10x10x40_gpu.scn
SofaCUDA\scenes\benchmarks\TetrahedronFEMForceField_beam16x16x76_gpu.scn

alxbilger: No idea.
I added the scenes, but there pretty slow because of a tiny time step (explicit time scheme).

alxbilger: Haha it seems you @epernod  are guilty here: https://github.com/sofa-framework/sofa/commit/2e3849aac14d8619b61bf6627de9d7f5898d29c9

PR #2868, opened 2022-04-14
[SofaSphFluid] Fix ParticleSource topologicalChanges use new callbacks mechanism

epernod:
Old version using a specific TopologyHandler overriding callbacks will be totally remove in future PR. 
Use specific callbacks inside the TopologySubsetData


applications/plugins/SofaSphFluid/src/SofaSphFluid/ParticleSource.inl
```diff
@@ -83,8 +81,12 @@ void ParticleSource<DataTypes>::init()
     sofa::core::topology::BaseMeshTopology* _topology = this->getContext()->getMeshTopology();
     if (_topology != nullptr)
     {
-        m_pointHandler = new PSPointHandler(this, &m_lastparticles);
-        // m_lastparticles.createTopologyHandler(_topology, m_pointHandler);
+        m_lastparticles.createTopologyHandler(_topology);
+        m_lastparticles.setDestructionCallback([this](Index pointIndex, Index& val)
```
hugtalbot, line 85: As you know, I have no huge XP in topological changes, but is that normal not have ConstructionCallBack @epernod ?

epernod: yes because there is nothing to do at the creation. A callback is only needed if you need "more than the default constructor" of your Data .

Basic example would be for mass lumping: Let's say you have a TopologyData<Edge, SReal>  storing an mass value per edge of the topology. When an Edge is created, default constructor will add the element in the TopologyData array with the good mass value but you might need to add a callback inside you Mass component that will sum the mass value on the vertices.
Similarly when an Edge is deleted, by default the element destructor is called and removed from the TopologyData array. But you will need to set a destructionCallback that will remove this Edge weight from the vertices. 

I hope it is more clear, I swear... I will write the doc... when I have time :goberserk:

hugtalbot (approved):

PR #2875, opened 2022-04-15
[SofaMatrix] Eigen can be used to reorder a mesh in order to reduce fill-in

alxbilger:
The component FillReducingOrdering, introduced in https://github.com/sofa-framework/sofa/pull/2501, was using the METIS library to reorder the vertices. Now, it is able to use approximate minimum degree algorithm from the Eigen library.


fredroy:
[ci-build][with-all-tests]

alxbilger:
Because I know the SOFA community likes screenshots a lot:

image


Left image: the original matrix, without any permutation
Right image: permuted matrix

When the matrix is permuted, the lower matrix of the LU factorization has 80% less non-zero values than when the matrix is not permuted. Hence the speed up!

applications/plugins/SofaMatrix/src/SofaMatrix/FillReducingOrdering.inl
```diff
@@ -176,4 +221,24 @@ void FillReducingOrdering<DataTypes>::updateElements(
...
+template <class DataTypes>
+void FillReducingOrdering<DataTypes>::updateMesh()
+{
+    // Update topology
+    updateElements(l_topology->getHexahedra(), d_hexahedra);
+    updateElements(l_topology->getTetrahedra(), d_tetrahedra);
+
+    // Update mechanical state
+    auto pos = helper::getWriteOnlyAccessor(d_position);
+    pos.clear();
```
damienmarchal: I wonder if pos.resize(previousPos.size() then pos[0] is very different or equal in overhead?

applications/plugins/SofaMatrix/src/SofaMatrix/FillReducingOrdering.inl
```diff
@@ -23,13 +23,16 @@
...
+#include <Eigen/Sparse>
+
 namespace sofa::component::linearsolver
 {
 
 template <class DataTypes>
 FillReducingOrdering<DataTypes>::FillReducingOrdering()
     : l_mstate(initLink("mstate", "Mechanical state to reorder"))
     , l_topology(initLink("topology", "Topology to reorder"))
+    , d_orderingMethod(initData(&d_orderingMethod, sofa::helper::OptionsGroup(2, "Metis", "AMD"), "orderingMethod", "Ordering method. AMD means Approximate minimum degree."))
```
damienmarchal: If AMD means for something... is there a possible hint for the METIS one ?

applications/plugins/SofaMatrix/src/SofaMatrix/FillReducingOrdering.inl
```diff
@@ -41,6 +44,8 @@ FillReducingOrdering<DataTypes>::FillReducingOrdering()
     addOutput(&d_position);
     addOutput(&d_hexahedra);
     addOutput(&d_tetrahedra);
+
+    sofa::helper::getWriteAccessor(d_orderingMethod)->setSelectedItem(0);
```
damienmarchal, line 51: I wonder if there is no room for a PR to make this writing a bit more elegant (not in this PR).

applications/plugins/SofaMatrix/src/SofaMatrix/FillReducingOrdering.inl
alxbilger: Metis is the library

applications/plugins/SofaMatrix/src/SofaMatrix/FillReducingOrdering.inl
alxbilger: you are right. Usually, I `reserve` before calling the `push_back`'s, which is equivalent to your suggestion.

applications/plugins/SofaMatrix/src/SofaMatrix/FillReducingOrdering.inl
alxbilger: I've changed the options. Let me know what you think

applications/plugins/SofaMatrix/src/SofaMatrix/FillReducingOrdering.inl
```diff
@@ -23,13 +23,19 @@
 
 #include <SofaMatrix/FillReducingOrdering.h>
 
+#include <Eigen/Sparse>
```
guparan, line 26: Eigen dependency should be visible in CMakeLists.txt and Config.cmake.in

guparan (changes requested):

applications/plugins/SofaMatrix/src/SofaMatrix/FillReducingOrdering.inl
alxbilger: Done, even if Eigen was already a dependency of another module.

How can I test a Config.cmake.in script?

guparan: We don't have such test in our CI. 
Currently the best way is to generate binaries with the command `[ci-build][generate-binaries]` and then try to build a plugin on top of those binaries.

guparan:
[ci-build][with-all-tests][generate-binaries]

guparan (approved):

PR #2884, opened 2022-04-19
[Git] Clean project files

marques-bruno:
I switched to the dark side, running emacs as my IDE :smiling_imp:, and I was testing it on SOFA's codebase. (If you like coding & navigating through SOFA's codebase without having QtCreator transforming your laptop into an airplane, you should try it, it's great!) 
Anyway, I'd like to suggest adding some stuff to the .gitignore file:

* emacs auto-generates autosave files that have the following formats: `#*` and `~*` 
* When running emacs as an IDE, an emacs package called `projectile` creates a file called `.dir-locals.el`, containing project configuration, compilation, and running instructions for emacs in the project's root folder.
* Finally, Emacs uses `clangd` or `ccls` as its language servers. these programs store their index (LSP data for the project's codebase) in `.cache/` and `.ccls-cache/` respectively. clangd & ccls both require cmake to generate a `compile_commands.json` file, which we also don't want to be committed in the git repository.

I also noticed that SOFA doesn't have a `.clang-format` file. A clang-format file gives great hints to contributors, regarding the formating guidelines for SOFA. This .clang-format file respects the [coding guidelines described here](https://www.sofa-framework.org/community/doc/programming-with-sofa/guidelines/)

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,137 @@
+---
+Language:        Cpp
+# BasedOnStyle:  Microsoft
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveMacros: false
+AlignConsecutiveAssignments: false
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlines: Right
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllArgumentsOnNextLine: true
+AllowAllConstructorInitializersOnNextLine: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: Never
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: None
+AllowShortLambdasOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: Never
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: MultiLine
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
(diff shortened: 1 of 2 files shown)


alxbilger (approved):
Thanks for the clang-format file !

guparan:
Thanks @marques-bruno !
Could you confirm that your clang-format file complies with [the astyle configuration](https://github.com/sofa-framework/sofa/blob/master/tools/astyle/sofa.astyle)?

marques-bruno:
> Thanks @marques-bruno ! Could you confirm that your clang-format file complies with [the astyle configuration](https://github.com/sofa-framework/sofa/blob/master/tools/astyle/sofa.astyle)?

Almost... I went through each of the options in the style sheet to make sure, but:
* *--indent-labels*: this is not supported by clang-format. This astyle option indents C-style labels (goto labels) as if it were a normal instruction in its scope. clang-format offers an option `IndentGotoLabels` but it will either align it with the root of its scope (true,) or align them at the first column in the file. for more info on this option:
[StackOverflow post](https://stackoverflow.com/a/66511380/1376968). But to me, it seems like a bad idea to indent labels the way astyle does. clang-format's makes more sense to me.

* *--add-braces*: This one is more problematic: clang-format supports this option since its latest version (clang-format-15.0.0), BUT, here's what they say in the doc:

_**Warning**_

_**Setting this option to true could lead to incorrect code formatting due to clang-format’s lack of complete semantic information. As such, extra care should be taken to review code changes made by this option.**_

Linter tools, like clang-tidy, let you do that also...
Most importantly, clang-format's default package version on ubuntu is v10.0.0. v12 can be found in ubuntu's default package manager, but for v15 you gotta look for the package in external repos or compile it yourself. So I'd rather leave this option out, and keep it more as a general guideline in SOFA's documentation.

Other than that my last commit updated the file to adopt the other astyle options :+1:

guparan:
It's ok to me as long as it does not something directly opposed to the astyle config :+1:

PR #2894, opened 2022-04-22
[Sofa.LinearAlgebra] Compressed property in CRS matrix depends on tmp vector

alxbilger:
A CompressedRowSparseMatrix is compressed if the temporary vector of triplets has been converted into the compressed format. Therefore, the boolean `compressed` must be set depending on the size of the temporary vector, not the size of the compressed values.


alxbilger:
[ci-build][with-all-tests]

alxbilger:
After discussion with @JeremieA, this PR breaks the meaning of the variable `compressed`. The comment after this variable says `true if the additional storage is empty or has been transfered to the compressed data structure`. But @JeremieA claims the variable also says if there are zeros in the compressed values.

A suggestion was to use the refactored `CompressedRowSparseMatrix` class from https://github.com/InSimo/ISSofa, which brings more clarity on the meanings of the variables (among other important things).

alxbilger:
@hugtalbot The label is actually "postponed". The reason is I'm not convinced that the variable also states that there are zeros in the compressed values. I did not see anything about that property in the code. In any case, I do not need this PR to continue. So it can be closed.

PR #2856, opened 2022-04-11
[Sofa.GL.Component] OglModel: Uniformize floating point type

fredroy:
OglModel was : 
 - converting some data (vertices, normals) from (default) double to float, 'cause it was set to float for OpenGL
 - textures was hard-set on float  (and set as float for Ogl)
 - tangents on double (and set double for OGL)

so a weird mix of conversion or not.

So now everything is in `SReal` type (so double by default) and set as GL_DOUBLE for the GPU.

This should not have any consequence for the performance, as OpenGL will still use single float-point for calculations _a priori_
And some cleanings here and there as usual

Based on
 - #2855 

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #2907, opened 2022-04-29
[Sofa.Component] Fix compilation when FLOATING_POINT is set to float

fredroy:
(without the tests....)

Main problem is the linear solvers using CSPARSE, which only handles double.
Actually, SparseCholesky was using buffers of float with pointers of double (which obviously causes a crash).
SparseLU is more convenient, it tried to use SReal, and make the compilation failing directly.

There were at least two way to solve this problem:
 - disable in the CMakeLists directly (if floating_point is float)
 - do not compile the instance templated on float (effectivement does not compile the two solvers)

I chose the latter to keep them visible in the base code, but I totally agree with the former solution (but requires to access to the Cmake variable SOFA_FLOATING_POINT from Sofa.Config in the CMakefile of S.C.LS.Direct)

An other thing is that preconditionner wants to use SparseCholesky  as well so there is a conditional compilation happening here unfortunately. I dont know if it is possible to just use an other direct solver (probably)

This was needed to test the performance tests for 
 - #2856


Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.cpp
```diff
@@ -28,10 +28,18 @@ namespace sofa::component::linearsolver::direct
 
 using namespace sofa::linearalgebra;
 
-int SparseCholeskySolverClass = core::RegisterObject("Direct linear solver based on Sparse Cholesky factorization, implemented with the CSPARSE library")
-        .add< SparseCholeskySolver< CompressedRowSparseMatrix<SReal>,FullVector<SReal> > >()
-        ;
+// dont try to compile if floating point is float
+#ifdef SOFA_FLOAT
+SOFA_PRAGMA_WARNING("SparseCholeskySolver does support float as scalar.")
```
alxbilger:
```suggestion
SOFA_PRAGMA_WARNING("SparseCholeskySolver does not support float as scalar.")
```

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseCholeskySolver.cpp
```diff
@@ -28,10 +28,18 @@ namespace sofa::component::linearsolver::direct
 
 using namespace sofa::linearalgebra;
 
-int SparseCholeskySolverClass = core::RegisterObject("Direct linear solver based on Sparse Cholesky factorization, implemented with the CSPARSE library")
-        .add< SparseCholeskySolver< CompressedRowSparseMatrix<SReal>,FullVector<SReal> > >()
-        ;
+// dont try to compile if floating point is float
```
alxbilger:
```suggestion
```

Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLUSolver.cpp
```diff
@@ -29,13 +29,18 @@ namespace sofa::component::linearsolver::direct
 
 using namespace sofa::linearalgebra;
 
+// dont try to compile if floating point is float
+#ifdef SOFA_FLOAT
+SOFA_PRAGMA_WARNING("SparseLUSolver does support float as scalar.")
```
alxbilger:
```suggestion
SOFA_PRAGMA_WARNING("SparseLUSolver does not support float as scalar.")
```

alxbilger (changes requested):
Thanks for this unfunny work. Only minor changes requested

hugtalbot (approved):

fredroy:

alxbilger (approved):
- At some point, we will need to ask the question if the [Eigen solvers](https://github.com/alxbilger/EigenLinearSolvers) can replace the ones based on Csparse. If yes, the Eigen solvers are able to deal with `float`.
- PrecomputedLinearSolver uses the Cholesky factorization because it's the most generic solver. Using another is possible but it restricts the use of this solver. (In practice, we all use SparseLDLSolver, so it would not be a problem).

PR #2908, opened 2022-04-29
[SolidMechanics] TetrahedronFEMForceField: adds error message for Poisson's ratio 

EulalieCoevoet:
Poisson's ratio should be in [0,0.5[. This PR adds an error message if it's not the case. 
A value outside [0,0.5[ just leads to an odd behavior. 

**EDIT**: a warning message (not error)


Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/TetrahedronFEMForceField.inl
```diff
@@ -1349,6 +1349,12 @@ void TetrahedronFEMForceField<DataTypes>::init()
         if (youngModulus[i]>maxYoung) maxYoung=youngModulus[i];
     }
 
+    const Real& poissonRatio = _poissonRatio.getValue();
+    if (poissonRatio<0 || poissonRatio>=0.5)
+    {
+        msg_error() << "FEM Poisson Ratio in Hooke's law should be in [0,0.5[.";
```
fredroy: I guess something has to be done if the user encounters this case.

- Either setting the component to invalid (with componentstate) 
- or forcefully clamp the value between [0,0.5[ (but you are purposely modifying its simulation so maybe not the best solution)

alxbilger: In theory, negative Poisson's ratios are possible. But I don't know if it's possible in this component

Refs:

- https://en.wikipedia.org/wiki/Auxetics
- https://en.wikipedia.org/wiki/Poisson%27s_ratio

fredroy: I will let other people (more versed in mechanics than me ) to answer 😏
@hugtalbot maybe ?

EulalieCoevoet: FYI I've just copied the data's info. I don't think we should use a negative ratio with Hooke's law. https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.661.2128&rep=rep1&type=pdf

```cpp
Data<Real> _poissonRatio; ///< FEM Poisson Ratio [0,0.5[
```
```cpp
_poissonRatio(initData(&_poissonRatio,(Real)0.45f,"poissonRatio","FEM Poisson Ratio in Hooke's law [0,0.5["))
```

To be honest my first intention was to set the component to an invalid state but I got discouraged by the 45 methods to identify/update. In the current implementation adding a `return;` after the error message (which would set the state of the component to invalid) will just make the simulation crash, since it is not used at all elsewhere. 
I don't think that clamping the value is a good idea either, same as for setting the default value.

EulalieCoevoet: FYI we need at least a message because of this -> https://github.com/SofaDefrost/SoftRobots/discussions/130 (nan values if you set the ratio to 0.5).

Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/TetrahedronFEMForceField.inl
```diff
@@ -1349,6 +1349,12 @@ void TetrahedronFEMForceField<DataTypes>::init()
         if (youngModulus[i]>maxYoung) maxYoung=youngModulus[i];
     }
 
+    const Real& poissonRatio = _poissonRatio.getValue();
+    if (poissonRatio<0 || poissonRatio>=0.5)
+    {
+        msg_warning() << "FEM Poisson Ratio in Hooke's law should be in [0,0.5[.";
+    }
```
alxbilger:
```suggestion
    if (poissonRatio < 0 || poissonRatio >= 0.5)
    {
        msg_warning() << "FEM Poisson's Ratio in Hooke's law should be in [0,0.5[.";
    }
```

alxbilger:
_Suggestion_
There is also the option to limit the range of the Poisson's ratio in the UI:

image

You need to add `_poissonRatio.setWidget("poissonRatio");` in the constructor of the component.
TetrahedralCorotationalFEMForceField is the only component using this widget.

Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/TetrahedronFEMForceField.inl
hugtalbot: Changing the component state to Invalid would also make sense, right @EulalieCoevoet ?

damienmarchal: The two choice should be either:
- a warning + clamping the value to the closest boundary
- an error and setting the component to invalid

But please don't mix warning with invalid component state.

EulalieCoevoet:
poke @hugtalbot

alxbilger:
[ci-build][with-all-tests]

alxbilger:
Could you solve the conflicts?

PR #2960, opened 2022-05-18
[SofaGraphComponent] Move scene checking ability into its own library

fredroy:
Scene checkers were located in the SofaGraphComponent (even if they were not components themselves....)
So they needed to be moved out of it.
The first idea was to move them to Sofa.Simulation as they related to the simulation itself (or rather its initialization).
But scene checking could need dependencies on components themselves (in this case, it needs the contact manager from Sofa.Component.Collision.Response.Contact, and APIVersion from Sofa.Component.SceneUtility)
So finally a library was created into ~~applications/plugins, even if it is not really a plugin in the sofa-way of thinking (aka a set of components)~~ Sofa/plugins because it is not in Component or framework, but it could be placed in Sofa/ directly (where GUI is as well)

As usual, a compat layer has been added to warn devs of the move.


fredroy:
[ci-build][with-all-tests]

Sofa/plugins/SceneChecking/src/SceneChecking/init.cpp
```diff
@@ -0,0 +1,51 @@
...
+* Contact information: contact@sofa-framework.org                             *
+******************************************************************************/
+#include <SceneChecking/init.h>
+
+namespace sofa::scenechecking
+{
+    
+extern "C" {
+    SOFA_EXPORT_DYNAMIC_LIBRARY void initExternalModule();
+    SOFA_EXPORT_DYNAMIC_LIBRARY const char* getModuleName();
```
alxbilger: Following https://github.com/sofa-framework/sofa/pull/2948, you must also provide a version number

alxbilger (changes requested):
Just a version number is requested

Sofa/plugins/SceneChecking/src/SceneChecking/init.cpp
fredroy: 👍

alxbilger (approved):

PR #3001, opened 2022-06-01
[Mass] Remove DiagonalMass and replace with MeshMatrixMass (with lumping)

### LINKED ISSUE #2452: Merge DiagonalMass and MeshMatrixMass

hugtalbot:
This PR proposes to remove DiagonalMass and replace with MeshMatrixMass (with lumping)
Fix #2452 

One case remains to be covered: RigidTypes with non uniform vertex mass. A new PR is coming on this.


epernod:
[ci-build][with-all-tests]

Sofa/Component/Mass/src/sofa/component/mass/MeshMatrixMass.cpp
```diff
@@ -80,6 +80,7 @@ int MeshMatrixMassClass = core::RegisterObject("Define a specific mass for each
         .add< MeshMatrixMass<Vec1Types> >()
         .add< MeshMatrixMass<Vec1Types, Vec2Types> >()
         .add< MeshMatrixMass<Vec1Types, Vec3Types> >()
+        .addAlias("DiagonalMass")
```
alxbilger, line 83: I don't agree that `DiagonalMass` is an alias for `MeshMatrixMass`. Instead, I suggest to have a true `DiagonalMass` class, derived from `MeshMatrixMass` with the Data `isLumped` set to `true` in its constructor, and make it read-only. It also gives you the opportunity to deprecate it properly by warning the user. It is also possible to specialize the class for rigid types.

If you really want to remove totally the DiagonalMass, at least you need to check that `isLumped` is true when the user used the `DiagonalMass` alias.

applications/plugins/SofaCUDA/CMakeLists.txt
```diff
@@ -46,8 +46,6 @@ set(HEADER_FILES
 
 
     ### Mass
-    sofa/gpu/cuda/CudaDiagonalMass.h
-    sofa/gpu/cuda/CudaDiagonalMass.inl
```
alxbilger, line 50: Remove `CudaDiagonalMass.cpp` also line 126

PR #3010, opened 2022-06-07
[SofaSparseSolver] Fix the assembling of the constraint matrix

dessteph:
It seems that there was an error in the code of the SparseLDLSolver in the assembling of the constraint matrix.
I tried to see the execution of this function in the scene InextensiblePendulum.scn and it produced a wrong result at the first step (with the current master branch).
After this fix, it starts with a coherent behaviour but then it still fails.

I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco).

Reviewers will merge this pull-request only if

    it builds with SUCCESS for all platforms on the CI.
    it does not generate new warnings.
    it does not generate new unit test failures.
    it does not generate new scene test failures.
    it does not break API compatibility.
    it is more than 1 week old (or has fast-merge label).


alxbilger (commented):
Thanks for the fix! Now it needs a careful review.

In the meantime, I have questions:
1) I see that in `examples/Components/constraint/InextensiblePendulum.scn`, there is a `SparseCholeskySolver`, and not a `SparseLDLSolver`. I guess you replaced the solver in your tests, right? If I do that, the first iteration seems fine, but it is true that it explodes at some point. Can you give more details on what is going on in the first iteration?
2) I understand that you fixed the current implementation, and it's not a full refactoring. Nevertheless, can you say if it is still the most efficient implementation? For example, I see that `L^-1 *J^t` appears twice. Can it be reused for more efficiency?

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolver.inl
```diff
@@ -92,6 +92,12 @@ void SparseLDLSolver<TMatrix,TVector,TThreadManager>::invert(Matrix& M)
 template <class TMatrix, class TVector, class TThreadManager>
 bool SparseLDLSolver<TMatrix, TVector, TThreadManager>::doAddJMInvJtLocal(ResMatrixType* result, const JMatrixType* J, SReal fact, InvertData* data)
 {
+    /*
+    J*Minv*J = J*(L*D*L^t)^-1 * J^t
+             = J*(L^t)^-1*D^-1*L^-1*J^t
+             = J*(L^t)^-1*D^-1*(J*L^-1)^t
+             = (L^-1 *J^t)^t * D^-1 * (L^-1*J^t )
```
alxbilger:
```suggestion
    J*Minv*J^t = J*(L*D*L^t)^-1 * J^t
               = J*(L^t)^-1*D^-1*L^-1*J^t
               = J*(L^t)^-1*D^-1*(J*L^-1)^t
               = (L^-1 *J^t)^t * D^-1 * (L^-1*J^t )
```

dessteph:
1) Indeed I changed the linear solver used in the scene. In the master branch, the code actually compute another matrix which is not the one we need. Hence the simulation fail at the first step. For this branch the smilation failed when the message "D(k,k) is zero" appear. I don't know if there is an issue with this scene or with the solver.

2) In the code, we exploit the LDL factorization in order to solve only one triangular system (and not two). We are allowed to do that by rewriting the matricial product we need to compute. It will then write as A *D^-1 * A^t. We need to apply D^-1 and it won't commute, that is why there is two matrices in memory (we could store only one with the Cholesky factorization).

alxbilger:
@dessteph On my setup with the master branch, I manage to run the simulation several tens of iterations (see screenshot). So I would like to know what you mean by "the simulation fails".

image

dessteph:
After a certain number of iterations (on my computer), an error message appears in the terminal and the simulation becomes unstable (it explodes) and that doesn't append with LU or Cholesky.

alxbilger:
[ci-build][with-all-tests]

Sofa/GUI/Qt/src/sofa/gui/qt/init.cpp
```diff
@@ -24,11 +24,11 @@
 #include <sofa/gui/common/GUIManager.h>
 #include <sofa/gui/qt/RealGUI.h>
 
-#if SOFA_GUI_QT_HAVE_QGLVIEWER
+#if SOFA_GUI_QT_ENABLE_QGLVIEWER
 int QGLViewerGUIClass = sofa::gui::common::GUIManager::RegisterGUI("qglviewer", &sofa::gui::qt::RealGUI::CreateGUI, nullptr, 3);
 #endif
 
-#if SOFA_GUI_QT_HAVE_QTVIEWER
+#if SOFA_GUI_QT_ENABLE_QTVIEWER
```
hugtalbot: a rebase might be needed this is already merged

alxbilger:
@dessteph I suggest you start again your investigation but taking into account https://github.com/sofa-framework/sofa/pull/3072. In other words, in your CMake, enable the variable `METIS-GKLIB_GKRAND`. This way we should have comparable results across different architectures.

Sofa/GUI/Qt/src/sofa/gui/qt/init.cpp
alxbilger: done

alxbilger:
Please add unit tests based on the PR #3050

alxbilger:
@dessteph I prepared the test in Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp. Could you add the check of the result at the end? You will need to use another software (matlab?) to compute the ground truth and compare it to the variable `JMInvJt` containing our result. Thanks

alxbilger:
[ci-build]

alxbilger:
@dessteph What is the status of this pull request? Can you provide details why we should continue working on it or not?

dessteph:
I made an unitary test to compare the results of the LDL solver to those of the LU solver. The difference goes up to 9e-13. I would have like to compare LU and Cholesky to see if that difference is realy significative.

alxbilger:
And what about the result of the LDL solver with and without your changes?

dessteph:
I tested the master branch.

alxbilger:
So you cannot say if your fix really fixes anything

dessteph:
It seems it didn't need any fixing, there is a math error but compensated by a roundabout use of the CSC format.

alxbilger:
This PR can be closed. Nevertheless, I'll check SparseLDLSolver again, and add a unit test in a future PR

PR #3030, opened 2022-06-09
[MeshGmsh]Fixed false error detection in MeshGmsh.cpp file

adagolodjo:
Fixed false error detection in MeshGmsh.cpp file  when reading .msh type files.
Without these lines, we have this kind of error `'$Elements' expected, found '$Elements`

### DIFF
--- a/Sofa/framework/Helper/src/sofa/helper/io/MeshGmsh.cpp
+++ b/Sofa/framework/Helper/src/sofa/helper/io/MeshGmsh.cpp
@@ -385,7 +385,7 @@ bool MeshGmsh::readGmsh(std::ifstream &file, const unsigned int gmshFormat)
         }
 
         std::getline(file, cmd);
-        if (cmd != "$EndNodes")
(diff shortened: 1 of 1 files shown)


alxbilger:
From the error you get, I deduce that `cmd` is equal to `Elements` (without the $). In this case, I don't see how `cmd.substr(0, 9)` can be equal to `$Elements`

adagolodjo:
Sorry @alxbilger, typo in my rewrite of the output. In fact, the error starts with "$EndNodes", so no possibility to read the elements.
The `cmd=="$EndNodes\r"`, so the comparison is always wrong. 
@hugtalbot, yes, it works for the other file I have. This shouldn't change anything for files that don't put '\r' at the end of keywords.

alxbilger:
Thanks @Younesssss, it makes more sense now

alxbilger (approved):

alxbilger:
[ci-build][with-all-tests]

epernod (approved):
From what I understand cmd is longer than the code tested. Ok for me.

PR #3069, opened 2022-06-24
[runSofa] Fix CMake error if all plugins are disabled

fredroy:
With SofaNG now, the mandatory libs are neither plugin nor the deprecated "module" notion,

So while trying to configure SOFA for a minimum build (i.e no plugin) and without the compat layer (so no module),  I stumbled in this error which was trying to generate the list of plugins/modules. 
There was no check on the list itself (as the empty modules/plugins could never have been empty before)

### DIFF
--- a/applications/projects/runSofa/cmake/GeneratePluginConfig.cmake
+++ b/applications/projects/runSofa/cmake/GeneratePluginConfig.cmake
@@ -9,46 +9,44 @@ macro(sofa_generate_plugin_config config_filename)
 
     list(LENGTH _allTargets nbTargets)
 
-    # if no plugin (or no deprecated "module") then quit
-    if (${nbTargets} EQUAL 0)
-        return()
-    endif()
+    # do the generation only if there is any plugin
+    if (${nbTargets} NOT EQUAL 0)
+        math(EXPR len "${nbTargets} - 1")
 
-    math(EXPR len "${nbTargets} - 1")
+        set(_modulePrefix "MODULE")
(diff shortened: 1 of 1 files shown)


applications/projects/runSofa/cmake/GeneratePluginConfig.cmake
```diff
@@ -8,6 +8,12 @@ macro(sofa_generate_plugin_config config_filename)
     get_property(_allTargetNames GLOBAL PROPERTY __GlobalTargetNameList__)
 
     list(LENGTH _allTargets nbTargets)
+
+    # if no plugin (or no deprecated "module") then quit
+    if (${nbTargets} EQUAL 0)
+        return()
```
guparan: I'm worried when I see a `return` in a macro.
See [this part](https://cmake.org/cmake/help/latest/command/macro.html#macro-vs-function) of CMake doc

> A macro is executed as if the macro body were pasted in place of the calling statement. This has the consequence that a [return()](https://cmake.org/cmake/help/latest/command/return.html#command:return) in a macro body does not just terminate execution of the macro; rather, control is returned from the scope of the macro call. To avoid confusion, it is recommended to avoid [return()](https://cmake.org/cmake/help/latest/command/return.html#command:return) in macros altogether.

fredroy: 👍

fredroy:

guparan (approved):

PR #3091, opened 2022-07-01
[SofaPhysicsAPI] Clean CMake file and compile without compat

fredroy:
... by making optional SofaValidation (as its remaining components are deprecated)

The ifdef is a bit ugly but some refactoring are needed to add futures features like this (instead of adding getXXX and getNbXXX for each type of component in the API)

Some cleanups for the cmake file itself (more up-to-date Sofa cmake syntax) and the dependencies.

small user experience question:
- Should Sofa.component be forcefully linked with it? so the user knows that all components are available from the start
- or should the actual dependencies be managed by the user afterwards in its application?


guparan (approved):
CMake LGTM

PR #3105, opened 2022-07-06
[GUI.Qt] Move default format setup before application creation

alxbilger:
To fix rendering bug reported in https://github.com/sofa-framework/sofa/discussions/3100
Also restore the `msaa` command line option


Sofa/GUI/Qt/src/sofa/gui/qt/RealGUI.cpp
```diff
@@ -226,8 +227,23 @@ class RealGUIFileListener : public sofa::helper::system::FileEventListener
 
 //======================= STATIC METHODS ========================= {
 
+void RealGUI::setupSurfaceFormat()
+{
+    if(!SOFA_GUI_QT_ENABLE_VSYNC)
+    {
+        static std::once_flag flag;
```
fredroy: I did not know about that 👌

Sofa/GUI/Qt/src/sofa/gui/qt/RealGUI.cpp
```diff
@@ -226,8 +227,23 @@ class RealGUIFileListener : public sofa::helper::system::FileEventListener
...
+        std::call_once(flag, []
+        {
+            QSurfaceFormat format;
+            format.setSwapInterval(0);
+            QSurfaceFormat::setDefaultFormat(format);
```
fredroy: Should we not set here the  compatibility profile and so on ?
(and the MSAA sampling here but we dont have the arg here unfortunately)

alxbilger: Compatibility profile is set only in QtViewer, right?
QtViewer setups a new `QSurfaceFormat` based on the default format, which was modified previously to support Vsync (or not). So, I guess that for QtViewer the compatibility profile is set properly. What do you think?
Do you think all this setup must go in the default QSurfaceFormat?

fredroy: Humm I was thinking just to uniformize/factorize/gather in one location the surfaceFormat initialization (both qt viewers would share the same surfaceformat once for all)

alxbilger: Done. Let me know if it matches your thoughts

PR #3112, opened 2022-07-07
[Sofa.Component] Put the initilization code in init() instead of the entrypoint initExternalPlugin()

fredroy:
Investigation from the problem I got with regression, which was that the sofa::component::init()
In GCC, it was crashing, whereas on clang/msvc it was not.

After debugging, it appears that there was a infinite loop when calling all the init() of the submodules.
E.g Sofa.Component calls sofa::component::collision::init() which calls initExternalModule of Sofa.Component.Collision. 
But with gcc, it was not calling initExternalModule() of Sofa.Component.Collision but the one of Sofa.Component. and then tries to call again init() etc.

I suspect problems with demangling the function name in the binary (extern "C") and stuff like that, but I did not go further.
The solution anyway is to put the call of the initialization in init() and not in initExternalModule() (and not vice versa)

### DIFF
--- a/Sofa/Component/AnimationLoop/src/sofa/component/animationloop/init.cpp
+++ b/Sofa/Component/AnimationLoop/src/sofa/component/animationloop/init.cpp
@@ -32,11 +32,7 @@ extern "C" {
 
 void initExternalModule()
 {
-    static bool first = true;
-    if (first)
-    {
-        first = false;
-    }
+    init();
 }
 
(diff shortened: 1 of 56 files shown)


alxbilger (approved):

alxbilger (approved):

alxbilger:
[ci-build][with-all-tests]

epernod (approved):

PR #3116, opened 2022-07-07
[Mass] Make warnings become info_msg in masses

hugtalbot:
A PR to feed the discussion : the data link `l_topology` is emitting a warning when empty ONLY in the two masses DiagonalMass and MeshMatrixMass which is inconsistent to all other components having such a data (with the exact same behavior)


fredroy (approved):
I agree on the change!
And it should be changed back later to warn, **if** we decide to go visitor-less...

epernod:
For me those 2 components are the ONLY one updated right now. If a topology is **needed** for the good behavior of the component, it should be set using the link, otherwise you end up with unsure behavior.
Removing the warning will just let lazy people continue putting mass in the node without taking attention to what they are doing.

PR #3103, opened 2022-07-06
[runSofa] Fix the generation of the autoload list

fredroy:
Based on 
- #3102

The autoloading libraries (plugin/module/whatever) feature was not including the "new" libraries (i.e  Sofa.Component* ) since the fact that are neither plugin nor module (but still loadable)
This PR fixes it.
Autoloading is not really recommended but a lot of scenes outside this repo relies on it implicitely.

TODO: decide if autoloading should off by default (and force the users to explicitly load their own necessary libs in their scenes)

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #3120, opened 2022-07-08
[CMake] Use standard macros + fix out-of-tree builds

guparan:
- Use `sofa_add_module` intensively for all new NG modules.
- Use `sofa_add_collection` for all old modules that are now simple collections.
- Make Sofa.GUI a real module.
- Fix disabled targets inclusion in SDK

:warning: This PR changes basically all CMake option names. E.g. `PLUGIN_SOFAGENERALLOADER` becomes `COLLECTION_SOFAGENERALLOADER`. The good thing is that it's way clearer.

This PR also contains the same fix as #3103

EDIT 
New changes:
- Move Sofa.GL module under Sofa directory
- SofaKernel is gone.
- Everything was moved under applications/collections/deprecated
- New cosmetic feature in SofaMacros to hide sub-options when base-option is disabled (like cmake_dependent_option but simpler). E.g. when I disable SOFA_COMPONENT, all SOFA_COMPONENT_* are hidden.
- A bit of cleaning


alxbilger:
Does it mean that #3103 should be closed?

(Note that the build is failing)

CMakeLists.txt
```diff
@@ -159,86 +158,68 @@ endif()
...
+    sofa_add_module(Sofa/Component/Compat Sofa.Component.Compat ON)
 endif()
 
 ### SOFA (framework and components)
 add_subdirectory(Sofa)
 
 option(SOFA_WITH_OPENGL "Compile Sofa with OpenGL support. This will define the SOFA_WITH_OPENGL and SOFA_NO_OPENGL CMake variables." ON)
 if(SOFA_WITH_OPENGL)
-    add_subdirectory(${SOFA_KERNEL_SOURCE_DIR}/modules/Sofa.GL)
+    sofa_add_module(${SOFA_KERNEL_SOURCE_DIR}/modules/Sofa.GL Sofa.GL ON)
```
fredroy: Sofa.GL is supposed to be enabled/built before Sofa.GUI could be activated, as some packages inside Sofa.GL needs OpenGL

fredroy (changes requested):
I am a bit sad that we lost simplicity in cmake files 😢. 
I liked the fact that Sofa/framework was (supposed to?) be buildable by iteelf; and that the root SOFA CMakefile was kind of slimmer/less complicated and actually was (supposed to be ?) not mandatory.

CMakeLists.txt
fredroy: Maybe this is the opportunity to move simply (Sofa.)GL into the Sofa/ folder (similarly to (Sofa,)GUI)

guparan:
> I am a bit sad that we lost simplicity in cmake files cry. I liked the fact that Sofa/framework was (supposed to?) be buildable by iteelf; 

How? By unchecking the SOFA_COMPONENT option? It's still possible this way. The thing is that Framework is (unfortunately) not a module and thus should not be treated as one if we want to be consistent. I'm not against reverting the framework/CMakeLists.txt deletion though.

> and that the root SOFA CMakefile was kind of slimmer/less complicated and actually was (supposed to be ?) not mandatory.

This file was and still is mandatory. At least, I think it's clearer now with the changes I propose because we see what is directly below the root CMakeLists (all the remaining add_subdirectory calls) and what could be done from somewhere else. Ideally we should just have a some options/variables and a bunch of add_subdirectory calls in this file.

guparan:
I pushed a lot of new changes and file moves.
I'm ok to split the PR in 2 if necessary.

guparan:
New changes:
- Move Sofa.GL module under Sofa directory
- SofaKernel is gone.
- Everything was moved under applications/collections/deprecated
- New cosmetic feature in SofaMacros to hide sub-options when base-option is disabled (like cmake_dependent_option but simpler). E.g. when I disable SOFA_COMPONENT, all SOFA_COMPONENT_* are hidden.
- A bit of cleaning

CMakeLists.txt
guparan: Done.

guparan:
> Does it mean that #3103 should be closed?
> 
> (Note that the build is failing)

Yep, doing it right now.

guparan:
https://github.com/guparan/sofa/blob/cmake_ng_use_standard_macros/Sofa/GUI/src/sofa/gui/init.cpp should not be a plugin

fredroy:

fredroy:

guparan:

guparan:

guparan:
Here we are. Ready to re-review @fredroy :wink:

fredroy (approved):
I guess I would need more info with a sofadev meeting but unfortunately I wont be able to attend this week 😢
as far as I understood you introduce a new macro defining the type of the project, which looks good to me (and defined library/collection/module/plugin as we said) and removed the _external suffix. 
Looks good to me as far as I can see ! (even if I would have wished you keep the Sofa/framework/CMakeLists.txt :p )

guparan:
> I guess I would need more info with a sofadev meeting but unfortunately I wont be able to attend this week cry as far as I understood you introduce a new macro defining the type of the project, which looks good to me (and defined library/collection/module/plugin as we said) and removed the _external suffix. Looks good to me as far as I can see ! (even if I would have wished you keep the Sofa/framework/CMakeLists.txt :p )

Exactly, I refactored the sofa_add_* macros under one sofa_add_subdirectory macro. The reason is that I didn't want to create a misleading sofa_add_library macro. I like the fact that sofa_* macro have the same name than the standard CMake macro they override.
I'll resolve the conflicts and revert the framework/CMakeLists removal :wink:

guparan:
Here you go @fredroy

guparan:
> Several points here:
> 
>     * this PR might be breaking for plugin compilation (including move of files)

No it is not. The only change that could be called "breaking" is the renaming of the CMake options to enable/disable libraries and modules. However, those options keep the same default behavior so it will only affect people who build their own SOFA with custom options. New option names are very close to old option name so it's pretty easy to do the transition. 

> 
>     * this PR changes the way we organize / name the code (which is good!)
> 
> 
> It would be therefore nice to rediscuss it with you both @guparan and @fredroy to all agree on the organization and merge it, but a question comes: does this seem realistic to merge it before the release?

Yes because having everything called PLUGIN_* is a nightmare.

PR #3124, opened 2022-07-11
[Sofa.Type] Deprecate unsafe Vec constructor from pointer

fredroy:
One of the constructor of type::Vec takes a raw pointer and fill itself from this pointer, using its **own size**.
But the ctor does not know the size of the given array so this can contribute to unsafe behavior.
E.g 
```
SReal array[2] = {1.0, 1.0};
sofa::type::Vec<3, SReal> vec(array);
```
is valid and will compile; but it will lead to unexpected result (in the best case, a crash)
GCC detected this case while compiling SpringForcefield with the bbox function on 2d and 1d datatypes.
And this leads to an other problem, with the case of TBoundingBox, which relies fully on this mechanism.
This PR removes as much as possible the usage of the unsafe ctor (effectuvely removing TBoundingBox usage) but does not deprecate TBoundingBox by itself (I suppose it should be)


fredroy:
[ci-build]

fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

PR #3135, opened 2022-07-18
[Lagrangian.Correction] LinearSolverConstraintCorrection: Trivial optimisations for MSVC

fredroy:
It seems msvc is quite (much!) less clever than gcc.

2 optimizations (only useful when not building in Release mode)
 - add and use a NOINIT RigidDeriv constructor (avoiding to init a deriv which is filled afterwards)
 - default ctor directly call the identity of its components (which is their default ctor by the way)

and 1 ridiculous optimization
 - avoid calling getPositionIntegrationFactor() in the nested loop
 
This optimization brings a whooping 30% speedup on my setup (VC2022, Release/LTO enabled)
Test scene: 3instruments_collis.scn from BeamAdapter, 2000 iterations in batch mode
 - before: 16.5669 s ( 120.723 FPS)
 - after: 12.829 s ( 155.897 FPS)

GCC was doing the same by the way with or without these optims (and is faster than msvc, even in a VM 🤔)

NOTE: do not forget to enable the LTO option for Windows when deploying, it brings full inlining & stuff and it is really useful for perf
(w/o LTO, before the optims it is 21.9883 s ( 90.9575 FPS) )

### DIFF
--- a/Sofa/Component/Constraint/Lagrangian/Correction/src/sofa/component/constraint/lagrangian/correction/LinearSolverConstraintCorrection.inl
+++ b/Sofa/Component/Constraint/Lagrangian/Correction/src/sofa/component/constraint/lagrangian/correction/LinearSolverConstraintCorrection.inl
@@ -572,6 +572,7 @@ void LinearSolverConstraintCorrection<DataTypes>::addConstraintDisplacement(doub
 
     _new_force = false;
 
+    const auto positionIntegrationFactor = odesolver->getPositionIntegrationFactor();
     // TODO => optimisation => for each bloc store J[bloc,dof]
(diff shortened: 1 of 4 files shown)


fredroy:
[ci-build][with-all-tests]

alxbilger:
@fredroy thanks to Jason Turner ;)

fredroy:
> @fredroy thanks to Jason Turner ;)

It was more "Thanks to the Performance Profiler "😅

alxbilger:
Ah ok, because Turner made two recent videos about LTO

alxbilger (approved):
:+1:

PR #3162, opened 2022-07-22
[Component.Setting] Introduce PrintLog component to set printLog to all the graph components

alxbilger:
A detailed description of the component is provided in the code
Unit tests are added.


alxbilger:
[ci-build][with-all-tests]

epernod:
I understand the feature, I understand the need and I agree I would use it.... but I find really weird to set the option using a component in the graph...

Do you know if there is an option in the Node? we could use the same logic but per Node

epernod:
Ok sorry my message was not totally clear. I meant, you check the option in the Node and it propagate the printLog to all child Node and their components.
So if you want all the printLog, you activate it in the root Node.
Just suggesting.

alxbilger:
Thanks for you feedback @epernod. I think your suggestion is the more natural. Actually, I tried to set `printLog` in a log, to see if it was already working (but no). But I always feel uncomfortable when modifying such an important base class. I don't want that some classes become God classes. That's why I did not touch the Node class. But I am definitively open to discussion

alxbilger:
I am ok with the suggestion

PR #3178, opened 2022-07-27
[objectmodel] Reduce memory leaks

alxbilger:
`TClass<T, Parents>` was instantiated on the heap and never free, causing memory leaks. An option is to instantiate the class on the stack so it is destroyed at the end of the program. But currently, the compilation fails if `TClass<T, Parents>` is instantiated on the stack with T an abstract class. That is why there are now two implementations.


alxbilger:
Replaced by https://github.com/sofa-framework/sofa/pull/3183

PR #3183, opened 2022-07-28
[objectmodel] Remove memory leaks

alxbilger:
The following code create memory leaks:

```cpp
static const BaseClass* get()
{
  static TClass<T, Parents> *theClass=new TClass<T, Parents>();
  return theClass;
}
```

To remove the memory leak, it has to be replaced by:

```cpp
static const BaseClass* get()
{
  static TClass<T, Parents> theClass;
  return &theClass;
}
```

However, `Parents` can be a `std::pair`, which one of the pair type can be an abstract type. Since it is not possible to instantiate an abstract type, the compilation fails.

This PR removes the use of `std::pair` to introduce a structure that does not need to instantiate the types. The compilation now succeeds.


alxbilger:
[ci-build][with-all-tests]

PR #3225, opened 2022-08-17
[CMake] FIX parent modules library consistency

guparan:
Parent modules must always interface the same libraries.
e.g. Sofa.Component.dll must always be the same

Thus, if some of its child modules are disabled, parent module will not generate a library.

This PR creates and uses a new macro `sofa_add_subdirectory_modules`.
This macro is meant for parent modules to easily add their child modules and sets a variable with all added targets.
If a child module is missing or disabled, a message will be shown (not an error) and the library will not be created.


fredroy (approved):

fredroy:
[ci-build][with-all-tests]

PR #3227, opened 2022-08-18
[all] Set of warning fixes

hugtalbot:
some day a warning-free SOFA!

### DIFF
--- a/Sofa/Component/SolidMechanics/FEM/Elastic/tests/TriangleFEMForceField_test.cpp
+++ b/Sofa/Component/SolidMechanics/FEM/Elastic/tests/TriangleFEMForceField_test.cpp
@@ -450,7 +450,6 @@ public:
                 
                 Real factor = triFEM->getTriangleFactor(id); // ((Real)0.5)/(ti.bx*ti.cy); -> 1/(2 * det) = 1/area                
                 Real correctiveFactorStiff = 1 / (4 * factor); // TODO: epernod 2021-08-03: there is a big diff here regarding the equation used in TriangleFEMForceField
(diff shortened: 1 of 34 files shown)


alxbilger (changes requested):
Usually, it is preferred to use `static_cast` instead of a C-style cast for the following reasons:
- Compile-time check that the type can be cast
- You can search for `static_cast`
- More readable

If the destination type is narrower, use `sofa::helper::narrow_cast`.

In general, for loop indices, there are ways to avoid casts. Usually, types can be changed to make them compatible. Sometimes, it requires a lot of work.

Sofa/Component/SolidMechanics/FEM/Elastic/tests/TriangleFEMForceField_test.cpp
```diff
@@ -450,7 +450,6 @@ class TriangleFEMForceField_test : public BaseTest
                 
                 Real factor = triFEM->getTriangleFactor(id); // ((Real)0.5)/(ti.bx*ti.cy); -> 1/(2 * det) = 1/area                
                 Real correctiveFactorStiff = 1 / (4 * factor); // TODO: epernod 2021-08-03: there is a big diff here regarding the equation used in TriangleFEMForceField
-                Real correctiveFactorStrainD = factor * 2; // TODO: epernod 2021-08-03: there is a big diff here regarding the equation used in TriangleFEMForceField
```
alxbilger, line 453: I guess you removed this line because the variable is not used?

alxbilger (approved):

Sofa/Component/SolidMechanics/FEM/Elastic/tests/TriangleFEMForceField_test.cpp
hugtalbot: yes

PR #3260, opened 2022-08-31
[DrawToolGL] Fix ill-formed drawLines with multiple colors

fredroy:
Got a GlError() somewhere else and finding where it came from was a pain...
It appears that drawLines (version with multiple colors) was incorrectly implemented, as it was calling setMaterial/resetMaterial between glBegin()/glEnd() and this is not possible apparently because those 2 functions call glDisable() and glDepthMask() and it is not possible to call them between glBegin/glEnd. (yes old OpenGL is... marvelous)

Anyway this PR regroups the lines per color and call the drawLine() with one color. Implementation with map is not nice IMO but all the DrawTool implementation itself is, anyway.

\+ UnilateralInteractionConstraint was calling twice saveLastState()

### DIFF
--- a/Sofa/GL/src/sofa/gl/DrawToolGL.cpp
+++ b/Sofa/GL/src/sofa/gl/DrawToolGL.cpp
@@ -144,19 +144,25 @@ void DrawToolGL::drawLines(const std::vector<Vector3> &points, float size, const
         return drawLines(points, size, RGBAColor::red());
     }
 
-    glLineWidth(size);
-    if (getLightEnabled()) disableLighting();
-    glBegin(GL_LINES);
+    // gather lines with same colors
+    std::map<RGBAColor, std::vector<Vector3> > colorPointsMap;
+    for (std::size_t i = 0; i < colors.size(); ++i)
     {
-        for (std::size_t i=0; i<points.size()/2; ++i)
+        if (colorPointsMap.find(colors[i]) == colorPointsMap.end())
         {
-            setMaterial(colors[i]);
-            internalDrawLine(points[2*i],points[2*i+1], colors[i] );
-            resetMaterial(colors[i]);
+            colorPointsMap.insert({ colors[i] , {} });
         }
-    } glEnd();
-    if (getLightEnabled()) enableLighting();
-    glLineWidth(1);
+
+        colorPointsMap[colors[i]].push_back(points[2 * i]);
+        colorPointsMap[colors[i]].push_back(points[2 * i + 1]);
+    }
+
+    // call the drawLine method which takes only one color
+    for (auto [color, points] : colorPointsMap)
+    {
+        drawLines(points, size, color);
+    }
+
 }
 
 /////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
--- a/Sofa/GL/src/sofa/gl/DrawToolGL.cpp
+++ b/Sofa/GL/src/sofa/gl/DrawToolGL.cpp
@@ -148,17 +148,12 @@ void DrawToolGL::drawLines(const std::vector<Vector3> &points, float size, const
     std::map<RGBAColor, std::vector<Vector3> > colorPointsMap;
     for (std::size_t i = 0; i < colors.size(); ++i)
     {
-        if (colorPointsMap.find(colors[i]) == colorPointsMap.end())
-        {
-            colorPointsMap.insert({ colors[i] , {} });
-        }
-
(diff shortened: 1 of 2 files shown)


Sofa/GL/src/sofa/gl/DrawToolGL.cpp
```diff
@@ -144,19 +144,25 @@ void DrawToolGL::drawLines(const std::vector<Vector3> &points, float size, const
         return drawLines(points, size, RGBAColor::red());
     }
 
-    glLineWidth(size);
-    if (getLightEnabled()) disableLighting();
-    glBegin(GL_LINES);
+    // gather lines with same colors
+    std::map<RGBAColor, std::vector<Vector3> > colorPointsMap;
+    for (std::size_t i = 0; i < colors.size(); ++i)
```
alxbilger, line 149: Here is what I suggest to replace your loop:
```cpp
for (std::size_t i = 0; i < colors.size(); ++i)
{
    auto& colorPoints = colorsPointsMap[colors[i]];
    colorPoints.push_back(points[2 * i]);
    colorPoints.push_back(points[2 * i + 1]);
}
```

With your version: 3 searches in the map
With my version: 1 search in the map (and more concise)

Sofa/GL/src/sofa/gl/DrawToolGL.cpp
```diff
@@ -144,19 +144,25 @@ void DrawToolGL::drawLines(const std::vector<Vector3> &points, float size, const
         return drawLines(points, size, RGBAColor::red());
     }
 
-    glLineWidth(size);
-    if (getLightEnabled()) disableLighting();
-    glBegin(GL_LINES);
+    // gather lines with same colors
+    std::map<RGBAColor, std::vector<Vector3> > colorPointsMap;
```
alxbilger, line 148: an `unordered_map` should be faster

Sofa/GL/src/sofa/gl/DrawToolGL.cpp
```diff
@@ -144,19 +144,25 @@ void DrawToolGL::drawLines(const std::vector<Vector3> &points, float size, const
...
-    } glEnd();
-    if (getLightEnabled()) enableLighting();
-    glLineWidth(1);
+
+        colorPointsMap[colors[i]].push_back(points[2 * i]);
+        colorPointsMap[colors[i]].push_back(points[2 * i + 1]);
+    }
+
+    // call the drawLine method which takes only one color
+    for (auto [color, points] : colorPointsMap)
```
alxbilger: I am not sure here if you copy a pair of key/value or if it is implicitly a reference. That's why I would have written `for (const auto& [color, points] : colorPointsMap)`

alxbilger (changes requested):
Congratulations for having found this problem

Sofa/GL/src/sofa/gl/DrawToolGL.cpp
fredroy: At the beginning I used a unordered_map indeed, but I got some compile error (about compare/hashing) and I was lazy to solve it

Sofa/GL/src/sofa/gl/DrawToolGL.cpp
fredroy: 👍
I am still in the process to force myself to use the [X, Y] syntax

Sofa/GL/src/sofa/gl/DrawToolGL.cpp
fredroy: Sure that `     auto& colorPoints = colorsPointsMap[colors[i]];`  wont generate an error ?
I mean at the beginning there is no colors[i] in the map, and there is no std::vector<> allocated/built but maybe I am wrong

alxbilger: See https://en.cppreference.com/w/cpp/container/map/operator_at. This operator performs an insertion if the key is not in the map.

Sofa/GL/src/sofa/gl/DrawToolGL.cpp
alxbilger: Indeed, a hash function should be available for `RGBAColor`. Keep using a map 👍

Sofa/GL/src/sofa/gl/DrawToolGL.cpp
fredroy: Okay at() was behaving like I thought operator[] was 🤫

hugtalbot (approved):

PR #3264, opened 2022-09-01
[Core] Reduce calls to getValue in collisions

alxbilger:
... and also reduce complexity of the search of common groups.

### DIFF
--- a/examples/Components/collision/CollisionGroups.scn
+++ b/examples/Components/collision/CollisionGroups.scn
@@ -1,0 +1,67 @@
+<?xml version="1.0"?>
+
+<!--
+In this scene, 3 cubes fall on the floor. They must not collide together, but they must collide with the floor. To
+achieve this behavior, collision groups are used. Collision models contained in the same group don't collide together.
+Here, the collision models of the 3 cubes are defined in the group 1. Therefore, they don't collide. The floor is not
+in the group 1, therefore cubes collide with it.
+-->
+
+<Node name="root" dt="0.01" gravity="0 -9.81 0">
+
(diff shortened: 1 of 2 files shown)


fredroy:
[ci-build][with-all-tests]

examples/Components/collision/CollisionGroups.scn
```diff
@@ -0,0 +1,67 @@
...
+            <MechanicalObject name="Cube_RigidDOF" template="Rigid3d" translation="0 16 1" />
+            <UniformMass name="UniformMass" totalMass="10.0" />
+            <OBBCollisionModel group="1"/>
+        </Node>
+
+    </Node>
+
+    <Node name="Floor">
+        <MechanicalObject name="Floor Particles" template="Rigid3d"/>
+        <OBBCollisionModel name="Floor Triangle For Collision" extents="5 1 5" moving="0" simulated="0" />
```
hugtalbot, line 64: Maybe specify a group for the Floor, no?

```suggestion
        <OBBCollisionModel name="Floor Triangle For Collision" group="0" extents="5 1 5" moving="0" simulated="0" />
```

alxbilger: Well, it is not necessary and `group="0"` is different than not having a group (see the algorithm changed by this pull request). But in this scene, the behavior will be the same. So I guess the decision to add it or not is an educational point of view (the scene is in the example folder). What do we want to teach to the users?

hugtalbot: you're right and what you did suit me too (I was just used to define a group for all when used) 
:+1:

hugtalbot (approved):

fredroy (approved):

PR #3267, opened 2022-09-02
[core] remove shadow variables (modern version)

maddin200:

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
+++ b/Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
@@ -82,14 +82,14 @@ ElementIntersector* IntersectorMap::get(core::CollisionModel* model1, core::Coll
 
 
     std::stringstream tmp;
-    for(auto myit = intersectorsMap.begin(), itEnd = intersectorsMap.end(); myit != itEnd; ++myit)
+    for(auto intersectorsMapIt = intersectorsMap.begin(), itEnd = intersectorsMap.end(); intersectorsMapIt != itEnd; ++intersectorsMapIt)
     {
-        auto typeInfo1 = myit->first.first;
-        auto typeInfo2 = myit->first.second;
+        auto typeInfo1 = intersectorsMapIt->first.first;
+        auto typeInfo2 = intersectorsMapIt->first.second;
         tmp << "  "
                 << gettypename(typeInfo1) << "-"
                 << gettypename(typeInfo2);
-        ElementIntersector* i = myit->second;
+        ElementIntersector* i = intersectorsMapIt->second;
         if (!i)
             tmp << "  nullptr";
         else
--- a/Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
+++ b/Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
@@ -48,11 +48,11 @@ helper::TypeInfo IntersectorMap::getType(core::CollisionModel* model)
     if (it == castMap.end())
     {
         helper::TypeInfo t2 = t;
-        for (std::set<const objectmodel::ClassInfo* >::iterator it = classes.begin(); it != classes.end(); ++it)
+        for (auto myit = classes.begin(); myit != classes.end(); ++myit)
         {
-            if ((*it)->isInstance(model))
+            if ((*myit)->isInstance(model))
             {
-                t2 = (*it)->type();
+                t2 = (*myit)->type();
                 break;
             }
         }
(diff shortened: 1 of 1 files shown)


Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
```diff
@@ -48,11 +48,11 @@ helper::TypeInfo IntersectorMap::getType(core::CollisionModel* model)
     if (it == castMap.end())
     {
         helper::TypeInfo t2 = t;
-        for (std::set<const objectmodel::ClassInfo* >::iterator it = classes.begin(); it != classes.end(); ++it)
+        for (auto myit = classes.begin(); myit != classes.end(); ++myit)
```
hugtalbot: maybe rename `it` above with `typeInfoIt` and `myit` with `classInfoIt` for readibility

Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
```diff
@@ -82,14 +82,14 @@ ElementIntersector* IntersectorMap::get(core::CollisionModel* model1, core::Coll
 
 
     std::stringstream tmp;
-    for(InternalMap::const_iterator it = intersectorsMap.begin(), itEnd = intersectorsMap.end(); it != itEnd; ++it)
+    for(auto myit = intersectorsMap.begin(), itEnd = intersectorsMap.end(); myit != itEnd; ++myit)
```
hugtalbot: and this `myIt` into `intersectorsMapIt`

Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
```diff
@@ -82,14 +82,14 @@ ElementIntersector* IntersectorMap::get(core::CollisionModel* model1, core::Coll
...
     {
-        helper::TypeInfo t1 = it->first.first;
-        helper::TypeInfo t2 = it->first.second;
+        auto tmpt1 = myit->first.first;
+        auto tmpt2 = myit->first.second;
```
hugtalbot:
```suggestion
        auto typeInfo1 = myit->first.first;
        auto typeInfo2 = myit->first.second;
```

Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
```diff
@@ -82,14 +82,14 @@ ElementIntersector* IntersectorMap::get(core::CollisionModel* model1, core::Coll
...
         tmp << "  "
-                << gettypename(t1) << "-"
-                << gettypename(t2);
-        ElementIntersector* i = it->second;
+                << gettypename(tmpt1) << "-"
+                << gettypename(tmpt2);
```
hugtalbot:
```suggestion
                << gettypename(typeInfo1) << "-"
                << gettypename(typeInfo2);
```

hugtalbot (changes requested):

maddin200:
changed

Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
```diff
@@ -48,11 +48,11 @@ helper::TypeInfo IntersectorMap::getType(core::CollisionModel* model)
     if (it == castMap.end())
     {
         helper::TypeInfo t2 = t;
-        for (std::set<const objectmodel::ClassInfo* >::iterator it = classes.begin(); it != classes.end(); ++it)
+        for (auto classInfoIt = classes.begin(); classInfoIt != classes.end(); ++classInfoIt)
```
alxbilger, line 51: To avoid such long for loop declarations and dealing with useless iterators, there are range-based for loops: https://en.cppreference.com/w/cpp/language/range-for

So I would have written this loop: `for (const auto& classInfo : classes)`

I think it is more readable.

But ultimately, this for loop is just a search for an element in a `std::set`. There is `std::find_if` for that:

```cpp
const auto findIt = std::find_if(classes.begin(), classes.end(),
    [model](const auto* classInfo){ return classInfo->isInstance(model); });
if (findIt != classes.end())
    t2 = (*findIt)->type();
```

Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
```diff
@@ -82,14 +82,14 @@ ElementIntersector* IntersectorMap::get(core::CollisionModel* model1, core::Coll
 
 
     std::stringstream tmp;
-    for(InternalMap::const_iterator it = intersectorsMap.begin(), itEnd = intersectorsMap.end(); it != itEnd; ++it)
+    for(auto intersectorsMapIt = intersectorsMap.begin(), itEnd = intersectorsMap.end(); intersectorsMapIt != itEnd; ++intersectorsMapIt)
```
alxbilger, line 85: Here, since the container is a `std::map`, you can use structured binding (https://en.cppreference.com/w/cpp/language/structured_binding):

```cpp
for (const auto& [typeInfoPair, intersector] : intersectorsMap)
{
    tmp << "  "
            << gettypename(typeInfoPair.first) << "-"
            << gettypename(typeInfoPair.second);
    tmp << "  " << (intersector ? intersector->name() : "nullptr") << msgendl;
}
```

alxbilger (changes requested):
Thanks for the pull request. It is indeed important to make the code more readable while taking advantages of modern C++. Here are my suggestions

maddin200:
probably I have no time to test it.

Sofa/framework/Core/src/sofa/core/collision/Intersection.cpp
hugtalbot: maybe avoiding using again `t2` @alxbilger ;)

PR #3271, opened 2022-09-05
[Topology] Improve TopologyHandler registration

epernod:
- Use std::set instead of std::list in TopologyContainer to store TopologyHandler
- Use bool m_isRegistered to tack info if the topologyHandler is registed inside the TopologyContainer list. Catch return of std::insert to update  this bool.
- Factorize method to link topologyHandler using sofa::geometry::ElementType


epernod:
[ci-build][with-all-tests]

epernod:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/topology/BaseTopology.cpp
```diff
@@ -168,11 +169,11 @@ void TopologyContainer::resetTopologyHandlerList()
 {
     for (auto& topologyHandlerList : m_topologyHandlerListPerElement)
     {
-        for (auto it = topologyHandlerList.begin();
-            it != topologyHandlerList.end(); ++it)
+        std::for_each(topologyHandlerList.begin(), topologyHandlerList.end(), [](TopologyHandler* topoHandler) 
         {
-            *it = nullptr;
-        }
+            topoHandler = nullptr;
```
alxbilger: I am not sure this is doing what you think. `topoHandler` is a copy, not a reference

epernod: you are right.. 
And not sure the entire loop is really useful for this PR (I have more change incoming)

Sofa/framework/Core/src/sofa/core/topology/BaseTopology.cpp
```diff
@@ -98,12 +98,13 @@ void TopologyContainer::addStateChange(const TopologyChange *topologyChange)
     m_stateChangeList.endEdit();
 }
 
-void TopologyContainer::addTopologyHandler(TopologyHandler *_TopologyHandler, sofa::geometry::ElementType elementType)
+bool TopologyContainer::addTopologyHandler(TopologyHandler *_TopologyHandler, sofa::geometry::ElementType elementType)
```
hugtalbot, line 101: nodiscard?

Sofa/framework/Core/src/sofa/core/topology/TopologyHandler.cpp
```diff
@@ -161,5 +161,10 @@ bool TopologyHandler::registerTopology(sofa::core::topology::BaseMeshTopology* _
     return true;
 }
 
+void TopologyHandler::linkToTopologyDataArray(sofa::geometry::ElementType elementType)
+{
+    SOFA_UNUSED(elementType);
+}
+
```
hugtalbot, line 168: isn't it strange to define an empty function in TopologyHandler?
can this function move in TopologyDataHandler directly?

hugtalbot (commented):
Nice refactoring @epernod 
A few comments below

Sofa/framework/Core/src/sofa/core/topology/TopologyHandler.cpp
epernod: I need it in TopologyHandler as it is the base class (not templated) used in TopologyData.
It is overridden in TopologyDataHandler.

epernod:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/topology/TopologyHandler.h
```diff
@@ -169,6 +166,9 @@ class SOFA_CORE_API TopologyHandler : public sofa::core::objectmodel::DDGNode
     /// use to define data handled name.
     std::string m_data_name;
 
+    /// Bool to store the information if this topology handler is registered into Topology handler lists
+    int m_isRegistered = false;
```
alxbilger:
```suggestion
    bool m_isRegistered = false;
```

Sofa/framework/Core/src/sofa/core/topology/TopologyDataHandler.inl
```diff
@@ -63,160 +59,35 @@ void TopologyDataHandler<TopologyElementType,  VecT>::init()
...
-    if (m_edgesLinked) // avoid second registration
-        return;
 
-    if (m_topology == nullptr)
+    if (m_topology->linkTopologyHandlerToData(this, elementType))
     {
-        msg_error(m_topologyData->getOwner()) << "Owner topology has not been set. Data '" << m_data_name << "' won't be linked to Edge Data Array.";
-        return;
-    }
+        bool res = m_topology->addTopologyHandler(this, elementType);
```
alxbilger:
```suggestion
        const bool res = m_topology->addTopologyHandler(this, elementType);
```

alxbilger (changes requested):

Sofa/framework/Core/src/sofa/core/topology/TopologyHandler.h
```diff
@@ -137,19 +137,16 @@ class SOFA_CORE_API TopologyHandler : public sofa::core::objectmodel::DDGNode
     virtual void ApplyTopologyChange(const TopologyChangeElementInfo<Topology::Tetrahedron>::EMoved* /*event*/) {}
     virtual void ApplyTopologyChange(const TopologyChangeElementInfo<Topology::Hexahedron>::EMoved* /*event*/) {}
 
-
-
-    virtual bool isTopologyDataRegistered() { return false; }
+    /// Method to notify that this topologyHandler is not anymore registerd into a Topology Container
+    void unRegisteredTopologyHandler() { m_isRegistered = false; }
```
hugtalbot: you meant `unregisterTopologyHandler()` ?

Sofa/framework/Core/src/sofa/core/topology/TopologyHandler.h
```diff
@@ -137,19 +137,16 @@ class SOFA_CORE_API TopologyHandler : public sofa::core::objectmodel::DDGNode
...
+    /// Method to get the information if this topologyHandler is registered into a Topology Container
+    bool isTopologyHandlerRegistered() const { return m_isRegistered; }
```
hugtalbot, line 143: is there no way to avoid the use of the bool to know whether the handler is registered or not?
std::find or similar?

epernod:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/topology/TopologyHandler.h
epernod: Yes and no.
Theoretically it could be possible to not use the bool and this method would return something like `m_topologyContainer->isTopologyHandlerRegistered(this)`  where isTopologyHandlerRegistered will do a find in its std::set of topologyHandler.

But I use this bool to "fake a smart pointer mechanism". 
This is needed because TopologyData/handler are created at init by a FEM for example, and then they are managed by the TopologyContainer. In a classic simulation, at the end, Topology is deleted and the set of topologyHandler is only clear (not deleted), and when the FEM is deleted, the topologyHandler is deleted as well.

But in the scenario we checked where the mouse pickHandler is crashing, in fact a springFF is creating, adding a new TopologyHandler and when you release the mouse interaction, the springFF is deleted, which also delete the topologyHandler but not removed from TopologyContainer list. So in a incoming PR (to not mix all the changed, I thought this one would be quickly merged) I'm adding all the mechanism to unregister a TopologyHandler when it is destroyed.

But this raised another problem when the scene is deleted. To unregister itself, the handler need to access the topologyContainer which could have already been destroyed (depending on the graph order). That's why if the TopologyContainer is deleted first it sets all its topologyHander as unregistered.  

Using SPtr would have work as expected I think but now I don't remember why I couldn't use it (probably because we don't know which class of topologyContainer will be used/find in the scene)

epernod: Full picture available in #3369

epernod:
[ci-build][with-all-tests]

fredroy:
One unit test is still failing on mac and unbutu ...👺
```
3271/macos_clang_options_release/src/Sofa/Component/Topology/Container/Dynamic/tests/EdgeSetTopology_test.cpp:352
Expected equality of these values:
  edgeH0->getName()
    Which is: "TopologyDataHandler( VectorSpringForceField )springs"
  "TopologyDataHandler( MeshMatrixMass )edgeMass"
 ```

hugtalbot (approved):

fredroy:
@hugtalbot, test is fixed somehow so... yes

alxbilger:
There is crashing scene on Windows no?

epernod:
yes flashing, it is already failing on master sometimes: https://www.sofa-framework.org/dash/?branch=origin/master

fredroy:
@hugtalbot 🙆

PR #3305, opened 2022-09-15
[cmake] Introduce CMake presets

alxbilger:
See doc https://cmake.org/cmake/help/latest/manual/cmake-presets.7.html

It gives 3 different presets similarly to what we are used to on the CI. The idea is the same (3 different config) but it cannot mimic it perfectly. For example, the CI checks if dependencies are available before setting a variable. It is not possible with presets.

An example of CMakeUserPresets.json I use: https://gist.github.com/alxbilger/bee1135fbe222943906d0f595699c1a3

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -46,6 +46,7 @@ CMakeLists.txt.user*
 
 # custom cmake file
 custom.cmake
+CMakeUserPresets.json
 
 # caching dirs (clangd, ccls....)
 .cache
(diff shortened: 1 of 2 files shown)


fredroy:
👍
It is the official way of what the custom.cmake file was supposed to do.

As a side note, cmake-presets has been introduced in cmake v3.19 (so it will not be possible to use this feature for Ubuntu 20.04 for example)

alxbilger:
A screenshot where the preset can be selected in the CMake GUI:

image

PR #3313, opened 2022-09-19
[Config] Option to set the number of MSVC processes

alxbilger:
Quite intrusive in SOFA but I don't see another way to define it without an overwrite each time CMake runs


Sofa/framework/Config/CMakeLists.txt
```diff
@@ -172,7 +172,14 @@ if(WIN32)
     list(APPEND SOFACONFIG_COMPILE_OPTIONS "-D_CRT_SECURE_NO_WARNINGS")
     list(APPEND SOFACONFIG_COMPILE_OPTIONS "-D_CRT_NONSTDC_NO_DEPRECATE")
 
-    list(APPEND SOFACONFIG_COMPILE_OPTIONS "/MP;/wd4250;/wd4251;/wd4275;/wd4675;/wd4661")
+    set(SOFA_MSVC_NB_COMPILER_PROCESSES "0" CACHE STRING "The maximum number of processes that the compiler can create. 0 (default) for the number of effective processors on the computer")
+    if (SOFA_MSVC_NB_COMPILER_PROCESSES EQUAL "0")
+        list(APPEND SOFACONFIG_COMPILE_OPTIONS "/MP")
+    else()
+        list(APPEND SOFACONFIG_COMPILE_OPTIONS "/MP${SOFA_MSVC_NB_COMPILER_PROCESSES}")
```
fredroy: This is UX and cmake is shit for this kind of stuff, but I think you should do a limited choice for the number of processes. Otherwise the string can contain anything and the flag /MPblabla will be wrong (and I suppose Visual Studio wont be happy 🙊')
This is only for the Cmake GUI but you can make a combobox instead with 
`set_property(CACHE SOFA_MSVC_NB_COMPILER_PROCESSES PROPERTY STRINGS 0 1 2 4 8 16)`
and it will limit the choices for the string to 0, 1, 2 etc
I dont know how it will behave if you use only cmake commands line options.

alxbilger: It's a shame it is not possible to set integer variables. So we have two solutions:

1) Define a string where the user can set anything
2) Define a combobox where the user is limited

What about filtering the string with a regex?

fredroy: it could be a good solution I think, at least to detect stupid inputs at cmake config step

fredroy (approved):

Issue #3318, opened 2022-09-21
Memory leak in Simulation.unload() caused by collision pipeline


ScheiklP:
### Problem
Hi,
I discovered a memory leak that seems to be connected to the collision detection pipeline / models.

The scene contains 2 nodes with `MechanicalObject` and `PointCollisionModel`.
I repeatedly load and unload the scene, first with no collision models, then one collision model, and then two collision models.
-> each graph has 3 segments.

```python
import Sofa
import Sofa.Core
import Sofa.Simulation
from enum import Enum


class TestCase(Enum):
    ONE_MODEL = 0
    TWO_MODELS = 1
    NONE = 2


class SimulationHandler:
    def __init__(self):

        self.root_node = Sofa.Core.Node("root")
        self.nodes = createScene(self.root_node, TestCase.NONE)
        Sofa.Simulation.init(self.root_node)

    def step(self):
        Sofa.Simulation.animate(self.root_node, self.root_node.getDt())

    def reload(self, case: TestCase = TestCase.NONE):

        Sofa.Simulation.unload(self.root_node)
        self.root_node = Sofa.Core.Node("root")
        self.nodes = createScene(self.root_node, case)
        Sofa.Simulation.init(self.root_node)


def createScene(root, case: TestCase):
    plugins = [
        "Sofa.Component.Collision.Detection.Algorithm",
        "Sofa.Component.Collision.Detection.Intersection",
    ]
    for plugin in plugins:
        root.addObject("RequiredPlugin", pluginName=plugin, name=plugin)

    root.addObject("DefaultAnimationLoop")
    root.addObject("DefaultPipeline")
    root.addObject("BruteForceBroadPhase")
    root.addObject("BVHNarrowPhase")
    root.addObject("DefaultContactManager")
    root.addObject(
        "LocalMinDistance",
        alarmDistance=5.0,
        contactDistance=0.5,
    )

    node_1 = root.addChild("child_1")
    node_1.addObject("MechanicalObject", position=[0, 0, 0] * 5)

    node_2 = root.addChild("child_2")
    node_2.addObject("MechanicalObject", position=[1, 1, 1] * 5)

    if case == TestCase.ONE_MODEL:
        node_1.addObject("PointCollisionModel")
    elif case == TestCase.TWO_MODELS:
        node_1.addObject("PointCollisionModel")
        node_2.addObject("PointCollisionModel")
    elif case == TestCase.NONE:
        pass

    return node_1, node_2


[...]

PR #3337, opened 2022-09-23
[Core.Collision] Fix memory leak in NarrowPhaseDetection and IntersectorFactory

ScheiklP:
Hi! :)

The original issue and nice chronological story is described here
https://github.com/sofa-framework/sofa/issues/3318

TLDR:
1. The destructor of `NarrowPhaseDetection` should clean up its map of `DetectionOutputVector`s
2. `IntersectorFactory` should also keep track of already created `TIntersectorClass`s and release the old ones.

Cheers,
Paul


Sofa/framework/Core/src/sofa/core/collision/Intersection.h
```diff
@@ -45,6 +45,7 @@ class BaseIntersector
     template<class Model1, class Model2>
     sofa::core::collision::TDetectionOutputVector<Model1,Model2>* createOutputVector(Model1*, Model2*)
     {
+        // NOTE: this raw pointer is stored in NarropPhaseDetection's m_outputsMap and cleaned up when NarropPhaseDetection is destroyed
```
alxbilger:
```suggestion
        // NOTE: this raw pointer is stored in NarrowPhaseDetection's m_outputsMap and cleaned up when NarrowPhaseDetectionis destroyed
```

Sofa/framework/Core/src/sofa/core/collision/NarrowPhaseDetection.h
```diff
@@ -51,7 +51,20 @@ class SOFA_CORE_API NarrowPhaseDetection : virtual public Detection
...
+        for (DetectionOutputMap::iterator it = m_outputsMap.begin(); it != m_outputsMap.end(); it++)
+        {
+            DetectionOutputVector* do_vec = (it->second);
+
+            if (do_vec != nullptr)
+            {
+                do_vec->clear();
+                do_vec->release();
+            }
+            m_outputsMap.erase(it);
```
alxbilger: Warning here. You erase elements of the map while iterating. To check if it is a valid operation.
Since the map is cleared anyway, I recommend to drop this operation.

alxbilger: I also suggest to move the destructor definition in the cpp file

alxbilger (changes requested):

.gitignore
```diff
@@ -34,6 +34,13 @@
 # Python byte code
 *.pyc
 
+# From PrecomputedConstraintCorrection
```
alxbilger: Could you make another PR for this file?

ScheiklP: Sure :)

Sofa/framework/Core/src/sofa/core/collision/IntersectorFactory.h
```diff
@@ -52,6 +54,9 @@ class IntersectorFactory
     typedef BaseIntersectorCreator<TIntersectionClass> Creator;
     typedef std::vector<Creator*> CreatorVector;
     CreatorVector creatorVector;
+    // keep track of already created TIntersectorClass instances for each template combination
+    // when the factory is destroyed, the refcount of the instances is zero and they are cleaned up correctly
+    std::unordered_map<std::type_index, std::shared_ptr<void>> intersectorCache;
```
alxbilger, line 59: Why is it necessary to store the pointer associated to a key, and not simply a list of pointers?

.gitignore
ScheiklP: https://github.com/sofa-framework/sofa/pull/3344

alxbilger: Thanks!

Sofa/framework/Core/src/sofa/core/collision/IntersectorFactory.h
ScheiklP: There should only ever be one instance of the same template combination. So this is basically a hacked version of a singleton. If the factory creates the same type again, the old one is removed.

hugtalbot: another question, why using `std::shared_ptr<void>>` ? would it not work to `std::shared_ptr<TIntersectionClass>>` ?

ScheiklP: Because the Template `TIntersectorClass` is not known in the parent :(

https://github.com/sofa-framework/sofa/blob/926377e46494638bc3e402a8ed7f54986ca1b6c5/Sofa/framework/Core/src/sofa/core/collision/IntersectorFactory.h#L39-L48

alxbilger:
[ci-build][with-all-tests]

alxbilger (approved):
Good job finding this memory leak ;)

PR #3339, opened 2022-09-27
[Mapping] Add checks at init in TopologicalMapping

hugtalbot:
Adding check on pointers and type of input/output topology (mostly input) before using them in TopologicalMapping


Sofa/Component/Topology/Mapping/src/sofa/component/topology/mapping/Edge2QuadTopologicalMapping.cpp
```diff
@@ -97,6 +97,53 @@ void Edge2QuadTopologicalMapping::init()
...
+        // If it is null at this point, it means no input mesh topology could be found.
+        msg_error() << "No input mesh topology found. Consider setting the '" << fromModel.getName() << "' data attribute.";
+        modelsOk = false;
+    }
+    else
+    {
+        // Making sure the input topology is derived from the edge topology container
+        if (!dynamic_cast<container::dynamic::EdgeSetTopologyContainer *>(fromModel.get()))
+        {
+            msg_error() << "The input topology '" << fromModel.getPath() << "' is not a derived class of EdgeSetTopologyContainer. "
```
alxbilger: I am concerned about the term "derived class", which is a programming term whereas you talk to a user that may not be familiar with this term.

You can also improve the message by giving the list of available classes:
```cpp
sofa::core::ObjectFactory::getInstance()->listClassesDerivedFrom<container::dynamic::EdgeSetTopologyContainer>()
```

alxbilger (commented):
Thanks for making this PR. It will improve the UX for sure.

However, I see a lot of duplicated code. I consider it as a technical debt. To discuss if we temporary merge your changes, or think bigger ;)

Sofa/Component/Topology/Mapping/src/sofa/component/topology/mapping/Edge2QuadTopologicalMapping.cpp
hugtalbot: thanks for the comment @alxbilger 
What about the following message:
``` cpp
            msg_error() << "The input topology '" << fromModel.getPath() << "' is not homogeneous with a EdgeSetTopologyContainer. The '" << fromModel.getName() << "' data attribute must be linked to a valid component, among the following list of eligible components:" << msgendl
                        << sofa::core::ObjectFactory::getInstance()->listClassesDerivedFrom<container::dynamic::EdgeSetTopologyContainer>();
```

alxbilger: Very nice!

epernod (commented):
I think there is already some checks in TopologyMapping parse method. Checking fromModel and toModel. 
Might be the good timing to unify it.
Also it is not possible to put the same message in a single method inside TopologyMapping which is called by the different components. 
Using toModel.GetClassName, toModel.getName() etc... ?
Which part of the check is not generic?

Sofa/Component/Topology/Mapping/src/sofa/component/topology/mapping/Edge2QuadTopologicalMapping.cpp
```diff
@@ -97,6 +97,51 @@ void Edge2QuadTopologicalMapping::init()
...
+    bool modelsOk = true;
+
+    // Check input topology
+    if (!fromModel)
+    {
+        // If the input topology link isn't set by the user, the TopologicalMapping::create method tries to find it.
+        // If it is null at this point, it means no input mesh topology could be found.
+        msg_error() << "No input mesh topology found. Consider setting the '" << fromModel.getName() << "' data attribute.";
+        modelsOk = false;
+    }
```
epernod: 2 remarks regarding this check:
- This case is never possible because fromModel is already checked in the mother class TopologicalMapping in the canCreate method. If input and output don't link to valid BaseMeshTopology, the mapping is not create (canCreate return false)
- Hopefully this case is never possible because if fromModel is null, printing error message with fromModel.getName() will crash 😄

Sofa/Component/Topology/Mapping/src/sofa/component/topology/mapping/Tetra2TriangleTopologicalMapping.cpp
```diff
@@ -65,38 +65,54 @@ Tetra2TriangleTopologicalMapping::Tetra2TriangleTopologicalMapping()
...
-        if (!to_tstm) {
-            msg_error() << "No TriangleSetTopologyModifier found in the output topology node '"
-                        << toModel->getContext()->getName() << "'.";
+    else
+    {
+        // Making sure the output topology is derived from the triangle topology container
+        if (!dynamic_cast<container::dynamic::TriangleSetTopologyContainer *>(toModel.get()))
+        {
+            msg_error() << "The input topology '" << toModel.getPath() << "' is not homogeneous with a TriangleSetTopologyContainer. The '" << toModel.getName() << "' data attribute must be linked to a valid component, among the following list of eligible components:" << msgendl
+                                   << sofa::core::ObjectFactory::getInstance()->listClassesDerivedFrom<container::dynamic::TriangleSetTopologyContainer>();
```
epernod: I tested what would be the error message. In this mapping case, it is: 
``` data attribute must be linked to a valid component, among the following list of eligible components: TetrahedronSetTopologyContainer, TriangleSetTopologyContainer```

which doesn't work, because putting a TetrahedronSetTopologyContainer as output will result in a strange behavior. Not sure if it will crash or if it will be like a "TopologicalIdentityMapping" 🤔 

I suggest to move the checks inside a commun method in TopologicalMapping and just test the topologyType of the containters

File: Sofa/Component/Topology/Mapping/src/sofa/component/topology/mapping/Triangle2EdgeTopologicalMapping.cpp [hunk 4]
```diff
@@ -66,25 +66,38 @@ Triangle2EdgeTopologicalMapping::~Triangle2EdgeTopologicalMapping()
...
+    // Check input topology
     if (!fromModel)
     {
         msg_error() << "Pointer to input topology is invalid.";
         modelsOk = false;
     }
+    else
+    {
+        // Making sure the input topology includes triangle elements
+        if (fromModel.get()->getNbTriangles() != 0)
```
epernod: This could still be valid if we start from an empty mesh and we add progressively triangles with a mapping on the edges... btw I thought I had a test on that.

epernod (changes requested):
several changes are not compatible with the current api of the topology and some tests are already present in TopologicalMapping create and canCreate methods.

@hugtalbot if you agree I will push a new proposal directly in this PR.

epernod:
[ci-build][with-all-tests]

Sofa/Component/Topology/Mapping/src/sofa/component/topology/mapping/Edge2QuadTopologicalMapping.cpp
```diff
@@ -97,7 +97,25 @@ void Edge2QuadTopologicalMapping::init()
...
+    // Check input/output topology
+    if (!this->checkTopologyInputTypes()) // method will display error message if false
+    {
+        this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
+        return;
+    }
+
+    
+    // Making sure the output topology is derived from the quad topology container
+    if (!dynamic_cast<container::dynamic::QuadSetTopologyContainer *>(toModel.get()))
```
hugtalbot: a cast, really?

Sofa/Component/Topology/Mapping/src/sofa/component/topology/mapping/Edge2QuadTopologicalMapping.cpp
```diff
@@ -97,7 +97,25 @@ void Edge2QuadTopologicalMapping::init()
     }
 
     unsigned int N = d_nbPointsOnEachCircle.getValue();
+    bool modelsOk = true;
```
hugtalbot: missing definition of `m_inputType` and `m_outputType`

Sofa/framework/Geometry/src/sofa/geometry/ElementType.h
```diff
@@ -43,4 +43,21 @@ enum class ElementType : sofa::Size
...
+    case ElementType::POINT: { return "Point"; }
+    case ElementType::EDGE: { return "Edge"; }
+    case ElementType::TRIANGLE: { return "Triangle"; }
+    case ElementType::QUAD: { return "Quad"; }
+    case ElementType::TETRAHEDRON: { return "Tetrahedron"; }
+    case ElementType::HEXAHEDRON: { return "Hexahedron"; }
+    case ElementType::PENTAHEDRON: { return "Pentahedron"; }
+    case ElementType::PYRAMID: { return "Pyramid"; }
+    default: 
+        return "Unknown";
```
hugtalbot: is there a nicer way to do? @fredroy

epernod:
I moved all changes related to `TopologicalMapping `into the PR #3428 for better clarity.
Therefor this one depends on this new PR.

Sofa/framework/Geometry/src/sofa/geometry/ElementType.h
epernod: To be done in #3428 which is dedicated to changes in TopologicalMapping

hugtalbot: ok so shall we keep it as is @epernod ?

epernod: As you want (we can always upgrade it later when we see the limitation). But to be decided in PR #3428 then I will rebase this one.

epernod:
et 💥 !!

epernod:
yes and it is "funny" because I had the same issue in https://github.com/sofa-framework/sofa/pull/3368
this is just because Grid inherite from MeshTopology which doesn't define the TopoogyType as Hexahedron

epernod:
[ci-build][with-all-tests]

fredroy:
Failing scenes were fixed in #3453 a fortiori, lets merge with master and have a final check

PR #3345, opened 2022-09-28
[Config] Option to enable interprocedural optimization

alxbilger:
IPO option was already available for MSVC. This PR is the CMake way to enable it.
~~However, it does not seem to apply on my setup with MSVC. That's why I kept the `/GL` and `/LTCG` compilation options.~~

### DIFF
--- a/Sofa/framework/Config/CMakeLists.txt
+++ b/Sofa/framework/Config/CMakeLists.txt
@@ -138,7 +138,7 @@ if (IS_IPO_SUPPORTED)
     # Focus on max speed with link-time optimization
     option(SOFA_ENABLE_LINK_TIME_OPTIMIZATION "Enable interprocedural optimization" OFF)
     if (SOFA_ENABLE_LINK_TIME_OPTIMIZATION)
-        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE TRUE PARENT_SCOPE)
     endif()
 else()
     message(STATUS "IPO / LTO not supported: <${IPO_ERROR}>")
@@ -204,11 +204,6 @@ if(WIN32)
 
     if(MSVC_TOOLSET_VERSION GREATER 140) # > VS 2015
         list(APPEND SOFACONFIG_COMPILE_OPTIONS "/Zc:__cplusplus")
-    endif()
-
-    if(SOFA_ENABLE_LINK_TIME_OPTIMIZATION)
-        list(APPEND SOFACONFIG_COMPILE_OPTIONS_RELEASE "/GL")
-        list(APPEND SOFACONFIG_LINK_OPTIONS_RELEASE "/LTCG")
     endif()
 
     # Experimental: compilation with MSVC/Clang-cl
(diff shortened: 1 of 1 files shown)


alxbilger:
[ci-build][with-all-tests]

fredroy:
I tried with the fully updated VS2022 and it seems to work yes.

Sofa/framework/Config/CMakeLists.txt
```diff
@@ -130,6 +130,20 @@ set(SOFACONFIG_LINK_OPTIONS "")
...
 
+include(CheckIPOSupported)
+check_ipo_supported(RESULT IS_IPO_SUPPORTED OUTPUT IPO_ERROR)
+
+## Link-time optimization
+if (IS_IPO_SUPPORTED)
+    # Focus on max speed with link-time optimization
+    option(SOFA_ENABLE_LINK_TIME_OPTIMIZATION "Enable interprocedural optimization" OFF)
+    if (SOFA_ENABLE_LINK_TIME_OPTIMIZATION)
+        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
```
fredroy, line 141:
```suggestion
        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE TRUE)
```

I would set the IPO only for release mode
(thanks to https://github.com/ninja-build/ninja/blob/master/CMakeLists.txt)

fredroy (changes requested):

Sofa/framework/Config/CMakeLists.txt
fredroy:
```suggestion
        set(CMAKE_INTERPROCEDURAL_OPTIMIZATION_RELEASE TRUE PARENT_SCOPE)
```
OK I think it should do the trick
`CMAKE_INTERPROCEDURAL_OPTIMIZATION` value will only be set for the present scope (aka Sofa.Config) ; if we set the additional `PARENT_SCOPE` argument , this option will propagate to the other scopes.
The other method would be to move this option directly to the main CMakeFile.txt at the root of the project.

I just tried on gcc and the flags `flto=auto -fno-fat-lto-objects` are added sucessfully
And I added `_RELEASE` suffix to only enable the LTO for Release mode

fredroy (changes requested):

Sofa/framework/Config/CMakeLists.txt
alxbilger: It makes sense. Thanks.
I'll try on Windows, and check if the flags are still required.

alxbilger: I confirm your suggestion adds the /GL and /LTCG flags.
However, it is available only in release, not in RelWithDebInfo. Would that make sense to add it also for RelWithDebInfo?

fredroy:
[ci-build][with-all-tests]

Sofa/framework/Config/CMakeLists.txt
fredroy: In my mind, one should use LTO only for "production-ready" build so IMO it makes sense to use it only for Release. (I see RelWithDebInfo as a speedy debug build 😅)
But I am not against setting it for RelWithDebInfo as well, especially because IPO is only an option.

fredroy (approved):

PR #3348, opened 2022-09-29
[Topology.Container.Grid] trivial optimization in GridTopology

fredroy:
Just call getValue once and use the const ref afterwards.

This is particularly showing for the function getIndex() which is called a lot of times (really a lot)

Release mode (full optim), MSVC and scene from Components/topology/RegularGridTopology.scn:
before:  `5000 iterations done in 22.5747 s ( 221.487 FPS).`
after  : `5000 iterations done in 20.7355 s ( 241.132 FPS).`
so 10% speedup

But the optimization is more  effective the more complex  the mesh is
(I changed the mesh in the previous scene with a liver with 500K triangles and the speed up was like around 30%)

In any case, it shows that `getValue()` is not really "free" and must be called only once if the value is needed more than one time, especially in the "hot" functions.

### DIFF
--- a/Sofa/Component/Topology/Container/Grid/src/sofa/component/topology/container/grid/GridTopology.cpp
+++ b/Sofa/Component/Topology/Container/Grid/src/sofa/component/topology/container/grid/GridTopology.cpp
@@ -344,14 +344,16 @@ void GridTopology::computePointList()
 
 GridTopology::Index GridTopology::getIndex( int i, int j, int k ) const
 {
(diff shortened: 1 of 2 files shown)


alxbilger:
[ci-build][with-all-tests]

alxbilger (approved):
Good job!

All those functions seems to be called a lot of times. I think we can get even better performances if we cache `d_n` value at a higher level. I'll merge this PR. My suggestion can be done in a second PR.

PR #3360, opened 2022-10-04
[DefaultType] fix rigidcoord compilation

bakpaul:

### DIFF
--- a/Sofa/framework/DefaultType/src/sofa/defaulttype/RigidCoord.h
+++ b/Sofa/framework/DefaultType/src/sofa/defaulttype/RigidCoord.h
@@ -269,7 +269,7 @@ public:
     void toMatrix( Mat& m) const
     {
         m.identity();
-        orientation.toMatrix(m);
+        orientation.toHomogeneousMatrix(m);
         m[0][3] = (typename Mat::Real)center[0];
         m[1][3] = (typename Mat::Real)center[1];
         m[2][3] = (typename Mat::Real)center[2];
--- a/Sofa/framework/DefaultType/src/sofa/defaulttype/RigidCoord.h
+++ b/Sofa/framework/DefaultType/src/sofa/defaulttype/RigidCoord.h
@@ -268,11 +268,22 @@ public:
     template<class Mat>
     void toMatrix( Mat& m) const
     {
+        static_assert(std::is_same_v<Mat, sofa::type::Mat<3,3,real> > || std::is_same_v<Mat, sofa::type::Mat<4,4,real> >, "toMatrix() only supports sofa::type::Mat<3,3,Real> or sofa::type::Mat<4,4,Real>");
+
         m.identity();
-        orientation.toHomogeneousMatrix(m);
-        m[0][3] = (typename Mat::Real)center[0];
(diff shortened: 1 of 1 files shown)


alxbilger (approved):
Thanks for your contribution. I wonder how it was not detected before

fredroy:
I dont think the fix is good ; the "toMatrix()" is available for Mat3x3 but not for Mat4x4 (there was a ` toMatrix`  for Mat4x4 but it was deleted and replace with `toHomogenousMatrix()` ).
This RigidCoord::toMatrix() is templated on the Mat type actually, and is really ill-formed (no assert on the size of the given matrix, etc).
I think the solution is to either:
 - test if the matrix is 3x3 or 4x4 (and failed to compile with a message with a static assert) and calling the respective ` toMatrix()`  or ` toHomogenousMatrix ()` 
 - or remove the template, force it to Mat3x3 and create an other `toHomogenousMatrix` in RigidCoord. Consisntent with Quat but duplicating code and breaking.

I am more for the first solution.

> Thanks for your contribution. I wonder how it was not detected before

I guess this was not detected before because nobody was calling RigidCoord::toMatrix with a Mat4x4 with the code compiled on the CI. Actually, nobody calls RigidCoord::toMatrix, either with mat3 or mat4  apparently.

Sofa/framework/DefaultType/src/sofa/defaulttype/RigidCoord.h
```diff
@@ -269,7 +269,7 @@ class RigidCoord<3,real>
     void toMatrix( Mat& m) const
     {
         m.identity();
-        orientation.toMatrix(m);
+        orientation.toHomogeneousMatrix(m);
```
fredroy:
```suggestion
       if constexpr (std::is_same_v<Mat, sofa::type::Mat<3,3,real>)
       {
            orientation.toMatrix(m);
       }
       else if constexpr (std::is_same_v<Mat, sofa::type::Mat<4,4,real>)
       {
           orientation.toHomogeneousMatrix(m);
       }
       else
       {
           static_assert(false, "toMatrix() only supports sofa::type::Mat<3,3,Real> or sofa::type::Mat<4,4,Real>");
       }
```

YOLO, I did not try to compile it so typos may exist 🤫

fredroy (changes requested):

Sofa/framework/DefaultType/src/sofa/defaulttype/RigidCoord.h
bakpaul: I did the modifications with some fix.

alxbilger (changes requested):
If only 2 Mat types are allowed, I suggest to define 2 different overloads:

```cpp
void toMatrix( sofa::type::Mat<3,3,real>& m) const
{
...
}

void toMatrix( sofa::type::Mat<4,4,real>& m) const
{
...
}
```

Also, check if the 2d rigid needs to be changed as well.

A unit test would be welcomed as well. I think this template function is not used at all in SOFA. That's why this problem was not detected.

alxbilger:
The changes I suggested are here: https://github.com/mimesis-inria/sofa/pull/18

PR #3386, opened 2022-10-13
[all] Implement getModuleComponentList for most modules

alxbilger:

### DIFF
--- a/Sofa/Component/AnimationLoop/src/sofa/component/animationloop/init.cpp
+++ b/Sofa/Component/AnimationLoop/src/sofa/component/animationloop/init.cpp
@@ -58,7 +58,7 @@ void init()
 const char* getModuleComponentList()
 {
     /// string containing the names of the classes provided by the plugin
-    static std::string classes = core::ObjectFactory::getInstance()->listClassesFromTarget(sofa_tostring(MODULE_NAME));
+    static std::string classes = core::ObjectFactory::getInstance()->listClassesFromTarget(MODULE_NAME);
     return classes.c_str();
 }
 } // namespace sofa::component::animationloop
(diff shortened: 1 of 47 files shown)


alxbilger:
[ci-build][with-all-tests]

Sofa/Component/AnimationLoop/src/sofa/component/animationloop/init.cpp
```diff
@@ -54,4 +55,10 @@ void init()
     }
 }
 
+const char* getModuleComponentList()
+{
+    /// string containing the names of the classes provided by the plugin
+    static std::string classes = core::ObjectFactory::getInstance()->listClassesFromTarget(sofa_tostring(SOFA_TARGET));
```
fredroy: I think the constexpr var `MODULE_NAME` (which is defined for each module normally, and in its own namespace) would be better than to use the SOFA_TARGET preproc macro

fredroy (changes requested):

Sofa/Component/AnimationLoop/src/sofa/component/animationloop/init.cpp
alxbilger: you're right. Done

Sofa/Component/AnimationLoop/src/sofa/component/animationloop/init.cpp
```diff
@@ -54,4 +55,10 @@ void init()
     }
 }
 
+const char* getModuleComponentList()
+{
+    /// string containing the names of the classes provided by the plugin
+    static std::string classes = core::ObjectFactory::getInstance()->listClassesFromTarget(sofa_tostring(MODULE_NAME));
```
fredroy: I dont think `sofa_tostring` is still required ?

fredroy (changes requested):

Sofa/Component/AnimationLoop/src/sofa/component/animationloop/init.cpp
alxbilger: you're right

fredroy (approved):

PR #3400, opened 2022-10-18
[SofaDistanceGrid] Set invalid state if cannot load mesh

alxbilger:
Fix a crash at init of example `DistanceGridForceField-liver.scn`, when FlowVR library is not compiled.

### DIFF
--- a/applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/components/forcefield/DistanceGridForceField.inl
+++ b/applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/components/forcefield/DistanceGridForceField.inl
@@ -67,6 +67,12 @@ void DistanceGridForceField<DataTypes>::init()
                                     nx.getValue(),ny.getValue(),nz.getValue(),
                                     box.getValue()[0],box.getValue()[1]);
 
+    if (grid == nullptr)
+    {
+        sofa::core::objectmodel::BaseObject::d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
+        return;
+    }
+
     if (this->stiffnessArea.getValue() != 0 && this->mstate)
     {
         core::topology::BaseMeshTopology* topology = this->getContext()->getMeshTopology();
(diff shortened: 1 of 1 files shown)


applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/components/forcefield/DistanceGridForceField.inl
```diff
@@ -67,6 +67,12 @@ void DistanceGridForceField<DataTypes>::init()
                                     nx.getValue(),ny.getValue(),nz.getValue(),
                                     box.getValue()[0],box.getValue()[1]);
 
+    if (grid == nullptr)
+    {
+        sofa::core::objectmodel::BaseObject::d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
+        return;
```
epernod, line 74: maybe adding a msg_Error() to warn user the Distance grid has not been created and why.

epernod (changes requested):

applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/components/forcefield/DistanceGridForceField.inl
alxbilger: It is a relevant suggestion. I added a generic message, and the reason is already the load function: https://github.com/sofa-framework/sofa/blob/6dbc39db7f7ba999783dee6d281fa2f76334ee7c/applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/DistanceGrid.cpp#L121

epernod (approved):

epernod:
You don't need to add checks, later in the code for: 
d_componentState.getValue() == sofa::core::objectmodel::ComponentState::Valid 

to avoid crash?

alxbilger:
> You don't need to add checks, later in the code for: d_componentState.getValue() == sofa::core::objectmodel::ComponentState::Valid
> 
> to avoid crash?

I only have a partial answer:
I tested one time step and it works. I suppose it is because the initialization defines the topology. If the topology is empty, I guess that the later functions just don't iterate over the elements. But it's just a guess.

epernod:
ok, as long as it doesn't crash and the user has error message. 👍

alxbilger:
[ci-build][with-all-tests]

PR #792, opened 2018-10-09
[MultiThreading] Fix Livers scenes crash 

fspadoni:
CHANGE: AdvanceTimer::Id::IdFactory is no more a singleton static class but  a thread local class to prevent different threads to share the same instance of the class.

FIX: DefaultTaskScheduler registration to TaskScheduler factory is forced to be executed after the TaskScheduler static initialization.

______________________________________________________

This PR: 
- [ ] builds with SUCCESS for all platforms on the CI.
- [ ] does not generate new warnings.
- [ ] does not generate new unit test failures.
- [ ] does not generate new scene test failures.
- [ ] does not break API compatibility.
- [ ] is more than 1 week old (or has fast-merge label).

**Reviewers will merge only if all these checks are true.**

PR #3422, opened 2022-10-24
[Helper] Portable thread local storage duration

alxbilger:
Since C++11, it is possible to declare a variable with one copy per thread with the keyword `thread_local`.

The consequence is that the code in `thread_specific_ptr.h` could be deprecated as it is not used anywhere else (not in this PR).

Bonus: remove a memory leak.

### DIFF
--- a/Sofa/framework/Helper/src/sofa/helper/AdvancedTimer.h
+++ b/Sofa/framework/Helper/src/sofa/helper/AdvancedTimer.h
@@ -216,12 +216,8 @@ public:
             /// return the instance of the factory. Creates it if doesn't exist yet.
             static IdFactory& getInstance()
             {
-                SOFA_THREAD_SPECIFIC_PTR(IdFactory, instance);
-                if (instance == nullptr)
-                {
-                    instance = new IdFactory;
-                }
-                return *instance;
+                static thread_local IdFactory factory;
+                return factory;
             }
         };
 


alxbilger:
[ci-build][with-scene-tests]

alxbilger:
[ci-build][with-scene-tests][force-full-build]

alxbilger:
Indeed. It would be interesting to know why. But I can see that the `thread_local` version did not stay long (4 hours). From https://github.com/sofa-framework/sofa/pull/792, I can see that the compilation failed when `thread_local` was introduced. And the compilation log is no longer available. But you can see now that the compilation succeeds. I guess that at the time the compilation failed because of old version of the compilers. And it is no longer a problem now.

Sofa/framework/Helper/src/sofa/helper/AdvancedTimer.h
```diff
@@ -216,12 +216,8 @@ class SOFA_HELPER_API AdvancedTimer
             /// return the instance of the factory. Creates it if doesn't exist yet.
             static IdFactory& getInstance()
             {
-                SOFA_THREAD_SPECIFIC_PTR(IdFactory, instance);
-                if (instance == nullptr)
-                {
-                    instance = new IdFactory;
-                }
-                return *instance;
+                static thread_local IdFactory factory;
```
hugtalbot, line 219: A priori (Francois J.) since C+11 a thread-safe instantiation is done when initializing a `static` variable
While here using `thread_local`, it creates one static variable per thread

```suggestion
                static IdFactory factory;
```

hugtalbot: AdvancedTimer may not itself be threadsafe

alxbilger: > A priori (Francois J.) since C+11 a thread-safe instantiation is done when initializing a `static` variable While here using `thread_local`, it creates one static variable per thread

Yes I agree, but this is just the initial intention. I did not modify any behavior. My goal was just to replace a complex per-thread variable declaration by a portable one. Nothing else.

> AdvancedTimer may not itself be threadsafe

Probably, but it is out of the scope of this pull request.


To be clear: I don't have the ambition to fix anything in `AdvancedTimer`, or change any behavior, in this pull request. Just "decomplexify" the code.

Issue #3499, opened 2022-11-21
Segfault in SparseLDLSolver


ScheiklP:
### Problem
Hi!
There seems to be a check missing that prevents `SparseLDLSolver` trying to solve an empty system.
When I create an object with n points, I can remove n-1 and get a segfault on the nth.

```
########## SIG 8 - SIGFPE: arithmetic exception, such as divide by zero ##########
  sofa::helper::BackTrace::sig(int)
  libmetis__FM_2WayCutRefine
  libmetis__RandomBisection
  libmetis__InitSeparator
  libmetis__MlevelNodeBisectionL1
  libmetis__MlevelNestedDissection
  METIS_NodeND
  void sofa::component::linearsolver::direct::SparseLDLSolverImpl<sofa::linearalgebra::CompressedRowSparseMatrix<double, sofa::type::vector<double, sofa::type::CPUMemoryManager<double> >, sofa::type::vector<unsigned int, sofa::type::CPUMemoryManager<unsigned int> > >, sofa::linearalgebra::FullVector<double>, sofa::component::linearsolver::NoThreadManager>::factorize<sofa::type::vector<int, sofa::type::CPUMemoryManager<int> >, sofa::type::vector<double, sofa::type::CPUMemoryManager<double> > >(int, int*, int*, double*, sofa::component::linearsolver::direct::SparseLDLImplInvertData<sofa::type::vector<int, sofa::type::CPUMemoryManager<int> >, sofa::type::vector<double, sofa::type::CPUMemoryManager<double> > >*)
  sofa::component::linearsolver::direct::SparseLDLSolver<sofa::linearalgebra::CompressedRowSparseMatrix<double, sofa::type::vector<double, sofa::type::CPUMemoryManager<double> >, sofa::type::vector<unsigned int, sofa::type::CPUMemoryManager<unsigned int> > >, sofa::linearalgebra::FullVector<double>, sofa::component::linearsolver::NoThreadManager>::factorize(sofa::linearalgebra::CompressedRowSparseMatrix<double, sofa::type::vector<double, sofa::type::CPUMemoryManager<double> >, sofa::type::vector<unsigned int, sofa::type::CPUMemoryManager<unsigned int> > >&, sofa::component::linearsolver::direct::SparseLDLImplInvertData<sofa::type::vector<int, sofa::type::CPUMemoryManager<int> >, sofa::type::vector<double, sofa::type::CPUMemoryManager<double> > >*)
[...]

PR #3500, opened 2022-11-22
[Solver.Direct] Unit tests on empty system

alxbilger:
Related to https://github.com/sofa-framework/sofa/issues/3499

Unit tests are added on empty systems

### DIFF
--- a/Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
+++ b/Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
@@ -22,6 +22,9 @@
 #include <sofa/testing/BaseTest.h>
 #include <sofa/component/linearsolver/direct/SparseLDLSolver.h>
 #include <sofa/component/linearsolver/direct/SparseCommon.h>
+#include <sofa/simulation/Node.h>
+#include <sofa/simulation/graph/DAGSimulation.h>
+#include <sofa/simulation/graph/SimpleApi.h>
 
 #include <sofa/testing/NumericTest.h>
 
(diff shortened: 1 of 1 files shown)


Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
```diff
@@ -178,3 +181,50 @@ TEST(SparseLDLSolver, MatrixFactorization)
...
+    using MatrixType = sofa::linearalgebra::CompressedRowSparseMatrix<SReal>;
+
+    MatrixType matrix;
+    matrix.resize(0, 0);
+    matrix.compress();
+
+    using Solver = sofa::component::linearsolver::direct::SparseLDLSolver<MatrixType, sofa::linearalgebra::FullVector<SReal> >;
+    Solver::SPtr solver = sofa::core::objectmodel::New<Solver>();
+
+    solver->init();
```
epernod, line 199: don't you think the init method should also send a warning? no mechanicalObject, empty sytem detected?

Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
```diff
@@ -178,3 +181,50 @@ TEST(SparseLDLSolver, MatrixFactorization)
...
+
+    EXPECT_MSG_EMIT(Warning);
+    solver->invert(matrix);
+}
+
+TEST(SparseLDLSolver, EmptyMState)
+{
+    // required to be able to use EXPECT_MSG_NOEMIT and EXPECT_MSG_EMIT
+    sofa::helper::logging::MessageDispatcher::addHandler(sofa::testing::MainGtestMessageHandler::getInstance() ) ;
```
epernod, line 208: you need this even if you are using simpleAPI ?

Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
```diff
@@ -178,3 +181,50 @@ TEST(SparseLDLSolver, MatrixFactorization)
...
+    sofa::simpleapi::importPlugin("Sofa.Component.StateContainer");
+
+    sofa::simpleapi::createObject(root, "DefaultAnimationLoop");
+    sofa::simpleapi::createObject(root, "EulerImplicitSolver");
+    sofa::simpleapi::createObject(root, "SparseLDLSolver", {{"template", "CompressedRowSparseMatrixd"}});
+    sofa::simpleapi::createObject(root, "MechanicalObject", {{"template", "Vec3"}, {"position", ""}});
+
+    sofa::simulation::getSimulation()->init(root.get());
+
+    {
```
epernod, line 224: why in a specific scope ?

alxbilger: Because of `EXPECT_MSG_EMIT(Warning);` which uses RAII. The specific scope isolates the warning emission to only what is inside the scope.

Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
alxbilger: I don't see how it is related

epernod: it is just that I never had to add manually the MessageDispatcher in the tests I wrote. So I'm wondering if the simpleapi is doing it maybe through the createNewGraph 
I will check, if I have time.

alxbilger: Actually you don't have to do that if you have a test deriving from `BaseTest`

Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
alxbilger: Fair question. This check is performed at each time step. It visits the mechanical objects and compute the size of the system. I don't know if this can be done in the init method if the mechanical objects are not initialized yet.
Also, the API allows to use a Linear Solver on a matrix, without defining a scene (see the test `TEST(SparseLDLSolver, MatrixFactorization)`). Adding this check in the init method would not make sense in this case.
To be discussed

Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
epernod: ok, thanks for the explanation

Sofa/Component/LinearSolver/Direct/tests/SparseLDLSolver_test.cpp
epernod: ok I move that discussion in #3504 to not block this PR.

epernod (approved):

PR #3505, opened 2022-11-25
[VectorTypeInfo] Change default handling of data buffer

Camille-K:
In the `getValuePtr` methods, a buffer on vector data is obtained by actually returning a reference to the vector first element. However the case where the method is called on an empty (but valid) vector is not handled. This may lead to an execution error, for instance when using SofaPython3 to assign values to a component's Data dynamically. This specific scenario is the object of a second PR in SofaPython3 [(link to the PR)](https://github.com/sofa-framework/SofaPython3/pull/318).

To change this, we propose to use the `std::vector.data()` method, which handles the case when the vector is actually empty (however returning a pointer that may or may not be null).

All unit tests work on my setup with this modification. However I don't fully understand the consequences of a change at this depth in SOFA's core, so additional opinions would be useful.

### DIFF
--- a/Sofa/framework/DefaultType/src/sofa/defaulttype/typeinfo/TypeInfo_Text.h
+++ b/Sofa/framework/DefaultType/src/sofa/defaulttype/typeinfo/TypeInfo_Text.h
@@ -32,8 +32,8 @@ struct DataTypeInfo<std::string> : public TextTypeInfo<std::string>
     static const std::string name() { return "string"; }
     static const std::string GetTypeName(){ return "string"; }
 
-    static const void* getValuePtr(const std::string& data) { return &data[0]; }
-    static void* getValuePtr(std::string& data) { return &data[0]; }
+    static const void* getValuePtr(const std::string& data) { return data.data(); }
+    static void* getValuePtr(std::string& data) { return data.data(); }
 };
 
 } /// namespace sofa::defaulttype
--- a/Sofa/framework/DefaultType/src/sofa/defaulttype/typeinfo/models/VectorTypeInfo.h
+++ b/Sofa/framework/DefaultType/src/sofa/defaulttype/typeinfo/models/VectorTypeInfo.h
@@ -193,12 +193,12 @@ struct VectorTypeInfo
 
     static const void* getValuePtr(const DataType& data)
     {
-        return &data[0];
+        return data.data();
     }
 
     static void* getValuePtr(DataType& data)
     {
-        return &data[0];
+        return data.data();
     }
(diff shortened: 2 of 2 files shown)


fredroy:
[ci-build][with-all-tests]

alxbilger:
The problem with this function is that you cannot trust the returned value forever due to pointer invalidation.

`std::vector` allocates dynamically the memory depending of the need. At creation, it allocates a small chunk of memory. If the memory is full when you try to insert an element, it will allocate a bigger chunk of memory and move everything in this new memory space. Everything that was in the previous memory space is invalidated. It stills points to something, but it is invalid. And using it leads to a crash.

I made a small example on godbolt of this situation: https://godbolt.org/z/aqec8Krzo

You can see that the result of `data()` is invalidated after elements insertions.

Conclusion: your solution is better than the previous. But this function must be used appropriately.

Camille-K:
Thanks @alxbilger for the explanation and the perfect illustration :+1:
I understand the issue, although I don't know if the "appropriate use" of this method (`getValuePtr`) should be left to the user, or if additional safeguards should be put in the code.

I did a quick search in the code of SOFA + a few plugins that I'm using, and I found a call to this particular method only in the `DataHelper` component in SofaPython3. I added a quick safeguard in this case, in the [PR](https://github.com/sofa-framework/SofaPython3/pull/318) I'm referring to in the description above. However the modification I suggested doesn't solve the type of problem you are describing here (so it would be left to the user to check data buffer coherence whenever the underlying data is modified...).

Apart from that, I saw that the initial code I replaced (`return &data[0]`) is also used in two others `getValuePtr` methods : in `FixedArrayTypeInfo.h` and `TypeInfo_Text.h`. As I didn't find any other call to these methods, I would say this doesn't change much, but maybe the PR could be extended to these as well. I can do the modification if someone can confirm it is a good idea.

alxbilger:
It's definitively a good idea to also replace the `&data[0]` calls by `.data()`.

Camille-K:
I did the modification in `TypeInfo_Text.h`, but build fails if I extend it to `FixedArrayTypeInfo.h`, apparently because of the template specialisations on Rigid types, for instance :
```
FixedArrayTypeInfo.h:195: error: ‘const DataType’ {aka ‘const class sofa::defaulttype::RigidDeriv<2, double>’} has no member named ‘data’
  195 |         return data.data();
      |                ~~~~~^~~~
```
(which makes sense).

fredroy (approved):

alxbilger (approved):

PR #3519, opened 2022-12-01
[SofaPhysicsAPI] Some small cleaning and add access to VisualModel*

epernod:

### DIFF
--- a/applications/projects/SofaPhysicsAPI/src/SofaPhysicsAPI/SofaPhysicsAPI.h
+++ b/applications/projects/SofaPhysicsAPI/src/SofaPhysicsAPI/SofaPhysicsAPI.h
@@ -76,6 +76,9 @@ public:
     /// Return the number of currently active output meshes
     unsigned int            getNbOutputMeshes();
 
+    SofaPhysicsOutputMesh* getOutputMeshPtr(unsigned int meshID);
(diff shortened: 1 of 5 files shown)


alxbilger (changes requested):
The methods you added can be const.

And by curiosity, why are the methods of the SofaPhysicsAPI class not defined in SofaPhysicAPI.cpp?

epernod:
> And by curiosity, why are the methods of the SofaPhysicsAPI class not defined in SofaPhysicAPI.cpp?

Very good question... I'm not sure, I think it was done so SofaPhysicsAPI can be shipped/share easily? or to look like an interface ?
@fredroy any idea?

fredroy:
> > And by curiosity, why are the methods of the SofaPhysicsAPI class not defined in SofaPhysicAPI.cpp?
> 
> Very good question... I'm not sure, I think it was done so SofaPhysicsAPI can be shipped/share easily? or to look like an interface ? @fredroy any idea?

At that time, it was to give a public API to any external program using SOFA as a lib (no exposition whatsoever to any SOFA code).
You just needed this .h and the compiled libs.
 
Note that because of the config.h stuff, this is not the case anymore 🙊

hugtalbot (approved):

fredroy (approved):

PR #3529, opened 2022-12-05
[LinearSolver.Direct] Segfault in SparseLDLSolver due to empty system matrix

ScheiklP:
Hi all,
As discussed with @alxbilger in https://github.com/sofa-framework/sofa/issues/3499

Add a check that catches an empty system to solve. This might happen due to topological changes.

Thanks again for the help, @alxbilger ! :)

Cheers,
Paul

### DIFF
--- a/Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolver.inl
+++ b/Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolver.inl
@@ -86,6 +86,7 @@ bool SparseLDLSolver<TMatrix, TVector, TThreadManager>::factorize(
 
     if (n == 0)
     {
+        msg_warning() << "Invalid Linear System to solve. Please insure that there is enough constraints (not rank deficient)." ;
         return true;
     }
 
--- a/Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolver.inl
+++ b/Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolver.inl
@@ -84,6 +84,11 @@ bool SparseLDLSolver<TMatrix, TVector, TThreadManager>::factorize(
 
     int n = M.colSize();
 
+    if (n == 0)
+    {
+        return true;
+    }
+
     int * M_colptr = (int *)Mfiltered.getRowBegin().data();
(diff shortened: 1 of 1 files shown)


Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolver.inl
```diff
@@ -84,6 +84,11 @@ bool SparseLDLSolver<TMatrix, TVector, TThreadManager>::factorize(
 
     int n = M.colSize();
 
+    if (n == 0)
+    {
+        return true;
```
alxbilger, line 90: My only concern is on the returned value. I don't know what this function must return. And the returned value is not used anywhere in SOFA.

alxbilger:
Note that this PR should fix the unit tests introduced in https://github.com/sofa-framework/sofa/pull/3501

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolver.inl
ScheiklP: That would be another PR, no? :D 
Right now, the function also returns `true` in the other check a few lines down.

The long term proper solution would be changing `invert` from a `void` to `bool` return type and then checking that in `solveSystem` for all `linearsolver`s, right?

alxbilger: I agree that it is not in the scope of this PR

hugtalbot (approved):

alxbilger:
The branch is already up-to-date. The failing unit tests are legit. They have to be fixed

ScheiklP:
Meh... it was just a missing warning that the test expected. :D 
I wanted to leave out the warning to not flood the output, once all points are gone. :P

hugtalbot (approved):

alxbilger:
@ScheiklP your case (topological changes) where you have `n==0` seems to be legit. So I would not emit a warning. The unit tests are probably to be changed

ScheiklP:
@alxbilger is that `TEST(SparseLDLSolver, EmptyMState)` and `TEST(SparseLDLSolver, EmptySystem)`?
Just removing the `EXPECT_MSG_EMIT(Warning);` is probably also not what we want, right?

PR #3548, opened 2022-12-15
[Core] Introduce parallel for each

alxbilger:
The idea is to provide tools to write multithreaded code more efficiently. The functions introduced in this pull request allow to apply a function to a range in parallel.
Unit tests are provided.
The newly added functions are used as example in two places:
- GenericConstraintSolver: the code is unique for both sequential and parallel execution
- ParallelHexahedronFEMForceField: no ambiguity: the code is only for parallel execution. But nothing prevent to have a code supporting both sequential and parallel execution. But the component is not totally ready for a merge with HexahedronFEMForceField. Note that it avoids a lot of boilerplate code, hence reducing the complexity of the file.

I think this will allow us to write more multithreaded code.

In the back, the parallel loop uses a task scheduler and divides the number of elements in `n` ranges, where `n` corresponds to the number of threads in the task scheduler. This is to execute exactly `n` tasks, instead of `m`, if `m` is the number of elements. It avoids the overhead of creating a task for each element. It is more efficient when the number of elements is large.

Benchmarks are yet to perform.

### DIFF
--- a/Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
+++ b/Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
@@ -32,6 +32,7 @@
 #include <sofa/simulation/DefaultTaskScheduler.h>
 #include <sofa/helper/ScopedAdvancedTimer.h>
 #include <sofa/simulation/MainTaskSchedulerFactory.h>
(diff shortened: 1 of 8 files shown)


fredroy:
[ci-build][with-all-tests]

fredroy:
I tried this feature with BeamAdapter, replacing the parallel std::for_each() on https://github.com/sofa-framework/BeamAdapter/pull/50
On a custom scene, the performance is the same, which is a good news by itself. The regression test did not report any discrepancy either.
And at least there is no special case (compiler not supporting std::execution, etc) to handle 👍

fredroy (approved):

PR #3550, opened 2022-12-16
[GUI] GUI libraries as plugins

fredroy:
This was preliminary introduced in SofaGLFW/SofaImgui, where a new GUI is loaded at run-time
Contrary to the batch/qt/headlessrecorder GUIs which are forcefully linked at compile-time through SofaGui, which could depend (or not) on Qt, etc according to the CMake config,

SOFANG works already made all the different GUIs as independent libraries but ultimately the SofaGui was still present because of the fact that runSofa was still relying on this mechanism.

Now :
- Qt/Batch/HeadLessRecorder are loadable plugins and register themselves in the GUIManager,  if loaded .
- Batch is still forcefully loaded by runSofa ; otherwise it could be possible to run SOFA without any GUI (I am not against having this possibility tho)
- SofaGui is officially deprecated 🎇

Remarks: 
- after this PR, I would really like to move HeadlessRecorder in applications/plugins (either that in the "official" SOFA trunk) as it is only supported on Linux, needs X libs dev and I dont even know if people really uses it.
- showing help will only show the help for the runSofa options not the ones introduced by the eventual plugins....


fredroy:
[ci-build][with-all-tests]

PR #3566, opened 2023-01-04
[MultiThreading] Implement domain decomposition for a lock-free parallelism

alxbilger:
Based on https://github.com/sofa-framework/sofa/pull/3548

In the `addDForce` method, the parallel algorithm consists in computing a thread-specific `dF` and then combine them all in the main `dF`. It requires a synchronisation mechanism as it may create race conditions.

This PR introduce an alternative method which is lock-free. It divides the model into smaller subdomains that can be solved in parallel without synchronization.

Since the domains are visited sequentially (it's the elements of each domain that are visited in parallel), it comes with an overhead. Therefore, this method is particularly suited for large models where the overhead can be amortized.

I have the following results on the provided example:

```
1000 steps 
17775 elements (hexa)

domain decomposition:
35.1703 s ( 28.4331 FPS).

naive:
76.5666 s ( 13.0605 FPS)

sequential:
192.688 s ( 5.18974 FPS).
```

That's right: 17k elements simulated in real-time on a laptop CPU

As a comparison with the same scene where everything is on GPU, not only the `addDForce` method of `HexahedronFEMForceField`, but also the other components:
```
21.7437 s ( 45.9904 FPS).
```
with a RTX A3000 Laptop GPU

### DIFF
--- a/Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
+++ b/Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
@@ -32,6 +32,7 @@
 #include <sofa/simulation/DefaultTaskScheduler.h>
 #include <sofa/helper/ScopedAdvancedTimer.h>
 #include <sofa/simulation/MainTaskSchedulerFactory.h>
+#include <sofa/simulation/ParallelForEach.h>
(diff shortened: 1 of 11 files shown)


alxbilger:
[ci-build][with-all-tests]

alxbilger:
This implementation is at the force field level because its goal is to parallelize the matrix-free CG.
Unfortunately, the benchmarks were measured with a bug. Now the bug is fixed, the performances are disappointing. It requires more investigation.

PR #3597, opened 2023-01-19
[Simulation] SceneCheck can be added in plugins

alxbilger:
SceneChecks were isolated in a project. But we may want that plugins add their own SceneCheck. An example with MultiThreading is provided here.

The class `SceneCheck` is now available in Sofa.Simulation.Core. Tools to register a SceneCheck are also available in the module.

The newly introduced SceneCheck checks if there are opportunities to use a multi-threaded component. It advices the user with a message looking like:

```
[SUGGESTION] [SceneCheckMultithreading] This scene is using components implemented sequentially while a parallel implementation is available. Using the parallel implementation may improve the performances. Here is the list of sequen
tial components in your scene and their parallel equivalent:
        BVHNarrowPhase -> ParallelBVHNarrowPhase
        BruteForceBroadPhase -> ParallelBruteForceBroadPhase
        HexahedronFEMForceField[Vec3d] -> ParallelHexahedronFEMForceField[Vec3d]
```


alxbilger:
[ci-build][with-all-tests]

PR #3601, opened 2023-01-24
[SofaCUDA] Reorganize following SOFA structure (1/n)

alxbilger:
Start with mappings

image

With a compatibility layer on headers.


alxbilger:
[ci-build][with-all-tests]

hugtalbot (approved):

alxbilger:
Retrospectively, I would prefer the following structure:
```
SofaCUDA
| - component
      | - mapping
            | - linear
            | - nonlinear
```

I find that the intermediate directories `sofa/gpu/cuda` are superfluous, and redundant with the name of the plugin. It is also in the prevision of changing the namespaces (which is not done in this PR).
What do you think @hugtalbot @epernod @damienmarchal @fredroy ?

epernod:
I agree. Sofa/gpu/cuda only make sens if we have sofa/gpu/opencl  or sofa/multithread....  but it would be in other plugin. So if you keep the SofaCuda as first level in your namespace/hierarchy, it will work.

alxbilger:
> I agree. Sofa/gpu/cuda only make sens if we have sofa/gpu/opencl or sofa/multithread.... but it would be in other plugin. So if you keep the SofaCuda as first level in your namespace/hierarchy, it will work.

It's also what I think. Thanks @epernod

alxbilger:
This is the new structure:

image

It uses the usual structure from any module: `moduleName/src/moduleName`

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaBaseVector.cu
```diff
@@ -19,8 +19,8 @@
 *                                                                             *
 * Contact information: contact@sofa-framework.org                             *
 ******************************************************************************/
-#include "CudaCommon.h"
-#include "CudaMath.h"
+#include <sofa/gpu/cuda/CudaCommon.h>
```
epernod, line 22: Wouldn't be worth it to start with those guys ?
SofaCUDA/type/CudaCommon.h
or
SofaCUDA/helper/CudaCommon

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaBaseVector.cu
```diff
@@ -19,8 +19,8 @@
...
+#include <sofa/gpu/cuda/CudaMath.h>
```
epernod, line 23: SofaCUDA/helper/CudaMath.h

epernod, line 23: SofaCUDA/helper/CudaMath.h

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaBaseVector.cu
epernod, line 22: Wouldn't be worth it to start with those guys ?
SofaCUDA/type/CudaCommon.h
or
SofaCUDA/helper/CudaCommon

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaBoxROI.cpp
```diff
@@ -19,7 +19,7 @@
 *                                                                             *
 * Contact information: contact@sofa-framework.org                             *
 ******************************************************************************/
-#include "CudaTypes.h"
+#include <sofa/gpu/cuda/CudaTypes.h>
```
epernod, line 22: Same SofaCuda/type/CudaTypes

might be a big mess!

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaBaseVector.cu
alxbilger: It is also in my plan to move those files. I started by the components because it is the easiest IMO. And also because I did not anticipate the newest structure (the one with `SofaCUDA/src/SofaCUDA`.

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaBoxROI.cpp
fredroy: if we want really to be consistent with SOFA, it is SofaCuda/defaulttype/CudaTypes (like sofa/defaulltype/VecTypes, etc)

applications/plugins/SofaCUDA/src/SofaCUDA/component/mapping/nonlinear/CudaRigidMapping.inl
```diff
@@ -0,0 +1,234 @@
...
+*                                                                             *
+* Contact information: contact@sofa-framework.org                             *
+******************************************************************************/
+#pragma once
+
+#include "CudaRigidMapping.h"
+#include <sofa/component/mapping/nonlinear/RigidMapping.inl>
+#include <sofa/helper/accessor.h>
+
+namespace sofa::gpu::cuda
```
hugtalbot, line 28: namespace seems still to be the old one, should we keep simply a SofaCUDA namespace ?

(just a comment let's agree that it does not matter much!)

alxbilger: Yes, actually I did not change the namespaces. I figured it was too many changes in a single PR. I agree that it could be changed, but I'll do it in a future PR. Would that work for you?

PR #3640, opened 2023-02-10
[SofaCUDA] Fix compilation with CUDA12

fredroy:
CUDA12 disabled the usage of a deprecated texture API, obviously used in our code 😅

Actually only TLED classes used necessarily texturing so no choice other than disable the compilation of these 2 classes if CUDA>=v12 at the cmake stage.

Other files were including CudaTexture.h and were using the old API in code only activated with Macros (which I suspect, were never enabled anymore....) so I enclosed the include with the same `#ifdef` tests.

### DIFF
--- a/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaSpringForceField.cu
+++ b/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaSpringForceField.cu
@@ -22,7 +22,10 @@
 #include <sofa/gpu/cuda/CudaCommon.h>
 #include <sofa/gpu/cuda/CudaMath.h>
 #include "cuda.h"
+
+#ifdef USE_TEXTURE
 #include "CudaTexture.h"
+#endif
 
 #if defined(__cplusplus) && CUDA_VERSION < 2000
 namespace sofa
--- a/applications/plugins/SofaCUDA/CMakeLists.txt
+++ b/applications/plugins/SofaCUDA/CMakeLists.txt
@@ -60,14 +60,13 @@ set(HEADER_FILES
     ### FEM
     sofa/gpu/cuda/CudaHexahedronFEMForceField.h
     sofa/gpu/cuda/CudaHexahedronFEMForceField.inl
-    sofa/gpu/cuda/CudaHexahedronTLEDForceField.h
+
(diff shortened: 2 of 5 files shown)


fredroy:
[ci-build][with-all-tests]

applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaSpringForceField.cu
```diff
@@ -22,7 +22,10 @@
 #include <sofa/gpu/cuda/CudaCommon.h>
 #include <sofa/gpu/cuda/CudaMath.h>
 #include "cuda.h"
+
+#ifdef USE_TEXTURE
```
hugtalbot, line 26: I did not know this compilation variable but shouldn t it be 

``` suggestion
#ifdef USE_TEXTURE_X
```

fredroy: Nope it is `USE_TEXTURE`, see https://github.com/sofa-framework/sofa/blob/8b8038c76e5cf7695d74e000938790ddb618e36a/applications/plugins/SofaCUDA/sofa/gpu/cuda/CudaSpringForceField.cu#L257

alxbilger:
@fredroy I worked on this problem too, but my approach was to refactor texture references by using texture objects. It is not trivial. It's not working yet. I'll try again in the week, otherwise I suggest to simply remove the usage of textures. I'd use simply raw GPU pointers. This alternative is easier than the texture objects approach (but it may cost a bit of performances (perhaps negligible)).
My point is, don't merge this PR right now, unless is super urgent.

fredroy:
> @fredroy I worked on this problem too, but my approach was to refactor texture references by using texture objects. It is not trivial. It's not working yet. I'll try again in the week, otherwise I suggest to simply remove the usage of textures. I'd use simply raw GPU pointers. This alternative is easier than the texture objects approach (but it may cost a bit of performances (perhaps negligible)). My point is, don't merge this PR right now, unless is super urgent.

Of course, it would be much better to effectively fix the problem than to ignore it. This PR was just providing a quick but brainless solution for a (hopefully) temporary fix.

fredroy:
Superseded by #3650

PR #3643, opened 2023-02-13
[Simulation.Core] Fix memory leak when deleting a Task

fredroy:
TODO

### DIFF
--- a/Sofa/framework/Simulation/Core/src/sofa/simulation/WorkerThread.cpp
+++ b/Sofa/framework/Simulation/Core/src/sofa/simulation/WorkerThread.cpp
@@ -157,8 +157,8 @@ void WorkerThread::runTask(Task *task)
         {
             // pooled memory: call destructor and free
             //task->~Task();
-            task->operator delete(task, sizeof(*task));
-            //delete task;
+            //task->operator delete(task, sizeof(*task));
+            delete task;
         }
     }
 


fredroy:
[ci-build][with-all-tests]

alxbilger (commented):
I don't understand the differences here.

fredroy:
> I don't understand the differences here.

That's the problem, this is supposed to do the same thing : either calling the delete operator with size or not size, because in this end both of them call the same code in the StdTaskAllocator https://github.com/sofa-framework/sofa/blob/d855365c71b4b83190ca4af3ce412bd4dbdc1b2f/Sofa/framework/Simulation/Core/src/sofa/simulation/DefaultTaskScheduler.cpp#L47

But in the end calling the version without size fixes the leak 😶‍🌫️
caduceus.scn with 1000steps: 
Before: (operator delete with size)
```
Direct leak of 63936 byte(s) in 999 object(s) allocated from:
    #0 0x7fdeaeac0488 in operator new(unsigned long) ../../../../src/libsanitizer/asan/asan_new_delete.cpp:95
    #1 0x7fdea4253138 in _M_create<sofa::component::animationloop::FreeMotionAnimationLoop::FreeMotionAndCollisionDetection(const sofa::core::ExecParams*, const sofa::core::ConstraintParams&, SReal, sofa::core::MultiVecId, sofa::core::MultiVecId, sofa::core::MultiVecDerivId, sofa::simulation::common::MechanicalOperations*)::<lambda()> > /usr/include/c++/12/bits/std_function.h:161
    #2 0x7fdea4253138 in _M_init_functor<const sofa::component::animationloop::FreeMotionAnimationLoop::FreeMotionAndCollisionDetection(const sofa::core::ExecParams*, const sofa::core::ConstraintParams&, SReal, sofa::core::MultiVecId, sofa::core::MultiVecId, sofa::core::MultiVecDerivId, sofa::simulation::common::MechanicalOperations*)::<lambda()>&> /usr/include/c++/12/bits/std_function.h:215
    #3 0x7fdea4253138 in _M_manager /usr/include/c++/12/bits/std_function.h:198
    #4 0x7fdea4253138 in _M_manager /usr/include/c++/12/bits/std_function.h:282
    #5 0x7fdea81a4486 in std::function<void ()>::function(std::function<void ()> const&) /usr/include/c++/12/bits/std_function.h:391
    #6 0x7fdea81a4486 in sofa::simulation::TaskScheduler::addTask
<lots and lots of other leaks...>
...
SUMMARY: AddressSanitizer: 808967 byte(s) leaked in 15938 allocation(s).
```
After:  (operator delete without size)
```
<no more Tasks leak>
SUMMARY: AddressSanitizer: 744967 byte(s) leaked in 14938 allocation(s).
```

And I don't understand why 😅

My primary intention was that  (without looking at the StdTaskAllocator, I noticed that calling the operator delete with sizeof(Task) was not correct as the pointer to Task is pointed actually to a CallableTask (which is bigger in memory) thus it was supposed to free less memory (as sizeof(Task) < sizeof(CallableTask). But after digging into the code, I saw that both of the operator delete are supposed to call the same code. I even checked with the debugger that, indeed both of them reach the function in StdMemoryAllocator...
So I guess some other things happen here but I cannot explain why 🤐

alxbilger:
Ok thanks for the explanation. Is there any information in the git history?

bakpaul:
Hey @fredroy @alxbilger, what about this pr ?

fredroy:
> Hey @fredroy @alxbilger, what about this pr ?

I did not really investigate for the moment so I guess I will close it and maybe come back on it later

PR #3666, opened 2023-02-20
[Helper,Geometry] Move proximity classes into free functions

alxbilger:
The code in `proximity.h` has several problems:

- The classes names are misleading. It makes think that it computes the distance, but it computes the closest points.
- Classes with a single function that could be static. It could be a free function.
- Difficult to find this file in Helper. To me, it makes more sense to put it in Geometry.
- Types mismatch

I separated the 3 functions in dedicated files.
I deprecated the classes and the header.

Note: I don't think the LCP approach is the fastest one to compute closest points. 
Not2: `Mat_solve_LCP.h` would deserve a bit of cleaning

### DIFF
--- a/Sofa/framework/Geometry/src/sofa/geometry/proximity/PointTriangle.h
+++ b/Sofa/framework/Geometry/src/sofa/geometry/proximity/PointTriangle.h
@@ -1,4 +1,4 @@
-/******************************************************************************
+﻿/******************************************************************************
 *                 SOFA, Simulation Open-Framework Architecture                *
 *                    (c) 2006 INRIA, USTL, UJF, CNRS, MGH                     *
 *                                                                             *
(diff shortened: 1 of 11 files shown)


alxbilger:
[ci-build][with-all-tests]

Sofa/framework/Geometry/src/sofa/geometry/proximity/PointTriangle.h
```diff
@@ -0,0 +1,77 @@
...
+
+namespace sofa::geometry::proximity
+{
+
+template<typename Node,
+     typename T = std::decay_t<decltype(*std::begin(std::declval<Node>()))>,
+     typename = std::enable_if_t<std::is_scalar_v<T>>
+>
+[[nodiscard]]
+constexpr bool computeClosestPointOnTriangleToPoint(
```
epernod, line 38: Why a new file and not adding a method directly in geometry::triangle::computeClosestPointToPoint ?

epernod (commented):
I trust you on the maths... even if there is no test? 😆

epernod (approved):

PR #3687, opened 2023-02-27
[Collision] Make the (CubeModel) BoundingTree deterministic

fredroy:
"Solve" the discrepancy of the regression test of MeshSpringForceField.
Sorting the bboxes of the different levels needs a predicate (a boolean function) to compare two bboxes
https://github.com/fredroy/sofa/blob/master/Sofa/Component/Collision/Geometry/src/sofa/component/collision/geometry/CubeModel.h#L72

But it can happen that v1 == v2 ; and std::sort does not guarantee stability in these cases.
I suppose the result will depend of the implementation of the STL so it would explain the difference between msvc, gcc, clang/lin and clang/mac.

By using stable_sort, we can get stability and thus the same results between the different platforms.
Questions are:
- stable_sort is a bit slower than sort https://stackoverflow.com/a/34668459 so set the stability as an option?
- the stability itself seems only relevant is some cases (scenes using penalities?)

And this will certainly need to regenerate some regression files... (at least MeshSpringForceField)

EDIT:
[ci-depends-on https://github.com/sofa-framework/Regression/pull/39]

### DIFF
--- a/Sofa/Component/Collision/Geometry/src/sofa/component/collision/geometry/CubeModel.cpp
+++ b/Sofa/Component/Collision/Geometry/src/sofa/component/collision/geometry/CubeModel.cpp
@@ -303,7 +303,7 @@ void CubeCollisionModel::computeBoundingTree(int maxDepth)
 
                     // Separate cells on each side of the median cell
                     CubeSortPredicate sortpred(splitAxis);
-                    std::sort(elems.begin() + subcells.first.getIndex(), elems.begin() + subcells.second.getIndex(), sortpred);
+                    std::stable_sort(elems.begin() + subcells.first.getIndex(), elems.begin() + subcells.second.getIndex(), sortpred);
 
                     // Create the two new subcells
                     Cube cmiddle(this, middle);


fredroy:
[ci-build][with-all-tests]

alxbilger:
Congratulations @fredroy for finding this problem.

If I understand correctly, the problem comes from the fact that two equivalent cubes (in terms of the comparison predicate) are not in the same order depending on the platform (compiler). In my opinion, it creates collision responses that are also ordered differently. And the order in collision responses is important (see https://github.com/sofa-framework/sofa/pull/2067).

But I also wonder how two cubes can be equivalent. We compare two floating-point types, more specifically, the center of a cube along the largest dimension. I suspect that both cubes are perfectly aligned in an orthogonal dimension. It is probably due to the construction of the scene (torus are aligned along the X axis), or the symmetry of the geometry. @fredroy do you know when the discrepancy happens? Is it at the very beginning?

If we don't want to use `stable_sort` because it is slower, maybe we can make a smarter comparison predicate? For example, if the largest dimension is equivalent, let's compare the second largest dimension etc.

fredroy:
> If I understand correctly, the problem comes from the fact that two equivalent cubes (in terms of the comparison predicate) are not in the same order depending on the platform (compiler). In my opinion, it creates collision responses that are also ordered differently. And the order in collision responses is important (see #2067).

Yes, but a small precision in this case the determinism is set across the different compilers (and more precisely the different implementations of the STL). A priori the same compiler/libstd will create the same result.

> But I also wonder how two cubes can be equivalent. We compare two floating-point types, more specifically, the center of a cube along the largest dimension. I suspect that both cubes are perfectly aligned in an orthogonal dimension. It is probably due to the construction of the scene (torus are aligned along the X axis), or the symmetry of the geometry. @fredroy do you know when the discrepancy happens? Is it at the very beginning?

Yes, at the very beginning.
Here is a dump of the container being sorted (left linux/gcc and right windows/msvc), after running 1 step ; showing { minBBox maxBBox } for each element

diff_gcc_msvc

It is trying to sort according to the axis X and you can see the highlighted lines have the same X

> 
> If we don't want to use `stable_sort` because it is slower, maybe we can make a smarter comparison predicate? For example, if the largest dimension is equivalent, let's compare the second largest dimension etc.

Yes, it could be an other solution, but in this end, does `sort() + multiple_test_axis_predicate < stable_sort() + oneaxis_predicate` ? not so sure and it would need more benchmarks. And I am sure it will also depend on the data itself...

fredroy:

alxbilger:
ZERO regressions! 🎉

alxbilger:
[ci-build]

PR #3691, opened 2023-02-27
[Spring] Check for nullptr in RestShapeSpringsForceField

alxbilger:

### DIFF
--- a/Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/PolynomialRestShapeSpringsForceField.inl
+++ b/Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/PolynomialRestShapeSpringsForceField.inl
@@ -170,9 +170,17 @@ void PolynomialRestShapeSpringsForceField<DataTypes>::recomputeIndices()
 
         if (m_useRestMState)
         {
-            for (sofa::Index i = 0; i < getExtPosition()->getValue().size(); i++)
+            if (const DataVecCoord* extPosition = getExtPosition())
             {
-                m_ext_indices.push_back(i);
+                const auto& extPositionValue = extPosition->getValue();
+                for (sofa::Index i = 0; i < extPositionValue.size(); i++)
+                {
+                    m_ext_indices.push_back(i);
+                }
+            }
+            else
+            {
+                this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
             }
         }
         else
@@ -186,7 +194,7 @@ void PolynomialRestShapeSpringsForceField<DataTypes>::recomputeIndices()
 
     if (m_indices.size() > m_ext_indices.size())
     {
-        msg_error() << "Error : the dimention of the source and the targeted points are different ";
+        msg_error() << "Error : the dimension of the source and the targeted points are different ";
         m_indices.clear();
         m_ext_indices.clear();
     }
@@ -227,6 +235,7 @@ void PolynomialRestShapeSpringsForceField<DataTypes>::addForce(const core::Mecha
     const DataVecCoord* extPosition = getExtPosition();
     if (!extPosition)
     {
+        this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
         return;
     }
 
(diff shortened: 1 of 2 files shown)


alxbilger:
[ci-build][with-all-tests]

Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/PolynomialRestShapeSpringsForceField.inl
```diff
@@ -209,7 +223,14 @@ void PolynomialRestShapeSpringsForceField<DataTypes>::addForce(const core::Mecha
 
     helper::WriteAccessor<DataVecDeriv> f1 = f;
     helper::ReadAccessor<DataVecCoord> p1 = x;
-    helper::ReadAccessor<DataVecCoord> p0 = *getExtPosition();
+
```
fredroy, line 234: is it a "probable" case that extPosition is nullptr ? In my mind, it means there is no d_restMState if m_useRestMState or mstate (if  !m_useRestMState ).
Should this test be done at the init() and set the component state at invalid instead ? (and replace the others if(!extPosition) with asserts)

alxbilger: I think there is already the check in `init()`. Perhaps the component state is not correctly set (to check). But I am not fully confident in relying on the component state. Today, there is no early return based on an invalid state. To discuss if it could be generalized. And the component state value must be changed depending on the change of some Data. Which is not currently the case.
To conclude, I am not against working with component state. But:
- It is complex to handle all cases
- Need to be generalized
- I prefer to take extra measures to prevent crashes. It means checking for `nullptr` even if the component state check passed.

epernod (approved):

Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/PolynomialRestShapeSpringsForceField.inl
```diff
@@ -372,7 +402,14 @@ void PolynomialRestShapeSpringsForceField<DataTypes>::draw(const core::visual::V
     if (!vparams->displayFlags().getShowForceFields() || !d_drawSpring.getValue())
         return;
 
-    helper::ReadAccessor< DataVecCoord > p0 = *getExtPosition();
+    const DataVecCoord* extPosition = getExtPosition();
+    if (!extPosition)
+    {
+        this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
```
epernod, line 408: I'm not sure changing the componentState inside the draw method is a good idea. BTW the draw method is not const?
In your case I suppose the componentState would already be Invalid and in an idyllic world, the draw method should not be called on a not "valid" component
```suggestion
```

Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/RestShapeSpringsForceField.inl
```diff
@@ -427,7 +465,14 @@ void RestShapeSpringsForceField<DataTypes>::draw(const VisualParams *vparams)
     const auto stateLifeCycle = vparams->drawTool()->makeStateLifeCycle();
     vparams->drawTool()->setLightingEnabled(false);
 
-    ReadAccessor< DataVecCoord > p0 = *getExtPosition();
+    const DataVecCoord* extPosition = getExtPosition();
+    if (!extPosition)
+    {
+        this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Invalid);
```
epernod, line 471:
```suggestion
```

same, not in draw method.

epernod (changes requested):

Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/PolynomialRestShapeSpringsForceField.inl
alxbilger: I confirm the `draw` method is not `const`.

I am not sure about your comment: I don't think we should rely only on the component state. I think that with Python everything is possible and we should not exclude that this pointer becomes `nullptr` even if the component state is valid.

PR #3707, opened 2023-03-07
[Mapping.NonLinear] Move DistanceMultiMapping in its own files

alxbilger:
It was initially defined in the same files than DistanceMapping


alxbilger:
[ci-build][with-all-tests]

Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/DistanceMapping.h
```diff
@@ -24,12 +24,12 @@
 #include <sofa/component/mapping/nonlinear/config.h>
 
 #include <sofa/core/Mapping.h>
-#include <sofa/core/MultiMapping.h>
 #include <sofa/linearalgebra/EigenSparseMatrix.h>
 #include <sofa/core/topology/BaseMeshTopology.h>
 #include <sofa/type/Mat.h>
 #include <sofa/type/Vec.h>
 #include <sofa/type/RGBAColor.h>
+#include <sofa/component/mapping/nonlinear/DistanceMultiMapping.h>
```
fredroy, line 34: Maybe add a comment to remove this header in v23.12, as it was to keep a compat with the previous behavior (i.e being able to use DistanceMultiMapping by including DistanceMapping.h)

fredroy (approved):

Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/DistanceMapping.h
alxbilger: I was thinking about it, but what does it change to remove it in v23.12 or now? In both cases, the user is not noticed (except with a compilation error). And I don't see how it is possible to warn the user if `DistanceMultiMapping` is used only via the include of `DistanceMapping`. What do you think?

fredroy: Ah yes I totally agree with you on the fact it is quite impossible to warn the user, I was suggesting to add a dev comment (as with `//` or `/* */` ) for the next time someone stumbled on this inclusion (and add a timeline for @hugtalbot when he will do the next deprecation round in the future 😁)

PR #3737, opened 2023-03-21
[Helper] Improve OptionsGroup

alxbilger:
- Clean the files
- Add simple tests
- Add constructor for initializer lists. Allow to write concise object creation
- Allow builder pattern through the `setSelectedItem` methods

### DIFF
--- a/Sofa/framework/Helper/src/sofa/helper/OptionsGroup.cpp
+++ b/Sofa/framework/Helper/src/sofa/helper/OptionsGroup.cpp
@@ -23,10 +23,7 @@
 #include <cstdlib>
 #include <sofa/helper/logging/Messaging.h>
 
-namespace sofa
-{
-
-namespace helper
+namespace sofa::helper
 {
 
 class OptionsGroup;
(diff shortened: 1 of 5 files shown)


alxbilger:
[ci-build][with-all-tests]

alxbilger:
The unit tests failure came from the fact that the order of the files changed in the CMakeLists.txt. It highlights the facts that `EXPECT_MSG_EMIT` works only if `MainGtestMessageHandler` is added to the `MessageDispatcher`. But `logging_test.cpp` calls `clearHandlers`. Since `logging_test.cpp` is now sorted before `FileSystem_test.cpp`, it clears the message handler required by `FileSystem_test.cpp`. The message handler must be added in any case, and not relying on the fact that it might have been added earlier.

fredroy:
> The unit tests failure came from the fact that the order of the files changed in the CMakeLists.txt. It highlights the facts that `EXPECT_MSG_EMIT` works only if `MainGtestMessageHandler` is added to the `MessageDispatcher`. But `logging_test.cpp` calls `clearHandlers`. Since `logging_test.cpp` is now sorted before `FileSystem_test.cpp`, it clears the message handler required by `FileSystem_test.cpp`. The message handler must be added in any case, and not relying on the fact that it might have been added earlier.

Well done!
I have added the `breaking` flag due to the change of some signatures for some methods

fredroy (approved):

PR #3740, opened 2023-03-21
[Mapping.NonLinear] Geometric stiffness method in an OptionsGroup

alxbilger:
Based on #3737 

Almost all nonlinear mappings have a Data to select how to compute the geometric stiffness (none, exact, stabilized).
The Data was a `int`. It is now replaced by a ` OptionsGroup`. It is therefore breaking because of the change of type.

Minor changes go along with this PR.

Note: for some reasons, RigidMapping is the only mapping which default value for geometric stiffness is not 2 (stabilized). It is 0 (no geometric stiffness computed).


alxbilger:
[ci-build][with-all-tests]

PR #3749, opened 2023-03-27
[GUI.Qt] Register meta type to fix asynchronous Qt call

### LINKED ISSUE #3610: Log mutex locked twice (leading to crash)
I encountered a situation where my code emitted a `dmsg_warning`. It triggers a Qt event to react, and change the component icon. However, something went wrong in the Qt side and a Qt error was emitted. Since we listen to Qt errors, it also emits a `msg_*`.
Therefore, a `msg_*`  is called inside a `dmsg_warning` call. The problem is that a mutex protects each call when entering the message functions (`PUBLIC_API_ENTRY_POINT_MUTEX`). And if a mutex is locked twice in the same thread, it crashes.
To solve this situation, a `std::recursive_mutex` would help. But I think there is a design problem in the first place. To be discussed...

alxbilger:
This is a fix for a very specific situation. Here is what happens:

1. A `msg_warning` is called in a component from a thread different from the main thread. It could be also `msg_error` or `msg_info`, as long as the message count of the element changes. It could be a change of name as well.
2. The `ObjectStateListener` associated to the component updates the associated `QTreeWidgetItem` (this class is not a `QObject`).
3. The icon is updated, so it warns the `QTreeWidget` (this class is a QObject) through a signal.
4. Since the call happens in a thread different from the main thread, the queued connection is used. For that, all signals/slots parameters must be known to Qt's meta-object system. This PR registers the class `QVector<int>`, so it can be used in the queued connection.

It partially fixes https://github.com/sofa-framework/sofa/issues/3610. The warning is no longer triggered by Qt. Therefore, the mutex is locked only once in one thread. I say it solves partially, because any other Qt error would trigger the same crash.

### DIFF
--- a/Sofa/GUI/Qt/CMakeLists.txt
+++ b/Sofa/GUI/Qt/CMakeLists.txt
@@ -132,6 +132,7 @@ set(MOC_HEADER_FILES
     ${SRC_ROOT}/QSofaListView.h
     ${SRC_ROOT}/QSofaStatWidget.h
     ${SRC_ROOT}/QTabulationModifyObject.h
+    ${SRC_ROOT}/QtMessageRedirection.h
     ${SRC_ROOT}/QTransformationWidget.h
     ${SRC_ROOT}/RealGUI.h
     ${SRC_ROOT}/SimpleDataWidget.h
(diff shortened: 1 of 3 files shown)


alxbilger:
[ci-build][with-all-tests]

fredroy (approved):

PR #3761, opened 2023-03-31
[Mapping.NonLinear] Fix SquareMapping applyDJT

alxbilger:
In `applyDJT`, the following product is computed:

$$
b \mathrel{{+}{=}} h^2 \frac{\partial \mathbf{J}}{\partial x_{\text{out}}} f_{\text{out}} v_\text{in}
$$

with `b` being the right side of the main equation, `h` the time step size ($h^2$ is the `kFactor`), `x` the position, `v` the velocity and `f` the force. Subscript `in` refers to the parent of the mapping, and `out` to the child. $J$ is the jacobian matrix of the mapping.

The method in `SquareMapping` had two ways to compute this product:

1) Using K which has been computed previsously. And multiply it by $v_\text{in}$
2) Using $f_{\text{out}}$ and $v_\text{in}$ only, without using a cached matrix

Only the first method was tested in unit tests. I forced the second method in a unit test by introducing a Data `useGeometricStiffnessMatrix`. It could be observed that the second method failed. It needed a fix. The product with $v_\text{in}$ was forgotten. It is now added in this pull request.

### DIFF
--- a/Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/SquareMapping.h
+++ b/Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/SquareMapping.h
@@ -87,6 +87,8 @@ public:
     void updateK( const core::MechanicalParams* mparams, core::ConstMultiVecDerivId childForce ) override;
     const linearalgebra::BaseMatrix* getK() override;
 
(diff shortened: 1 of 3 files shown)


alxbilger:
[ci-build][with-all-tests]

hugtalbot (approved):

PR #3768, opened 2023-04-06
[examples] Introduce example and tests for SquareMapping

alxbilger:
An example is introduced for SquareMapping. In this scene, the goal was to verify that two methods are equivalent:
1) A SquaredDistanceMapping is used to transform the DOFs from the 3d space to a 1d space (representing the squared distances between the DOFs). Then a spring acts in the 1d space
2) A combination of two mappings: DistanceMapping and SquareMapping. It also transforms the 3d space to a 1d space where a spring is added.

A unit test is added to check both methods are equivalent. However, we can observe that if the simulation runs long enough, both objects diverge. It will be a challenge to find the reasons.


alxbilger:
[ci-build][with-all-tests]

hugtalbot (approved):

PR #3805, opened 2023-04-20
[Visual] VisualModelImpl: Clean and optimize slightly ComputeNormals()

fredroy:
- clean (const, foreach loop)
- optimize a little bit when initializing the normals vector
   - if there is not index, use memset to clear all vectors at once
   - if there are indices, just dont clear the new vectors as there are supposed to be 0 already.

On a big mesh (315k vertices and 620k triangles) and mapped (to force the call to computeNormals)
 - before  `1000 iterations done in 14.95 s ( 66.8896 FPS).`
 - after `1000 iterations done in 14.075 s ( 71.048 FPS)`

The case is not really usual as it needs huge meshes to see the difference ; and mappings will certainly takes much more time

### DIFF
--- a/Sofa/Component/Visual/src/sofa/component/visual/VisualModelImpl.cpp
+++ b/Sofa/Component/Visual/src/sofa/component/visual/VisualModelImpl.cpp
@@ -1005,15 +1005,8 @@ void VisualModelImpl::computeNormals()
     }
     else
     {
-        vector<Coord> normals;
-        std::size_t nbn = 0;
-        for (std::size_t i = 0; i < vertNormIdx.size(); i++)
-        {
-            if (vertNormIdx[i] >= nbn)
-                nbn = vertNormIdx[i]+1;
-        }
-
-        normals.resize(nbn); // resize() will call the default ctor, which initializes with zeros
+        const std::size_t nbn = static_cast<std::size_t>(*std::max_element(vertNormIdx.begin(), vertNormIdx.end())) + 1;
+        sofa::type::vector<Coord> normals(nbn); // will call the default ctor, which initializes with zeros
 
         for (const auto& triangle : triangles)
         {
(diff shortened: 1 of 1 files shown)


Sofa/Component/Visual/src/sofa/component/visual/VisualModelImpl.cpp
```diff
@@ -1083,53 +1079,48 @@ void VisualModelImpl::computeNormals()
                 nbn = vertNormIdx[i]+1;
         }
 
-        normals.resize(nbn);
-        for (std::size_t i = 0; i < nbn; i++)
-            normals[i].clear();
+        normals.resize(nbn); // resize() will call the default ctor, which initializes with zeros
```
alxbilger: I suggest the following code:
```cpp
const std::size_t nbn = *std::max_element(vertNormIdx.begin(), vertNormIdx.end()) + 1;
sofa::type::vector<Coord> normals(nbn);
```
That would avoid the previous loop.

But I also wonder if simply an `unordered_map<visual_index_type, Coord>` would be more performant. It would avoid allocating too much memory. It would avoid to normalize all the elements of the vector, only the elements in the map.

fredroy: > But I also wonder if simply an `unordered_map<visual_index_type, Coord>` would be more performant. It would avoid allocating too much memory. It would avoid to normalize all the elements of the vector, only the elements in the map.

I guess it depends of the size of the diff between nbn and vertices.size(). 
If the diff is not too important, I think the fact that everything is contiguous will be faster (even if there is duplicate computation and memory) than the unordered_map insertion/iterations. So I would say "it depends" 😅

hugtalbot (approved):

epernod (approved):

PR #3834, opened 2023-05-03
[LinearAlgebra] Change access specifier for the method set

bakpaul:
Change access specifier for the method set. A public access is needed in SofaCUDASolver plugin. 

Moreover, regardless of the reasons of this modification, the fact that this method is at the beginning of the class declaration without any access specifier before is not wanted. Even if it is a class which makes its access private by default, it lacks of readability.  

### DIFF
--- a/Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/EigenBaseSparseMatrix.h
+++ b/Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/EigenBaseSparseMatrix.h
@@ -59,17 +59,6 @@ class EigenBaseSparseMatrix : public linearalgebra::BaseMatrix
 
 public:
 
-    void set(Index i, Index j, double v) override
-    {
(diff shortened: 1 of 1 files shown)


alxbilger (approved):
According to the git history, the method `set` was initially disabled for this class. That is why it was private. See the commit that implemented the `set` method: https://github.com/sofa-framework/sofa/commit/7a3ec34975ff0cf10b5ff328fe7ecda60cafcb0a#diff-0dff698f57bb671243ce9650f2cc92c0a8a2b7ee53dc9af31a0d60ec6c314716

Since it is now implemented, it makes sense to make it public, just as in the base class

fredroy:
Instead of moving the `public:` keyword , I would rather move the set() function in the other public section below  with the other methods 😅

alxbilger (approved):

alxbilger:
[ci-build][with-all-tests]

alxbilger (approved):

PR #3838, opened 2023-05-05
[Mapping.NonLinear] Warns when non-symmetric matrix is produced

alxbilger:
Warns the user when a non-symmetric matrix is produced by a non-linear mapping while a symmetric linear solver is used. The message looks like:
```
[ERROR]   [RigidMapping(RigidMapping1)] The geometric stiffness of this mapping is a non-symmetric matrix. It means a linear solver supporting non-symmetric matrices must be used, but it is not the case here. To fix your scene, you have two options: 1) Use a linear solver supporting non-symmetric matrices, 2) stabilize the geometric stiffness with the Data 'geometricStiffness' set to 'Stabilized'
```

Note that the detection was already considered in the past through a `bool` in the `MechanicalParams`, but not used at all.

Note that even if a linear system supporting non-symmetric matrices is used, the exact geometric stiffness computed in `RigidMapping` leads to large stability issues in `examples/Component/LinearSystem/MatrixLinearSystem.scn`

### DIFF
--- a/Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/NonLinearMappingData.h
+++ b/Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/NonLinearMappingData.h
@@ -22,6 +22,7 @@
 #pragma once
 
 #include <sofa/component/mapping/nonlinear/config.h>
+#include <sofa/core/MechanicalParams.h>
 #include <sofa/core/objectmodel/BaseObject.h>
 #include <sofa/helper/OptionsGroup.h>
 
@@ -35,6 +36,10 @@ public:
     Data<helper::OptionsGroup> d_geometricStiffness; ///< Method used to compute the geometric stiffness
 
     NonLinearMappingData();
+
+protected:
+
+    void checkLinearSolverSymmetry(const core::MechanicalParams* mparams) const;
 };
 
 template <bool HasStabilizedGeometricStiffness>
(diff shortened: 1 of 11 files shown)


Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/NonLinearMappingData.h
```diff
@@ -60,4 +65,34 @@ inline NonLinearMappingData<false>::NonLinearMappingData()
...
+            ss << "To fix your scene, use a linear solver supporting non-symmetric matrices";
+        }
+        else
+        {
+            ss << "To fix your scene, you have two options: 1) Use a linear solver "
+                  "supporting non-symmetric matrices, 2) stabilize the geometric stiffness with "
+                  "the Data '" + d_geometricStiffness.getName() + "' set to '" + stabilizedName + "'";
+        }
+        msg_error() << ss.str();
+    }
```
hugtalbot, line 96:
```suggestion
    if (mparams)
    {
        if (mparams->symmetricMatrix())
        {
            std::stringstream ss;
            ss << "The geometric stiffness of this mapping is a non-symmetric matrix. "
                "It means a linear solver supporting non-symmetric matrices must be used, but it is not"
                " the case here. ";
    
            const std::string stabilizedName = "Stabilized";
            if (d_geometricStiffness.getValue().isInOptionsList(stabilizedName) == -1)
            {
                if constexpr (HasStabilizedGeometricStiffness)
                {
                    dmsg_fatal() << "The option '" << stabilizedName << "' is not available in the Data '"
                       << d_geometricStiffness.getName() << "'";
                }
                ss << "To fix your scene, use a linear solver supporting non-symmetric matrices";
            }
            else
            {
                ss << "To fix your scene, you have two options: 1) Use a linear solver "
                      "supporting non-symmetric matrices, 2) stabilize the geometric stiffness with "
                      "the Data '" + d_geometricStiffness.getName() + "' set to '" + stabilizedName + "'";
            }
            msg_error() << ss.str();
        }
        else
        {
            msg_info() << "Linear solver does support 
        }
    }
```

Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/NonLinearMappingData.h
```diff
@@ -60,4 +65,34 @@ inline NonLinearMappingData<false>::NonLinearMappingData()
 )
 {}
 
+template <bool HasStabilizedGeometricStiffness>
+void NonLinearMappingData<HasStabilizedGeometricStiffness>::checkLinearSolverSymmetry(
+    const core::MechanicalParams* mparams) const
+{
+    if (mparams && mparams->symmetricMatrix())
```
hugtalbot, line 72: If would also create the case if !mparams->symmetricMatrix() with a msg_info

Sofa/framework/Simulation/Core/src/sofa/simulation/MechanicalOperations.cpp
```diff
@@ -451,6 +451,7 @@ void MechanicalOperations::m_setSystemMBKMatrix(SReal mFact, SReal bFact, SReal
     mparams.setMFactor(mFact);
     mparams.setBFactor(bFact);
     mparams.setKFactor(kFact);
+    mparams.setSymmetricMatrix(!s->supportNonSymmetricSystem());
```
hugtalbot, line 454: I am getting confused each time I face the variable `mparams->symmetricMatrix` to get back to what it actually means.
Shouldn't we rename it `mparams->supportOnlySymmetricMatrix` ?

Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/NonLinearMappingData.h
alxbilger: Why? The case `!mparams->symmetricMatrix()` is the regular code path. A `msg_info` would trigger every time step

Sofa/framework/Simulation/Core/src/sofa/simulation/MechanicalOperations.cpp
alxbilger: I agree, I also have this feeling

alxbilger:
[ci-build][with-all-tests]

PR #3822, opened 2023-04-25
[Rendering3D] OglModel: Use glMapBufferRange to update buffers

fredroy:
Based on #3815 and #3805 

... instead of glBufferSubData() as it is a faster method to upload stuff on the GPU directly

Also:
 - dont update indices (lines, triangles, quads) if there are not modified
 - dont copy to temporary float if already in float

(same mesh as #3805)
Before:
`1000 iterations done in 14.075 s ( 71.048 FPS)` 

After:
`1000 iterations done in 9.597 s ( 104.199 FPS)`

After (float)
` 1000 iterations done in 7.618 s ( 131.268 FPS)`

Based on :
 - #3815 

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #3841, opened 2023-05-08
[Rendering3D] OglModel: Revert back to glBufferSubData

fredroy:
Partial revert of #3822 where the usage of glBufferSubData  is reinstated; because glMapBufferRange is a OpenGL3 function and it does not behave well with macos and ogl compat mode (reports of crashes)

`1000 iterations done in 11.859 s ( 84.3241 FPS)`

A bit worse than with glMapBufferRange but still better than before #3822


alxbilger (approved):
Tested on MacOS, it works

PR #3888, opened 2023-05-25
[Constraint] Add tag to classify constraints

bakpaul:
Added constraint tag to be able to classify them and treat them differently in the solvers.
Example of use : Being able to filter the bilateral constraint to solve them directly.


alxbilger (commented):
I think `override` is redundant in `override final`

alxbilger:
[ci-build][with-all-tests]

alxbilger (commented):
I am disturbed by the fact that you need to call `d_identifiers.setValue(getIdentifiers());` in all constructors. I think it is easy to forget if you implement a new kind of constraint.

Since calling a virtual function in the base class is unsafe, I suggest to call it in the `init` function. You would need to call `d_identifiers.setValue(getIdentifiers());` only in `BaseConstraint::init`.

PR #3977, opened 2023-06-26
[Core] Suggestion of a better displayFlag name : ContactInteractions

hugtalbot:
So far, we used the "Interactions" DisplayFlag in the GUI (actually named as "InteractionForceFields" in the list:
1. it's confusing
2. this name is :question: $ :exclamation: :boom: :bird: 

Here is a new suggestion: :white_flower: "ContactInteractions"  :white_flower: 
Obviously, compat and generalization in the code will be done if we agree


alxbilger:
I agree with your analysis, but I don't with your suggestion. For example, `Contact` does not work for `StiffSpringForceField`.

Issue #3967, opened 2023-06-23
Implementation of new API since refactoring of matrix assembly


alxbilger:
Due to #2777, several methods must be implemented to replace the old API:

# Force fields

The new functions are `buildStiffnessMatrix` (replaces `addKToMatrix`) and `buildDampingMatrix` (replaces `addBToMatrix`)

- [x] AngularSpringForceField (#3993)
- [x] BeamFEMForceField (#2777)
- [x] ConicalForceField (#3997)
- [x] ConstantForceField (#2777)
- [x] CudaHexahedronTLEDForceField (#4000)
- [x] CudaTetrahedronTLEDForceField (#4000)
- [x] DiagonalVelocityDampingForceField (#3999)
- [x] EdgePressureForceField (#4004)
- [x] EllipsoidForceField (#4005)
- [x] FastTetrahedralCorotationalForceField (#3929)
- [x] FastTriangularBendingSprings (#4006)
- [x] HexahedralFEMForceField (#3969)
- [x] HexahedronFEMForceField (#2777)
- [x] LinearForceField (#3973)
- [x] MappingGeometricStiffnessForceField (#2777)
- [x] OscillatingTorsionPressureForceField (#4007)
- [x] PlaneForceField (#3972)
- [x] PolynomialRestShapeSpringsForceField (#4009)
- [x] QuadBendingFEMForceField (#4015)
- [x] QuadPressureForceField (#4018)
- [x] RestShapeSpringsForceField (#2777)
- [x] SphereForceField (#4099)
- [x] StandardTetrahedralFEMForceField (#4110)
- [x] SurfacePressureForceField (#4097)
- [x] TaitSurfacePressureForceField (#4116)
- [x] TetrahedralCorotationalFEMForceField (#3981)
- [x] TetrahedralTensorMassForceField (#4127)
- [x] TetrahedronDiffusionFEMForceField (#2777 and #4012)
- [x] TetrahedronFEMForceField (#2777)
- [x] TetrahedronHyperelasticityFEMForceField (#2777)
- [x] TorsionForceField (#4115)
- [x] TriangleFEMForceField (#2777)
- [x] TriangularFEMForceField (#3991)
- [x] TriangularFEMForceFieldOptim (#4105)
- [ ] ParticlesRepulsionForceField
- [ ] TrianglePressureForceField
- [ ] SPHFluidForceField
- [ ] QuadularBendingSprings
- [ ] TriangularAnisotropicFEMForceField
[...]

PR #4005, opened 2023-07-04
[MechanicalLoad] Implement buildStiffnessMatrix in EllipsoidForceField

alxbilger:
A task from #3967 + minor cleaning.

I suspect there is also a damping term, which was not taken into account in the derivative before. It needs a deeper investigation.


fredroy:
[ci-build][with-all-tests]

Sofa/Component/MechanicalLoad/src/sofa/component/mechanicalload/EllipsoidForceField.h
```diff
@@ -82,17 +82,25 @@ class EllipsoidForceField : public core::behavior::ForceField<DataTypes>
 
     };
 
-    Data<sofa::type::vector<Contact> > contacts; ///< Contacts
+    SOFA_ELLIPSOIDFORCEFIELD_RENAMEDDATA_DEPRECATED() Data<sofa::type::vector<Contact> > contacts; ///< Contacts
```
fredroy: I guess `SOFA_ELLIPSOIDFORCEFIELD_CONTACTS_DEPRECATED` was the first version of the macro and you change it for `SOFA_ELLIPSOIDFORCEFIELD_RENAMEDDATA_DEPRECATED`. But you forgot to update the config.h.in 🤐

Sofa/Component/MechanicalLoad/src/sofa/component/mechanicalload/EllipsoidForceField.h
```diff
@@ -82,17 +82,25 @@ class EllipsoidForceField : public core::behavior::ForceField<DataTypes>
...
-    Data<Coord> center; ///< ellipsoid center
-    Data<Coord> vradius; ///< ellipsoid radius
-    Data<Real> stiffness; ///< force stiffness (positive to repulse outward, negative inward)
-    Data<Real> damping; ///< force damping
-    Data<sofa::type::RGBAColor> color; ///< ellipsoid color. (default=0,0.5,1.0,1.0)
+    SOFA_ELLIPSOIDFORCEFIELD_RENAMEDDATA_DEPRECATED() Data<Coord> center; ///< ellipsoid center
+    SOFA_ELLIPSOIDFORCEFIELD_RENAMEDDATA_DEPRECATED() Data<Coord> vradius; ///< ellipsoid radius
+    SOFA_ELLIPSOIDFORCEFIELD_RENAMEDDATA_DEPRECATED() Data<Real> stiffness; ///< force stiffness (positive to repulse outward, negative inward)
+    SOFA_ELLIPSOIDFORCEFIELD_RENAMEDDATA_DEPRECATED() Data<Real> damping; ///< force damping
+    SOFA_ELLIPSOIDFORCEFIELD_RENAMEDDATA_DEPRECATED() Data<sofa::type::RGBAColor> color; ///< ellipsoid color. (default=0,0.5,1.0,1.0)
```
fredroy: I think this will be breaking, in the `behavior` way of meaning no ?
If one is using this class (inhering or composing) and was using those data (setValue or whatever), it will not do anything . I know there will be warnings at compilation time but still 🙊
Maybe at least inform the user at runtime with the `DeprecatedData`... or directly break the compilation with the DeprecatedAndRemoved thingy.

EDIT: I guess `DeprecatedData` wont do, as you rename the ""C++"" data, not the ""user space""(?) data itself 🤔

fredroy (changes requested):

Sofa/Component/MechanicalLoad/src/sofa/component/mechanicalload/EllipsoidForceField.h
alxbilger: you're right. It's fixed

Sofa/Component/MechanicalLoad/src/sofa/component/mechanicalload/EllipsoidForceField.h
alxbilger: The new version is not breaking

Sofa/Component/MechanicalLoad/src/sofa/component/mechanicalload/EllipsoidForceField.h
```diff
@@ -82,17 +82,42 @@ class EllipsoidForceField : public core::behavior::ForceField<DataTypes>
 
     };
 
-    Data<sofa::type::vector<Contact> > contacts; ///< Contacts
+    union
```
hugtalbot, line 85: wow cool didn´t know that !
is it right to say that then, no need to make any "redirections" from the deprecated contacts

hugtalbot (approved):

alxbilger:
The union comes with a problem: the ForceField constructor does not know which constructor from the union to call. The same for the destructor. For the constructor, the constructors of the Data are explicitly called so it is not a problem. However, it's a problem for the destructor. That is why an explicit call to Data destructors is needed.

See an example here: https://godbolt.org/z/svoo9za6G

Sofa/Component/MechanicalLoad/src/sofa/component/mechanicalload/EllipsoidForceField.h
alxbilger: yes, no redirection needed

fredroy (approved):

alxbilger:
@hugtalbot that's something to decide. It's not a magic method, it adds a new complexity in the deprecation mechanism. I am not sure

alxbilger:
@hugtalbot Note that you have to implement a destructor but you are allowed not to call the Data destructor. If you forget it, it's a memory leak.

PR #4044, opened 2023-07-14
[Core] Fix drawing of Objects when hiding Visual Models 

fredroy:
https://github.com/sofa-framework/sofa/pull/4044#issuecomment-1637217987

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #4045, opened 2023-07-15
[Core] Make doDrawVisual final

hugtalbot:
Based on #4044 

### DIFF
--- a/Sofa/framework/Simulation/Core/src/sofa/simulation/VisualVisitor.cpp
+++ b/Sofa/framework/Simulation/Core/src/sofa/simulation/VisualVisitor.cpp
@@ -49,9 +49,6 @@ Visitor::Result VisualVisitor::processNodeTopDown(simulation::Node* node)
 
 Visitor::Result VisualDrawVisitor::processNodeTopDown(simulation::Node* node)
 {
-    // don't draw if specified not to do so in the user interface
-    if (!vparams->displayFlags().getShowVisualModels())
-        return RESULT_PRUNE;
 
     // NB: hasShader is only used when there are visual models and getShader does a graph search when there is no shader,
     // which will most probably be the case when there are no visual models, so we skip the search unless we have visual models.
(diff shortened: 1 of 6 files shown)


bakpaul (approved):

alxbilger:
Changes in CudaVisualModel are expected. I tried to suggest them but `Permission to hugtalbot/sofa.git denied to alxbilger.`

Sofa/framework/Simulation/Core/src/sofa/simulation/VisualVisitor.cpp
```diff
@@ -49,9 +49,6 @@ Visitor::Result VisualVisitor::processNodeTopDown(simulation::Node* node)
 
 Visitor::Result VisualDrawVisitor::processNodeTopDown(simulation::Node* node)
 {
-    // don't draw if specified not to do so in the user interface
```
damienmarchal, line 52: Can you argue a bit why you want to call the rendering function of a visual model while the showvisualmodel is disabled ?

fredroy: https://github.com/sofa-framework/sofa/pull/4044#issuecomment-1637217987

damienmarchal: Well to me removing the test is not the way to fix that. VisualDrawVisitor shouldn't be used to call the drawing of non visual object. Doing that has the consequence of maintaining the fuzziness of use case between draw() and drawVisual(). 

So ok to merge  as is to restore the rendering of object using draw() but a more serious refactoring is needed there.

damienmarchal: Ok so if I understand this is already merged. So let's make this PR to ready. 

EDIT: The merged fixed is not one that should have been done because the VisualDrawVisitor is not supposed to process non VisualModel. But as it is merge let's move forward, maybe one day someone will refactor this code to split the draw() and drawVisual().

damienmarchal:
@bakpaul I added a comment and pass the PR to ready

PR #4061, opened 2023-07-21
[FEM.Elastic] Compute BBox in triangle FEM

alxbilger:


alxbilger:
[ci-build][with-all-tests]

hugtalbot (approved):

bakpaul:
I don't know if I understand your question well, but I guess any component responsible of drawing something should compute this BBox. And the mstate is not the only one. Furthermore, the forcefield may want to display information that are out of the volume (principal stress), so each BBox are potentially different.  So, if your question was "does the FF really need to compute it or would the Mstate be enought" I would say that we really need to compute it for the FF.

alxbilger:
@bakpaul is right.

PR #4122, opened 2023-08-29
[test] Introduce failing test on quaternions

alxbilger:
The test `QuaterTest::EulerAngles` randomly fails on the CI. Here is a test that fails 100% of the time. Anybody wants to fix it?

### DIFF
--- a/Sofa/framework/Type/test/Quater_test.cpp
+++ b/Sofa/framework/Type/test/Quater_test.cpp
@@ -73,6 +73,22 @@ TEST(QuaterTest, EulerAngles)
     }
 }
 
+// The test QuaterTest::EulerAngles relies on a RNG. This test is the same but with specific values
+// that were problematic in the other test.
+TEST(QuaterTest, EulerAnglesProblematicCase)
+{
+    Quat<double> q0(77 / 100.f - 1.f, 156 / 100.f - 1.f, 123 / 100.f - 1.f, 156 / 100.f - 1.f);
+    q0.normalize();
+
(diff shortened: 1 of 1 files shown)


olivier-roussel:
Looks like this particular case falls under the Euler angles singularities. The initial quaternion `q0` converted to Roll-Pitch-Yaw Euler angles has a pitch of `pi / 2`, which corresponds to a singularity case, i.e. there is no global and smooth solutions to the inverse problem of converting the rotation to Euler angles. In this particular case, one DOF is lost (I found the explanations of this quite intuitive and clear here: http://lavalle.pl/vr/node83.html ).
When we can afford it, a good way would be to only allow global parameterizations of SO(3) and simply not use Euler angles, but I guess we have reasons to have them in SOFA and have to deal with them.
I would not go to change anything in the `createQuaterFromEuler` or `toEulerVector` conversions functions, as to me they are valid as it and it is left to the user of Euler angles to know that they have to be used as their are, locally, as it is a local parameterization, i.e. not close to their singularities.
I guess this is what happened also in the random rotations failing tests.
What do you think about simply fixing our tests by excluding cases where quaternion `q0` is leading to a Euler angles singularity (here checking that pitch is not `pi / 2`), prior to converting them back to queternion `q1` ?

alxbilger:
@olivier-roussel thanks for your complete answer. I came to the same conclusion as you. I just wanted to be sure that the result of `toEulerVector` in this test is the right one. I just tried on matlab, and it ends up with one of the angle to be pi/2.

alxbilger:
Superseded by https://github.com/sofa-framework/sofa/pull/4126

PR #4151, opened 2023-09-08
[Mass] Fix UniformMass vertexMass value should not be set to 0 if nbr of points reach 0

epernod:
Fix problem encountered in SPH scenes where the number of particles is 0 at start. 
Therefor the vertexMass was set to 0 during the first steps and all particles created during the next simulation steps were still having no mass. 

### DIFF
--- a/Sofa/Component/Mass/src/sofa/component/mass/UniformMass.inl
+++ b/Sofa/Component/Mass/src/sofa/component/mass/UniformMass.inl
@@ -375,7 +375,6 @@ void UniformMass<DataTypes>::updateMassOnResize(sofa::Size newSize)
 {
     if (newSize == 0)
     {
-        d_vertexMass.setValue(static_cast<MassType>(0.0));
         d_totalMass.setValue(Real(0));
         return;
     }


alxbilger:
[ci-build][with-all-tests]

epernod:
> @epernod it means the function `updateOnResize()` is called at init? I fear that if we delete all DOFs during a simulation, when we add new DOFs again, `d_vertexMass` is set to zero therefore: `d_totalMass.setValue(Real(newSize) * Real(d_vertexMass.getValue()));`
> 
> the total mass will always remain null

yes that is why I remove the d_vertexMax setValue(0).  If you know you will add/remove dofs, you should rely on the d_vertexMax which doesn't change inside the code.

epernod:
> Have you checked then in the code that no access to vertexMass is done in any other function without checking the "newSize" ? @epernod

sorry I'm not sure to understand the point. I'm modifying the method `updateMassOnResize` which is a callback called when the number of dof has changed. From my understanding the VertexMass value should not change when the number of dof change, only the totalMass.

if we need to add more check in all method to check the size, this means the callback mechanism is useless no? and if it is.. what about us @hugtalbot 😆 ?

hugtalbot (approved):

PR #4158, opened 2023-09-12
[LinearSystem] Introduce constant sparsity matrix assembly

alxbilger:
As a description of the new feature, here is the doxygen of the new component:

> Assembling method benefitting from the constant sparsity pattern of the linear system in order to increase the
> performances.
> 
> This method must not be used if the sparsity pattern changes from a time step to another.
> 
> The method uses the first iteration to build the data structures required to accelerate the matrix assembly, based
> on a constant sparsity pattern.
> 
> First iteration:
> 1) During the accumulation of matrix contributions, the local matrix stores an ordered list of matrix elements (row
> and column) corresponding to the order of insertion. This order must not change along the simulation.
> 2) After the matrix assembly, the global matrix is compressed so that the ordered list of matrix elements can be
> converted into an ordered list of ids corresponding to locations in the values array of the compressed matrix.
> 
> Second time step and after:
> 1) The local matrices assume the order of insertion did not change. Therefore, they rely only on the ordered list of
> ids to know where in the values array to insert the matrix contribution. The row and column ids are useless.

An example scene is introduced. 

peek

This scene assembles mapped matrices to show that the method works in those conditions. Here is the assembled matrix represented as a binary image:

image

This scene is used for the benchmarks on 1000 time steps. The timer to focus on is `buildStiffness`. But be aware that this timer also includes the computation of the contributions in addition to the assembly itself.

1. The scene without modification, i.e. with `<ConstantSparsityPatternSystem template="CompressedRowSparseMatrixd" name="A" printLog="true"/>`

```
[INFO]    [BatchGUI] 1000 iterations done in 12.3859 s ( 80.7368 FPS).

   7       0.82    1       1.57    8.67    1.81    0.34    1.81   14.74 .......buildStiffness
```

2. The scene with a regular matrix assembly, i.e. with `<MatrixLinearSystem template="CompressedRowSparseMatrixd" name="A" printLog="true"/>`

```
[INFO]    [BatchGUI] 1000 iterations done in 15.0063 s ( 66.6387 FPS).

   7       0.69    1       4.69    6.45    4.95    0.18    4.95   33.30 .......buildStiffness
```

3. The scene with a regular matrix assembly, i.e. with `<MatrixLinearSystem template="CompressedRowSparseMatrixMat3x3d" name="A" printLog="true"/>` (blocks of 3x3)

```
[INFO]    [BatchGUI] 1000 iterations done in 16.765 s ( 59.648 FPS).

   7       0.84    1       4.71   13.93    5.51    1.11    5.51   33.16 .......buildStiffness
```

In benchmark 1, we have a x2.7 speed-up compared to benchmark 2.


alxbilger:
[ci-build][with-all-tests]

Sofa/Component/LinearSystem/src/sofa/component/linearsystem/ConstantSparsityPatternSystem.cpp
```diff
@@ -0,0 +1,382 @@
...
+        {
+            mat.second->fullRows();
+            buildHashTable(*mat.second, m_constantCRSMappingMappedMatrices[i++]);
+        }
+
+        //replace the local matrix components by new ones that use the hash table
+        replaceLocalMatrices(mparams, getLocalMatrixMap<Contribution::STIFFNESS>());
+        replaceLocalMatrices(mparams, getLocalMatrixMap<Contribution::MASS>());
+        replaceLocalMatrices(mparams, getLocalMatrixMap<Contribution::DAMPING>());
+        replaceLocalMatrices(mparams, getLocalMatrixMap<Contribution::GEOMETRIC_STIFFNESS>());
```
hugtalbot: It seems that the sparsity mapping is computed in the `applyProjectiveConstraints` but I can not understand why the projection step is the good moment to save the sparsity

Sofa/Component/LinearSystem/src/sofa/component/linearsystem/ConstantSparsityPatternSystem.cpp
```diff
@@ -0,0 +1,382 @@
...
+
+        m_isConstantSparsityPatternUsedYet = true;
```
hugtalbot: This is the only place where it is set to true, I guess that we could implement a mechanism (e.g. when topo changes occur) modifying this flag.
Is this correct?

Sofa/Component/LinearSystem/src/sofa/component/linearsystem/ConstantSparsityPatternSystem.cpp
```diff
@@ -0,0 +1,382 @@
...
+    {
+        if (n != getSystemMatrix()->rowSize() || n != getSystemMatrix()->colSize())
+        {
+            getSystemMatrix()->resize(n, n);
+        }
+        else
+        {
+            // In the CRS format, the pattern is unchanged from a time step to the next
+            // Only the values are reset to 0
+
```
hugtalbot:
```suggestion
```

fredroy:
I tried with a scene from BeamAdapter, but those are all using BTDLinearSolver which is templated on BTDMatrix6.
But ConstantSparsityPatternSystem  is only templated on CompressedRowSparseMatrix so I dont think it is possible to use it with BTDLinearSolver.

alxbilger:
I'll check how difficult it is to write this component for BTD.
In the meantime, I also have unit tests to fix.

alxbilger:

alxbilger:
@fredroy @hugtalbot For a BTD linear system, we already use its constant sparsity pattern in a dedicated data structure. It makes the `add` method quite cheap (to be measured). Therefore, there is no need to implement `ConstantSparsityPatternSystem` for `BTDMatrix`.

bakpaul:
It works well on my side, unfortunately my scene didn't present any topological modification, so the result is quite the same as yours : 

Without the sparsity pattern (with  `<MatrixLinearSystem template="CompressedRowSparseMatrixd" name="A" printLog="true"/>`)

```
[INFO]    [BatchGUI] 1000 iterations done in 29.7571 s ( 33.6055 FPS). 

   8       8.45    2       0.05   17.80    5.49    5.48   10.98   35.15 ........buildStiffness
```


With the sparsity pattern (with  `<ConstantSparsityPatternSystem template="CompressedRowSparseMatrixd" name="A" printLog="true"/>`)

```
[INFO]    [BatchGUI] 1000 iterations done in 24.846 s ( 40.248 FPS). 

   8       8.24    2       0.06    6.65    2.64    2.59    5.28   21.03 ........buildStiffness
```

bakpaul (approved):

hugtalbot (approved):

fredroy (approved):

Sofa/Component/LinearSystem/src/sofa/component/linearsystem/ConstantSparsityPatternSystem.cpp
alxbilger: Yes, it could be

Sofa/Component/LinearSystem/src/sofa/component/linearsystem/ConstantSparsityPatternSystem.cpp
alxbilger: When a component contributes to the matrix, it does not know if a DoF is concerned by a projective constraint. So, the insertion must happen. For that, you need not to take into account the projective constraints.

PR #4160, opened 2023-09-12
[all] Use SimulationInitDoneEvent instead of AnimateBeginEvent

hugtalbot:
Based on #4075 , this PR proposes to use the `SimulationInitDoneEvent` instead of `AnimateBeginEvent` when an export is desired at the beginning of the simulation. By "beginning of the simulation" we actually mean, once the `init()` and `bwdInit()` are completed which is exactly the purpose of `SimulationInitDoneEvent`.


Sofa/framework/Simulation/Core/src/sofa/simulation/BaseSimulationExporter.cpp
```diff
@@ -121,24 +128,19 @@ void BaseSimulationExporter::updateFromDataField()
         d_filename.setValue(getName());
     }
 
-    /// Activate the listening to the event in order to be able to export file at the nth-step
-    if(d_exportEveryNbSteps.getValue() != 0)
+    /// Activate the listening to the event in order to be able to export file at first step or the nth-step
```
hugtalbot: update comment

hugtalbot:
```suggestion
    /// Activate the listening to the event in order to be able to export file once initialization is done or the nth-step
```

Sofa/Component/IO/Mesh/tests/VisualModelOBJExporter_test.cpp
```diff
@@ -95,6 +101,11 @@ class VisualModelOBJExporter_test : public BaseSimulationTest {
         ASSERT_NE(root.get(), nullptr) ;
         root->init(sofa::core::execparams::defaultInstance()) ;
 
+        // SimulationInitDoneEvent is used to trigger exportAtBegin
+        SimulationInitDoneEvent endInit;
+        PropagateEventVisitor pe{sofa::core::execparams::defaultInstance(), &endInit};
```
epernod, line 106: why do you need to propagate the event yourself, it is not already done in the animationLoop ?

epernod (commented):
In #4075 you introduce `bool m_isFirstStep = true;`
which is not anymore useful with this event. Do you want to merge this one first?

Sofa/Component/IO/Mesh/tests/VisualModelOBJExporter_test.cpp
hugtalbot: The function `init()` triggers the InitVisitor (init+bwdInit) but the event is not emitted.
Only the class Simulation (Simulation.cpp line 144) triggers it when you are in a simulation. Here, in a C++ scene it does not work this way.

Maybe the event could be triggered by the InitVisitor itself..

hugtalbot: About `bool m_isFirstStep = true;` introduced in #4075 it should disappear as updated in this PR
However, it would be better to first merge the #4075 then this one (#4160)

epernod (approved):

epernod:
[ci-build][with-all-tests]

fredroy (approved):

PR #4179, opened 2023-09-18
[LinearAlgebra] Fix segfault when accessing and empty CSR matrix (through iterators)

courtecuisse:

### DIFF
--- a/Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixGeneric.h
+++ b/Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixGeneric.h
@@ -254,7 +254,14 @@ public :
 
     const VecIndex& getRowIndex() const { return rowIndex; }
     const VecIndex& getRowBegin() const { return rowBegin; }
-    Range getRowRange(Index id) const { return Range(rowBegin[id], rowBegin[id+1]); }
+
+    Range getRowRange(Index id) const
+    {
+        if ((unsigned) id+1 >= rowBegin.size()) return Range();
+
+        return Range(rowBegin[id], rowBegin[id+1]);
+    }
+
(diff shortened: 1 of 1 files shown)


Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixGeneric.h
```diff
@@ -254,7 +254,14 @@ public :
 
     const VecIndex& getRowIndex() const { return rowIndex; }
     const VecIndex& getRowBegin() const { return rowBegin; }
-    Range getRowRange(Index id) const { return Range(rowBegin[id], rowBegin[id+1]); }
+
+    Range getRowRange(Index id) const
+    {
+        if ((unsigned) id+1 >= rowBegin.size()) return Range();
```
alxbilger, line 260:
```suggestion
        if ((unsigned) id >= rowBegin.size()) return Range();
```
`id == rowBegin.size()` would also lead to an out-of-bounds.

alxbilger (changes requested):
The change makes sense, but I would like to know if your code worked in the past, before some SOFA changes, as it appears that the method `getRowRange` never checked the bounds.

fredroy:
[ci-build][with-all-tests]

fredroy:
Well spotted @courtecuisse 

- This fix reminds me that some bound checks (e.g type::vector) does only check in DEBUG mode/assertion https://github.com/sofa-framework/sofa/blob/649a41bf386b78c95264d9a6fb3125f4662c5a58/Sofa/framework/Type/src/sofa/type/vector_T.h#L108
obviously for performance reason. Should we do that here ? Thing is, many people do not run in debug so.... 🤐

- I guess this is a far cry but @fjourdes  you would want to get the fix ? (if you dont have it already...)

fredroy:
Superseded by https://github.com/sofa-framework/sofa/pull/4236
(@courtecuisse)

PR #4180, opened 2023-09-19
[LinearSolver.Direct] SparseLDL: Fix crash in addJMInvJtLocal

fredroy:
**[Only Windows/MSVC]**

I cannot explain at all 👺

On MSVC2022(17.7.4) and Release mode, a crash happens with `examples/Demos/fallingBeamLagrangianCollision.scn` when a collision occurs. It crashes when trying to insert values in the triplets vector (suspecting a memory error somewhere)

Weirdly, it does not crash in RelWithDebug mode 🤐. 
*And* it does not crash as well on MSVC2019, as the CI (jenkins) uses it in release mode while testing the scene.

At the code generation time (compiler?optimizer?) it does throw a warning like:
`1>C:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.37.32822\include\vector(398): warning C4789: buffer 'triplets' of size 16 bytes will be overrun; 8 bytes will be written starting at offset 16`
only in Release mode

--

Two fixes possible (found by trial-and-error 🤡)
 - using std::vector instead of sofa::type::vector
 - using a struct instead of the tuple
 
I used the std::vector one as it was the simplest.

I suspect there is a bug(???)/issue with the optimizer/vectorizer of MSVC , or the allocator but i cannot prove anything 🙊


alxbilger:
I made some tests on godbolt, but I was not able to reproduce the crash. In any case, one day it would be great to remove the public inheritance from `std::vector` (but SofaCUDA probably doesn't agree).

fredroy:
> I made some tests on godbolt, but I was not able to reproduce the crash. In any case, one day it would be great to remove the public inheritance from `std::vector` (but SofaCUDA probably doesn't agree).

I tried to switch to at least to private inheritance but I can tell you that not only SofaCUDA is complaining 😅

alxbilger:
yes, private inheritance requires some changes. But I wonder if inheritance is really necessary at all. If it's only to get the bounds checking in debug, there are compilation options for some compilers.

fredroy:
Further tests: if LTO is not enabled, then it does not crash as well, so the culprit would be something about the optimizer/vectorizer 🧐

bakpaul:
Nice catch ! 
Didn't try, I only work on ubuntu. 
Yeah I guess that many custom containers appeared in SOFA at a time where it was not that advanced in the stl. Maybe that would be great to take a look at it and see what we can remove...

hugtalbot (approved):

PR #4182, opened 2023-09-19
[Config] Integrate Tracy profiler

alxbilger:
I integrated the Tracy profiler (https://github.com/wolfpld/tracy). It adds a ton of new feature to our internal profiler. In particular, I like:
- Multithreaded timeline
- Statistics
- Messages
- Memory management (need more work, I could not test it)
- See context switching
- Mutex profiling (not tested yet)

I tried not to be invasive, but it will require a change on the way we define timers such as in https://github.com/sofa-framework/sofa/compare/master...alxbilger:sofa:scoped_timer (the demo runs with this branch).

TODO: a dedicated MessageHandler

peek (1)

### DIFF
--- a/Sofa/GUI/Qt/src/sofa/gui/qt/RealGUI.cpp
+++ b/Sofa/GUI/Qt/src/sofa/gui/qt/RealGUI.cpp
@@ -55,11 +55,9 @@
 #include <sofa/component/visual/VisualStyle.h>
 #include <sofa/helper/AdvancedTimer.h>
 #include <sofa/helper/ScopedAdvancedTimer.h>
+#include <sofa/simulation/SimulationLoop.h>
 
 #include <sofa/helper/system/SetDirectory.h>
-#ifdef TRACY_ENABLE
-#include <tracy/Tracy.hpp>
-#endif
 using sofa::helper::system::SetDirectory;
 
 #include <sofa/helper/system/FileSystem.h>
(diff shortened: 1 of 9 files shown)


epernod:
Very cool and useful indeed! It really look like the one in unity!
I see the Tracy lib is under BSD license, is it a problem to add it in the core of SOFA ? @hugtalbot 
otherwise we should move the profiler into another repo?

fredroy:
> Very cool and useful indeed! It really look like the one in unity! I see the Tracy lib is under BSD license, is it a problem to add it in the core of SOFA ? @hugtalbot otherwise we should move the profiler into another repo?

@alxbilger did it in a way we dont even ship it with SOFA as it will downloaded from their repo at the cmake step.
Moreover, this is for developers so the binary distribution would not contain the compiled lib anyway.

In any case, BSD licence is very lax anyway so it should be OK.

fredroy:
The screenshots on the repo look mad 🤩 (even asm views of statement 😶‍🌫️)
I am especially interested of the multi-threading part, as the profiler of MSVC (and the valgrind/callgrind on linux) does not really handle it well.

- ~~Do you need to compile with debug info?~~
- It is tracing so obviously there will be overhead, but can you say how much is the performance hit (if any) ? (msvc's is almost as fast but requires RelWithDebugInfo and callgrind is slow like hell)

EDIT: tried myself, works in Release mode as well 😎👍

PR #4132, opened 2023-09-05
[Constraint.Lagrangian.Solver] GenericConstraintSolver: use given re-ordered list for unbuilt GS

fredroy:
The (potential) renumbering of the constraints by the ConstraintCorrection was not used in GenericConstraintSolver.
It is (only?) re-ordered when LinearSolverCC is set with "wire optimization" and gives a big speed-up when used with the partial_solve of BTDLinearSolver (and not building matrices)

This optimization was already in the LCPConstraintSolver and explain (but not totally) the big difference of speed for the scenes with LinearCC/BTD and unbuilt gaussseidel.

Numbers from `BeamAdapter/examples/3instruments_collis.scn -g batch -n 5000` , all constraint solvers use unbuilt GS.
```
(mu=0.1)
LCP        = [INFO]    [BatchGUI] 5000 iterations done in 61.9292 s ( 80.7373 FPS).
GCS before = [INFO]    [BatchGUI] 5000 iterations done in 261.396 s ( 19.1281 FPS)
GCS after  = [INFO]    [BatchGUI] 5000 iterations done in 85.083 s ( 58.7661 FPS)
```
```
(mu=0.0)
LCP        = [INFO]    [BatchGUI] 5000 iterations done in 43.2622 s ( 115.574 FPS).
GCS before = [INFO]    [BatchGUI] 5000 iterations done in 230.029 s ( 21.7364 FPS)
GCS after  = [INFO]    [BatchGUI] 5000 iterations done in 68.9986 s ( 72.4652 FPS)
```

As for the simulation itself, the results are not exactly the same (even if visually, looks the same) between LCPCS or GCS
`regression step-by-step, 2000 steps`
```
(mu=0.1)
TOTALERROR: 3.18005
ERRORBYDOF: 0.00310931
```
```
(mu=0.0)
TOTALERROR: 1.55088
ERRORBYDOF: 0.00154138
```

But I cannot tell which one of the two (LCPCS or GCS) is the most "precise" 🤷

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
[...]

PR #4195, opened 2023-09-25
[Constraint.Lagrangian.Solver] GenericConstraintSolver: avoid repeated allocation in loops

fredroy:
Use pre-allocated buffers instead of creating new vectors + allocations at every step of the loops.

Same benchs as 
 - https://github.com/sofa-framework/sofa/pull/4132

```
(mu=0.1)
LCP :       5000 iterations done in 58.1724 s ( 85.9514 FPS)
GCS before: 5000 iterations done in 83.4733 s ( 59.8994 FPS)
GCS after:  5000 iterations done in 69.2669 s ( 72.1846 FPS)
```
```
(mu=0.0)
LCP :       5000 iterations done in 44.0637 s ( 113.472 FPS)
GCS before: 5000 iterations done in 69.1173 s ( 72.3408 FPS)
GCS after:  5000 iterations done in 53.8957 s ( 92.7719 FPS)
```

Speedup of ~20-30% but still slower than LCP (but much less though ).
But this improvement should be felt whenever the GCS is using the unbuilt method (contrary to #4132 which was specialized for LinearConstraintCorrection with wire optimization)

Other bench on caduceus (5000 steps)
```
LCP
4       1.11    1       0       0.69    0.14    0.04    0.14    9.10 ....SolveConstraint: POSITION AND VELOCITY - SolveSystem
```
```
GCSunbuilt before
4       1.28    1       0      33.01    0.24    1.45    0.24   13.24 ....SolveConstraint: POSITION AND VELOCITY - SolveSystem
```
```
GCSunbuilt   after
4       1.28    1       0      19.89    0.16    0.98    0.16    9.38 ....SolveConstraint: POSITION AND VELOCITY - SolveSystem
```

### DIFF
--- a/Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintProblem.cpp
+++ b/Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintProblem.cpp
@@ -317,10 +317,9 @@ void GenericConstraintProblem::unbuiltGaussSeidel(SReal timeout, GenericConstrai
 
     tabErrors.resize(dimension);
 
-    // pre-allocated buffers
-    constexpr std::size_t bufferSize = 10;
-    std::vector<SReal> errF(bufferSize);
-    std::vector<SReal> tempF(bufferSize);
+    // temporary buffers
+    std::vector<SReal> errF;
+    std::vector<SReal> tempF;
 
     for(iter=0; iter < static_cast<unsigned int>(maxIterations); iter++)
     {
@@ -339,7 +338,7 @@ void GenericConstraintProblem::unbuiltGaussSeidel(SReal timeout, GenericConstrai
 
             //2. for each line we compute the actual value of d
             //   (a)d is set to dfree
-            if(nb > bufferSize)
+            if(nb > errF.size())
             {
                 errF.resize(nb);
             }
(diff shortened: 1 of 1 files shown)


alxbilger (commented):
Don't you think that moving `errF` and `tempF` out of the loop is enough? I would test to only resize them in the loop. The underlying memory allocation should not happen a lot. My suggestion is because I think it is simpler to understand compared to the buffer implementation.

In any case, I think the condition `if (nb > bufferSize)` is not necessary.

fredroy:
[ci-build][with-all-tests]

> Don't you think that moving `errF` and `tempF` out of the loop is enough? I would test to only resize them in the loop. The underlying memory allocation should not happen a lot. My suggestion is because I think it is simpler to understand compared to the buffer implementation.

I will try that 👍

> 
> In any case, I think the condition `if (nb > bufferSize)` is not necessary.

Thats what I was thinking but not sure 🧐

fredroy:
> Don't you think that moving `errF` and `tempF` out of the loop is enough? I would test to only resize them in the loop. The underlying memory allocation should not happen a lot. My suggestion is because I think it is simpler to understand compared to the buffer implementation.
> 
> In any case, I think the condition `if (nb > bufferSize)` is not necessary.

I obtained the same level of improvement so I will go with this easier way 👍

bakpaul:
[ci-build][with-all-tests]

bakpaul (approved):

hugtalbot (approved):

PR #4213, opened 2023-10-02
[Constraint.Lagrangian.Solver] Another step to factorize both constraint solvers

alxbilger:
Some methods are moved into a base class. They were exactly doing the same thing in both solvers. The result is that `buildSystem` now looks very similar in both classes. It allows to better understand that both solvers do some similar things. It could be factorized again in the future.

A lot of cosmetic changes

`applyCorrection` has been split into smaller functions

`build_LCP` and `build_problem_info` are factorized

`ConstraintSolver::solveConstraint` is condensed

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/ConstraintParams.h
+++ b/Sofa/framework/Core/src/sofa/core/ConstraintParams.h
@@ -35,13 +35,39 @@ class SOFA_CORE_API ConstraintParams : public sofa::core::ExecParams
 public:
 
     /// Description of the order of the constraint
-    enum ConstOrder
+    enum class ConstOrder
     {
         POS = 0,
         VEL,
         ACC,
         POS_AND_VEL
     };
+
+    SOFA_ATTRIBUTE_DEPRECATED__CONSTORDER() static constexpr auto POS = ConstOrder::POS;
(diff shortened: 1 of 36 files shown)


alxbilger:
[ci-build][with-all-tests]

alxbilger:
Yes

fredroy (approved):
No noticeable perf regression on 3instrument_collis.scn from BeamAdapter either with the LCPConstraintSolver or the GenericConstraintSolver (both with unbuilt method), with MSVC2022

Sofa/framework/Core/src/sofa/core/ConstraintParams.h
```diff
@@ -35,14 +35,40 @@ class SOFA_CORE_API ConstraintParams : public sofa::core::ExecParams
 public:
 
     /// Description of the order of the constraint
-    enum ConstOrder
+    enum class ConstOrder
```
fredroy: Food for thought:
I feel a bit confused by this `ConstOrder` ; on a first glance, I read "Constant Order" instead of "Constraint Order".
Especially that a bit further, we have an other "Const" with `ConstMultiVecCoordId` with the "Constant" meaning in this case.

alxbilger: That is a fair comment, I agree with you. I'll make some changes

PR #4219, opened 2023-10-04
[CollisionModel] New: added collision layer property

polyakovkrylo:
This PR introduces  collision layers, a collision filtering algorithm similar to the one used in [Unity's collsion pipeline](https://docs.unity3d.com/Manual/LayerBasedCollision.html). The overall idea is to omit collision detection for models that are in different layers.  This way, each layer can be used for different functionality, e.g. cutting simulation.

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/CollisionModel.h
+++ b/Sofa/framework/Core/src/sofa/core/CollisionModel.h
@@ -407,6 +407,9 @@ protected:
     /// models included in a common group (i.e. sharing a common id)
     Data< std::set<int> > group;
 
+    // No collision can occur between collision models in different layers
+    Data<int> layer;
+
     /// Number of collision elements
     Size size;
 
--- a/Sofa/framework/Core/src/sofa/core/CollisionModel.cpp
+++ b/Sofa/framework/Core/src/sofa/core/CollisionModel.cpp
@@ -165,6 +165,9 @@ bool CollisionModel::canCollideWith(CollisionModel* model)
     if (model->getContext() == this->getContext()) // models are in the Node -> is self collision activated?
         return bSelfCollision.getValue();
 
+    if (model->layer.getValue() != this->layer.getValue()) // no collisions between models in different layers
+        return false;
+
     const auto& myGroups = this->group.getValue();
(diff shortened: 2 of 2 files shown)


alxbilger:
Hey! Can you explain how it is different from the `group` property? Thanks

polyakovkrylo:
Hi, @alxbilger! It works the opposite way the collision groups work. With collision groups, collision can only occur between the models' group sets do not overlap (no common groups). With the collision layers it' s the opposite -- collision can only occur if the models are in the same layer.

Sofa/framework/Core/src/sofa/core/CollisionModel.h
```diff
@@ -407,6 +407,9 @@ class SOFA_CORE_API CollisionModel : public virtual objectmodel::BaseObject
     /// models included in a common group (i.e. sharing a common id)
     Data< std::set<int> > group;
 
+    // No collision can occur between collision models in different layers
+    Data<int> layer;
```
alxbilger, line 411:
```suggestion
    Data<int> d_layer;
```

alxbilger (changes requested):

Sofa/framework/Core/src/sofa/core/CollisionModel.h
alxbilger: It also needs to be initialized in the constructor

alxbilger: Does it make sense to define more than one layer for a collision model?

bakpaul:
Hi vlad, this PR has also been discussed during the dev meeting. 
The conclusions are the same as for the #4226 : we need more context/example of use (for instance to see how you use it in your scenes or plugin). 

We are also curious to know how you handle differences between this and the group mechanism : what happens if you are in the same time in the same group and in the same layer ? Are the two mechanism really necessary ? 
It could be also possible to remplace the group mechanism by yours to have a more positive mechanism (same layer = collision instead of diffrent group = collision). But for this it would need to handle multiple layer by object. 

And finally, again, some unit testing would be great if we manage to find a mechanism that suites every one.

PR #4226, opened 2023-10-09
[DetectionOutput] New: added barycentric coords

polyakovkrylo:
With this PR, I  suggest to return  barycentric coords to the detection output  members. In my use case , these coordinates are used in cutting simulation to create new vertices from ancestor points when propagating a cut.


bakpaul:
Hi ! This is of interest. But it would gain in usability if it was also generalized for triangles. Adding barycentric coords of size 2 limits its use to edges and thus cannot be hard coded in the general class. Have you done something similar for your cutting simulations @epernod ?

alxbilger:
Related to https://github.com/sofa-framework/sofa/commit/e2c9d544b279cb3b1315405c046fb604c8bd8f60

bakpaul:
Hi Vlad. We have discussed your PR during the sofa dev meeting. And this is what came out : 
This is of interest but would require an example/context (plugin using it) of use to see how you use it. More over, it still need a lot to do to integrate it in a more standardized way to be more generic regarding the type of element. And finally, and we could help you for this, this feature would need unit tests.

This pr is currently set to WIP, waiting for further comment from you.

Issue #4221, opened 2023-10-05
Use of embedded external dependencies


olivier-roussel:
Following the remarks we got from the SOFA conda package PR on conda-forge (https://github.com/conda-forge/staged-recipes/pull/23085), we have some problems regarding some external libraries that are embedded in SOFA code. 
Here is a try to sum up all of these embedded libraries:
| Lib name | Type | Location in code | License  | Status |
| --- | --- | --- | --- | --- |
| difflib | header-only | `extlibs` | MIT | OK |
| json nlohmann | header-only | `extlibs` | MIT ~~**(should be in a separate file)**~~  | OK. Fixed in #4272 |
| stb | header-only | `extlibs` | MIT | OK |
| TinyXML | static | `extlibs` | zlib **(should be in a separate file)**  | Externalize tinyxml dependency: #4240 (waiting for merge) |
| miniFlowVR | static | `applications/plugins/SofaDistanceGrid/extlibs` | LGPL (version ?) **(should be in a separate file)** | Not built by default. TODO: plugin to be externalized (#1442 )  |
| newmat | static | `applications/plugins/SofaNewmat/extlibs` | type inconnu | Not built by default. TODO: plugin to be externalized (#1442 )  |
| CImg | header-only | `applications/plugins/CImgPlugin/extlibs` | CeCILL-C / CeCILL | OK |
| csparse | static | `Sofa/Component/LinearSolver/Direct/extlibs` | **no license file** | Plugin to be externalized. WIP : #4258 |
| metis | static | `Sofa/Component/LinearSolver/Direct/extlibs` | Apache v2 | Externalize metis dependency : #4266 (waiting for merge) |
| gtest | shared | `Sofa/framework/Testing/extlibs` | BSD-3-Clause | Externalize ? |
| cxxopts | header-only | `Sofa/GUI/Common (fetched from GitHub)` | MIT | Add to possibility to use external dependency and prevent from fetching: Fixed in #4273 |
| libQGLViewer | shared | `Sofa/GUI/Qt` | GPL | Find or use embedded code (find module being fixed in #4290) |

The actions we should take to make our code compatible for the conda package on conda-forge are:
1. Fix the missing licenses files (points in bold)
2. Avoid all static libraries.

[...]

PR #4240, opened 2023-10-17
[all] Replace tinyxml by external tinyxml2

olivier-roussel:
This aims to tackle one point of #4221 . Required for the conda-forge package (https://github.com/conda-forge/staged-recipes/pull/23085).
This PR propose to replace tinyxml embedded library, which is unmaintained now for a long time, by tinyxml2 (https://github.com/leethomason/tinyxml2/tree/master, under Z-lib license) which looks to be its successor and best candidate, as it is maintened, well packaged with cmake and available on conda-forge. 
Also, it is compatible for a future "find package or fetch content" mechanism in cmake. I don't know if we want to split the implementation of this additional feature for this package in another PR because if we do so, current code would need an additional required dependency to tinyxml2. 
The "find or fetch" cmake mechanism for this package, avoiding the requirement for this dependency to be installed, can be implemented as it:

```
# find or fetch TinyXML2
include(FetchContent)
FetchContent_Declare(
  tinyxml2
  URL https://github.com/leethomason/tinyxml2/archive/refs/tags/9.0.0.tar.gz
  FIND_PACKAGE_ARGS NAMES TinyXML2
)
FetchContent_GetProperties(tinyxml2)
if(NOT tinyxml2_POPULATED)
  FetchContent_Populate(tinyxml2)

  set(tinyxml2_SHARED_LIBS ON CACHE INTERNAL "")

  add_subdirectory(${tinyxml2_SOURCE_DIR} ${tinyxml2_BINARY_DIR})
endif()
```
However, this would required cmake >=3.24 and this call to be at a toplevel cmake (duplicating this code to replace all current calls to `find_package(TinyXML2)` does not seem desirable neither, so we might think also about how we can implement that in the cmake chain).


alxbilger:
CMake has the version 3.16.3 on the CI. It will not configure if you use the find_or_fetch feature

extlibs/CMakeLists.txt
```diff
@@ -13,15 +10,4 @@ set(STB_ROOT "${CMAKE_CURRENT_LIST_DIR}/stb" CACHE PATH "STB directory")
...
-if(NOT TinyXML_FOUND)
-    add_subdirectory(tinyxml tinyxml)
-endif()
-
 ####################################
-
-foreach(target ${EXTLIBS_TARGETS})
-    set_target_properties(${target} PROPERTIES DEBUG_POSTFIX "_d")
-    set_target_properties(${target} PROPERTIES FOLDER "SofaExtlibs")
-endforeach()
```
alxbilger, line 27: Is it only related to tinyxml?

olivier-roussel: To me, yes, as it related to https://github.com/sofa-framework/sofa/blob/f6bee9f9cd69ebd102324e0181325329f96ada94/extlibs/CMakeLists.txt#L3-L5 . 
`${EXTLIBS_TARGETS}` is defined here only with the embedded version of `tinyxml` (contains all targets linked for embedded libraries). It is then empty and should not be used in the future to me as we are moving out non-header only embedded libraries.

olivier-roussel:
In addition to my initial comment, this version of "find or fetch" using `FetchContent_MakeAvailable()` (available from version 3.14) should be prefered:
```
# find or fetch TinyXML2
include(FetchContent)
FetchContent_Declare(
  tinyxml2
  URL https://github.com/leethomason/tinyxml2/archive/refs/tags/9.0.0.tar.gz
  FIND_PACKAGE_ARGS NAMES TinyXML2
)
set(tinyxml2_SHARED_LIBS ON)
FetchContent_MakeAvailable(tinyxml2)
```

olivier-roussel:
Alternatively, we could have a naive custom "find or fetch" mechanism such as 
```
# find or fetch TinyXML2
find_package(TinyXML2)
if(NOT TinyXML2_FOUND)
  include(FetchContent)
  FetchContent_Declare(
    tinyxml2
    URL https://github.com/leethomason/tinyxml2/archive/refs/tags/9.0.0.tar.gz
    URL_HASH MD5=afecd941107a8e74d3d1b4363cf52bd7
    )
  set(tinyxml2_SHARED_LIBS ON)
  FetchContent_MakeAvailable(tinyxml2)
endif()
```
But, we typically want to avoid this call at each dependency requirement (so here replacing the current `find_package(TinyXML2)` calls, especially in the fetch case.
We could have these lines in a top-level CMakeLists instead, but this go in the opposite direction of our CMake dependencies management where it seems that we search the specific dependency on each target CMakeLists.
I believe (to be confirmed) that this can be avoid with the new mechanism introduced in cmake v3.24, by calling once the `FetchContent_Declare` (so in a kind of top-level cmake) and calling `FetchContent_MakeAvailable` for each target that needs it.

alxbilger:
> ```
> # find or fetch TinyXML2
> find_package(TinyXML2)
> if(NOT TinyXML2_FOUND)
>   include(FetchContent)
>   FetchContent_Declare(
>     tinyxml2
>     URL https://github.com/leethomason/tinyxml2/archive/refs/tags/9.0.0.tar.gz
>     URL_HASH MD5=afecd941107a8e74d3d1b4363cf52bd7
>     )
>   set(tinyxml2_SHARED_LIBS ON)
>   FetchContent_MakeAvailable(tinyxml2)
> endif()
> ```
Can it be written in our own `FindTinyXML2.cmake` file?

olivier-roussel:
I guess so. What you have in mind is that it could be a way to avoid multiple fetches in case of multiple calls to `find_package` by each target ? Not yet sure on how this could be implemented through

olivier-roussel:
Following our recent discussions, the "find or fetch" is replaced by a more classical find package strategy. The package management is fully left to the user. 
As some package managers were not providing cmake config files for tinyxml2 (at least APT), this PR also provides a cmake find module to look for a config file first, and if none found, look manually for the tinyxml2 library. 
Tested with:
- APT package libtinyxml2-dev (v9.0.0), no cmake config file provided
- conda package tinyxml2 (v9.0.0), cmake config file provided
- vcpkg package tinyxml2 (v9.0.0), cmake config file provided.

bakpaul:
Tested and compiled on VS22/Ubuntu22.04 and it works.

fredroy (approved):

hugtalbot (approved):

PR #4242, opened 2023-10-18
[Topology] Check indices out-of-bound in TriangleSetTopologyContainer

olivier-goury:
Check that triangles index match the size of the triangleAroundVertexArray (Like it is done in EdgeSetTopologyContainer for EdgeAroundVertexArray)

### DIFF
--- a/Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/TriangleSetTopologyContainer.cpp
+++ b/Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/TriangleSetTopologyContainer.cpp
@@ -125,6 +125,13 @@ void TriangleSetTopologyContainer::createTrianglesAroundVertexArray()
     m_trianglesAroundVertex.resize(getNbPoints());
     for (size_t i = 0; i < m_triangle.size(); ++i)
     {
+        if (m_triangle[i][0] >= getNbPoints() || m_triangle[i][1] >= getNbPoints() || m_triangle[i][2] >= getNbPoints())
+        {
+            msg_warning() << "trianglesAroundVertex creation failed, Triangle buffer is not concistent with number of points: Triangle: " << m_triangle[i] << " for: " << getNbPoints() << " points.";
+            continue;
+        }
+
+
         // adding edge i in the edge shell of both points
         for (unsigned int j=0; j<3; ++j)
             m_trianglesAroundVertex[ m_triangle[i][j]  ].push_back( (TriangleID)i );
--- a/Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/TriangleSetTopologyContainer.cpp
+++ b/Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/TriangleSetTopologyContainer.cpp
@@ -127,7 +127,7 @@ void TriangleSetTopologyContainer::createTrianglesAroundVertexArray()
     {
         if (m_triangle[i][0] >= getNbPoints() || m_triangle[i][1] >= getNbPoints() || m_triangle[i][2] >= getNbPoints())
         {
-            msg_warning() << "trianglesAroundVertex creation failed, Triangle buffer is not concistent with number of points: Triangle: " << m_triangle[i] << " for: " << getNbPoints() << " points.";
+            msg_warning() << "trianglesAroundVertex creation failed, Triangle buffer is not consistent with number of points, Triangle: " << m_triangle[i] << " for: " << getNbPoints() << " points.";
             continue;
         }
 
--- a/Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/EdgeSetTopologyContainer.cpp
+++ b/Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/EdgeSetTopologyContainer.cpp
@@ -122,7 +122,7 @@ void EdgeSetTopologyContainer::createEdgesAroundVertexArray()
         
         if (edge[0] >= unsigned(nbPoints) || edge[1] >= unsigned(nbPoints))
         {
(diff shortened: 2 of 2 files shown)


Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/TriangleSetTopologyContainer.cpp
```diff
@@ -125,6 +125,13 @@ void TriangleSetTopologyContainer::createTrianglesAroundVertexArray()
     m_trianglesAroundVertex.resize(getNbPoints());
     for (size_t i = 0; i < m_triangle.size(); ++i)
     {
+        if (m_triangle[i][0] >= getNbPoints() || m_triangle[i][1] >= getNbPoints() || m_triangle[i][2] >= getNbPoints())
+        {
+            msg_warning() << "trianglesAroundVertex creation failed, Triangle buffer is not concistent with number of points: Triangle: " << m_triangle[i] << " for: " << getNbPoints() << " points.";
```
fredroy:
```suggestion
            msg_warning() << "trianglesAroundVertex creation failed, Triangle buffer is not consistent with number of points,  Triangle: " << m_triangle[i] << " for: " << getNbPoints() << " points.";
```

Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/TriangleSetTopologyContainer.cpp
```diff
@@ -125,6 +125,13 @@ void TriangleSetTopologyContainer::createTrianglesAroundVertexArray()
...
+            continue;
+        }
```
hugtalbot, line 132: should the componentState be invalid then?

fredroy (changes requested):

Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/TriangleSetTopologyContainer.cpp
hugtalbot: should it not break the loop ?

olivier-goury: It could yes! I just copy-pasted (included the typo on consistent!) what was done from EdgeSetTopologyContainer where there is exactly the same loop but for edges.

fredroy (approved):

epernod (commented):
Ok for me, but could you explain in which case you have this problem?
Setting the component to invalid is a good idea. Breaking the loop might create some bigger issue in components using the topology as it won't be set.

olivier-goury:
> Ok for me, but could you explain in which case you have this problem? Setting the component to invalid is a good idea. Breaking the loop might create some bigger issue in components using the topology as it won't be set.

We have this problem when using subtopologies defined using BoxROIs. The subTopo has fewer points than the original topology (Since it is a subset of it) but the triangles in that subtopo refer to nodes indexed with the numbering in the original topology that are out of bounds within the subtopo world. Not sure that is clear!

olivier-goury:
But for the component state set to invalid, I would actually leave it valid.
Indeed, it is just that the m_trianglesAroundVertex array won't be fully created, but the rest of the component remains valid and for example a forcefield can be defined on this container.

epernod:
> > Ok for me, but could you explain in which case you have this problem? Setting the component to invalid is a good idea. Breaking the loop might create some bigger issue in components using the topology as it won't be set.
> 
> We have this problem when using subtopologies defined using BoxROIs. The subTopo has fewer points than the original topology (Since it is a subset of it) but the triangles in that subtopo refer to nodes indexed with the numbering in the original topology that are out of bounds within the subtopo world. Not sure that is clear!

There is something un-catholic there... So if I understand well, you have triangle in your topology that refer to point out of bound to your current node mechanicalObject?
I don't understand how it can work. It will crash as soon as a component will try to access a dof passing through the topology no?

Sofa/Component/Topology/Container/Dynamic/src/sofa/component/topology/container/dynamic/TriangleSetTopologyContainer.cpp
hugtalbot: glups..

epernod: what? My télor is riche! 😆

olivier-goury:
> > > Ok for me, but could you explain in which case you have this problem? Setting the component to invalid is a good idea. Breaking the loop might create some bigger issue in components using the topology as it won't be set.
> > 
> > 
> > We have this problem when using subtopologies defined using BoxROIs. The subTopo has fewer points than the original topology (Since it is a subset of it) but the triangles in that subtopo refer to nodes indexed with the numbering in the original topology that are out of bounds within the subtopo world. Not sure that is clear!
> 
> There is something un-catholic there... So if I understand well, you have triangle in your topology that refer to point out of bound to your current node mechanicalObject? I don't understand how it can work. It will crash as soon as a component will try to access a dof passing through the topology no?

Sorry for the late reply!
So it is a case where the node where the subtopology is defined does not contain a MechanicalObject. It is only defined to add a ForceField applying onto a subpart of the node of the MechanicalObject. An example is in this scene: https://github.com/SofaDefrost/ModelOrderReduction/blob/master/tools/test/sofa_test_scene/quadruped.py

alxbilger:
I switched this PR back to "to review". It was in WIP, that's why the lack of activity. @epernod can you take a look on @olivier-goury 's answer?

epernod (approved):

epernod:
ok for me if it fix your case and it had a warning if we encounter the wrong behavior.

epernod:
[ci-build][with-all-tests]

PR #4247, opened 2023-10-20
[MatrixAccumulator] adds 6x6 matrix handling

EulalieCoevoet:
Needed for https://github.com/sofa-framework/BeamAdapter/pull/115


alxbilger (changes requested):
You can also add the support of 6x6 matrices in `DerivativeMatrix::DerivativeElement`. Right now, it uses the templated method which calls `matAdd` (also template). But if you add

```c++
void operator+=(const sofa::type::Mat<6, 6, float> & value) const { mat->add(row, col, value); }
void operator+=(const sofa::type::Mat<6, 6, double>& value) const { mat->add(row, col, value); }
```

it will call the `add` method you just introduced

alxbilger:
[ci-build][with-all-tests]

alxbilger (approved):

PR #4251, opened 2023-10-23
[Type, Helper] Remove unused __STL_MEMBER_TEMPLATES parts

fredroy:
It seems to come from some very old version of the STL from SGI (google is not very prolix on this)
In any case, this is not defined by the STL of msvc, libc++ (llvm) and libstdc++ (gnu gcc)


hugtalbot (approved):

alxbilger (commented):
I agree to remove the macro, but why removing the template? I think it matches better the constructors available in the STL.

fredroy:
> I agree to remove the macro, but why removing the template? I think it matches better the constructors available in the STL.

I was wondering the same, but everything was tested with ` __STL_MEMBER_TEMPLATES`  not defined for years so thats why I prefered to use the "legacy" code. But I can change and take the templated one otherwise 😶‍🌫️

alxbilger:
[ci-build][with-all-tests]

alxbilger:
> > I agree to remove the macro, but why removing the template? I think it matches better the constructors available in the STL.
> 
> I was wondering the same, but everything was tested with ` __STL_MEMBER_TEMPLATES` not defined for years so thats why I prefered to use the "legacy" code. But I can change and take the templated one otherwise 😶‍🌫️

alright, I am not against it. It's not that important

bakpaul (approved):

PR #4257, opened 2023-10-25
[all] Externalize nlohmann_json dependency with find or fetch mechanism

olivier-roussel:
Tackles one point of #4221.
Replace embedded version of nlohmann_json library by an external one, using the modern "find or fetch" mechanism from CMake >=v3.24 (see https://cmake.org/cmake/help/latest/guide/using-dependencies/index.html#fetchcontent-and-find-package-integration ), similarly to what has been done in https://github.com/sofa-framework/sofa/pull/4240 for TinyXML2.
This is not required by conda-forge for the SOFA conda package as the library is header-only, only recommanded.
We also discussed to remove this dependency as only used by AdvancedTimer, but meanwhile, if not removed by the next-release, we might consider this PR or not. Let me know.
As for other dependencies that will be handled by this "find or fetch" mechanism, additional work needs to be done to forbid the fetch if the find package failed (also required for the next release).


Sofa/framework/Config/cmake/Modules/FindJson.cmake
```diff
@@ -1,86 +1,13 @@
...
-
-    include(FindPackageHandleStandardArgs)
-    find_package_handle_standard_args(Json DEFAULT_MSG JSON_INCLUDE_DIR JSON_VERSION_OK)
-
-    mark_as_advanced(JSON_INCLUDE_DIR)
-endif()
+FetchContent_Declare(
+  nlohmann_json
+  URL https://github.com/nlohmann/json/releases/download/v3.11.2/json.tar.xz
+  FIND_PACKAGE_ARGS NAMES nlohmann_json
```
alxbilger, line 10: In the url, we are looking for the version v3.11.2 specifically. But `find_package` does not seem to be restricted to any version. Do we want that?

olivier-roussel: I agree in some way. 
On the other hand, I found that controlling the version for both with an exact requirement on the find_package for the same version very restrictive and would make dependencies requirement too strict. Maybe adding some version range with at least a min version (to be specified) and eventually a max version synchronized to the same than fetched one (so here 3.11.2) ?

olivier-roussel:
Closed as :
- "find or fetch" mechanism has been abandoned
- embedded header-only libraries are acceptable and compatible with conda-forge package requirements
- the only feature using this dependency **might** be removed in the future (or will be based on another mechanism)

PR #4272, opened 2023-10-30
[all] Fix headeronly extlibs licenses

olivier-roussel:
This PR fixes a requirement of https://github.com/conda-forge/staged-recipes/pull/23085 for the SOFA conda package on conda-forge.
It is required that header-only embedded libraries have their license in a separate file, that can be listed then in the conda package recipe.
This PR adds a missing license file for json nlohmann library (MIT) and homogenize the license file name for the two others SOFA header-only embedded libraries (difflib and stb, both also on MIT license). 
Cxxopts, which the last header-only embedded library in SOFA, will be treated separately in another PR to handle the code fetch case through cmake FetchContent.


fredroy (approved):
Looks fine to me but I guess this is up to @hugtalbot or @bakpaul to merge it 🧐

bakpaul:
Why the '.MIT' for the license file ? Is it required by conda ? Because I can see that being cumbersome for windows user to have to define an opening rule for each license type (eventhough I know no one read such a file...). Not very important in fact, but may be just having a '.txt' or nothing at all (as in the most cases) me be better ?

olivier-roussel:
Three reasons:
- It was named as-it in the json nlohmann repository, so I basically just picked their license file
- It is convenient (but not required by conda-forge) as all license files must be listed explicitly in the conda package recipe, so basically you know which kind of license you have with your package and all its embedded libs without having to dig each file
- I chose to rename the two other license files (diff lib and stb) accordingly, as they are both also MIT so looks a good idea to be homogeneous here (and in any case there was a typo in the initial license names, it was spelled the french way)

bakpaul (approved):

PR #4273, opened 2023-10-31
[all] Externalize cxxopts

olivier-roussel:
This PR fixes a requirement of https://github.com/conda-forge/staged-recipes/pull/23085 for the SOFA conda package on conda-forge.
This add a find package to the cmakelists to avoid the cmake fetch of cxxopts sources if a cxxopts is already available and found. This is a simplified version of a find or fetch mechanism which looks enough here, and does not require cmake >= 3.24 (which embedds a more sophisticated find or fetch within an improved FetchContent set of macros).
The PR also adds a cmake SOFA option, which may be re-used if other dependencies will be managed by a similar mechanism, to allow or disable the fetch content of cmake (default behavior is to allow fetching to keep compatibility). 
This is required by the conda-forge PR to disable the fetching of source.


fredroy (approved):
Could be nice to describe the list of the  fetched deps in the comment (or an other visible place)
 Otherwise the user will think it will fetch all dependencies (boost, eigen, qt... etc)

olivier-roussel:
> Could be nice to describe the list of the fetched deps in the comment (or an other visible place) Otherwise the user will think it will fetch all dependencies (boost, eigen, qt... etc)

Agreed. Done.

CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${LIBRARY_OUTPUT_DIRECTOR
 # Option for packaging
 option(SOFA_BUILD_RELEASE_PACKAGE "Run package specific configure" OFF)
 
+# Option to allow some dependencies such as cxxopts to be fetched by cmake if
+# the package is not found
+option(SOFA_ALLOW_FETCH_DEPENDENCIES "Allow compatible dependencies to be fetched is the package is not found by cmake.
```
epernod:
```suggestion
option(SOFA_ALLOW_FETCH_DEPENDENCIES "Allow compatible dependencies to be fetched if the package is not found by cmake.
```

CMakeLists.txt
```diff
@@ -34,6 +34,11 @@ set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/${LIBRARY_OUTPUT_DIRECTOR
...
+  Fetchable dependencies: cxxopts" ON)
```
epernod:
```suggestion
  List of dependencies that can be fetched: cxxopts" ON)
```

epernod (commented):
small typo suggestions

PR #4290, opened 2023-11-09
[Sofa.GUI.Qt] Add cmake module for QGLViewer

olivier-roussel:
This PR adds a CMake find module for the QGLViewer used in SOFA.GUI.Qt.
The behavior here (already in place) is to look for an external package of QGLViewer and if not found, use the embeded code of QGLViewer (which is linked dynamically).
I may have missed something but it seems that QGLViewer does not provide CMake config files (neither from source, or in package managers such as APT, Conda and vcpkg). And as SOFA was not including a CMake module for it, there was no chance for it to be found externally.
Then, this should not break current behavior but just fix the possibility to use an external GLViewer package.


olivier-roussel:
[ci-build][with-all-tests]

olivier-roussel:
[ci-build][with-all-tests]

cmake/Modules/FindQGLViewer.cmake
```diff
@@ -0,0 +1,69 @@
...
+    endif(NOT QGLViewer_FIND_QUIETLY)
+
+    if(NOT TARGET QGLViewer)
+      add_library(QGLViewer INTERFACE IMPORTED)
+      set_property(TARGET QGLViewer PROPERTY INTERFACE_LINK_LIBRARIES "${QGLViewer_LIBRARIES}" ${QT_TARGETS})
+      set_property(TARGET QGLViewer PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${QGLViewer_INCLUDE_DIR}")
+    endif()
+  endif()
+  mark_as_advanced(QGLViewer_INCLUDE_DIR QGLViewer_LIBRARY)
+endif()
```
hugtalbot:
```suggestion
endif()

```

PR #4302, opened 2023-11-17
[Constraint] Make name of constraints more explicit

bakpaul:
This PR is a first step for the issue #3952. 

I modified the names of the constraints following the rule that we agreed upon. Still need to change spring force-fields that can behave as penalties. 

The changes where straight, not more though than adding the type into the name was done. If any name doesn't fit your standard, feel free to suggest any changes. 

This PR also add a componentRenamed mechanism that creates an alias automatically for renamed objects.

TODO:

- [x]  Edit Sofa/framework/Helper/src/sofa/helper/ComponentChange.cpp
  - [x]  Add a API in ComponentChange.cpp for Renamed component (avoiding to add an alias manually in the class)
- [x] Move the C++ alias in the old header
- [x] Change name of projective constraints strating by `ProjectOn`

PR #4324, opened 2023-11-29
[All] Changed message for scene check

bakpaul:
Based on
- #4302 

The map `uncreatableComponents` is used both by the factory when the object could not be created and by the alias check. So, cometimes, mif you used an alias of a component that has also been moved, a non sens was displayed.

This pr divide this map in  two : `uncreatableComponents` and `movedComponents` and only use the `uncreatableComponents` in scene check and also display the original message.


bakpaul:
I might have emmêlé mes pinceaux with the "based on". It is based on because I've added the class `Renamed` in the other PR, but in the end, it would have been more logical to do it separately. What you really have to look at are the commits [1384deb](https://github.com/sofa-framework/sofa/pull/4324/commits/1384deba11b6b402dd4318ca14b1c215e31520cc) and [7bf84de](https://github.com/sofa-framework/sofa/pull/4324/commits/7bf84dee24d6a66224b94390d7db22c4394a33db). The rest are on the other PR

bakpaul:
[ci-build][with-all-tests]

bakpaul:
No there are two, I've modified my first message, this one and the next one, but the real improvement is in the one you pointed exactly !

PR #4327, opened 2023-12-01
[Simulation.Core] Deprecate LocalStorage feature

fredroy:
This PR deprecates LocalStorage and its (only) implementation CactusLocalStorage.
It seems it was planned for an eventual parallelization of the visitor traversal. Since then, it is not really relevant anymore (link, etc)

Present since the beginning (16 years ago!) but it seems it was never used,
as the only implementation of LocalStorage (CactusLocalStorage) is never instanced anywhere.

IMO we could even directly remove it as it is 99,99...9% sure nobody uses it.


bakpaul:
Well you did the deprecation job. But you're right, this could disappear totally...

bakpaul (approved):

alxbilger (approved):

PR #4328, opened 2023-12-01
[Simulation.Core] Remove usage of ill-used nodeData in MechanicalGetNonDiagonalMassesCountVisitor and MechanicalVDotVisitor

fredroy:
First task of removing(deprecating) the usage of nodeData (noticed in #4327)

Those two visitors are using the pointer nodeData for local computations so It was changed to local (weak) stored pointers. 
Not fan of pointers but it does not break API. (a SReal to count is weird btw)

### DIFF
--- a/Sofa/framework/Simulation/Core/src/sofa/simulation/mechanicalvisitor/MechanicalGetNonDiagonalMassesCountVisitor.h
+++ b/Sofa/framework/Simulation/Core/src/sofa/simulation/mechanicalvisitor/MechanicalGetNonDiagonalMassesCountVisitor.h
@@ -30,9 +30,12 @@ namespace sofa::simulation::mechanicalvisitor
 class SOFA_SIMULATION_CORE_API MechanicalGetNonDiagonalMassesCountVisitor : public MechanicalVisitor
 {
 public:
-    SReal* const m_nbNonDiagonalMassesPtr { nullptr };
+    sofa::Size* const m_nbNonDiagonalMassesPtr { nullptr };
 
-    MechanicalGetNonDiagonalMassesCountVisitor(const sofa::core::MechanicalParams* mparams, SReal* result)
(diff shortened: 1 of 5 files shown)


fredroy:
[ci-build][with-all-tests]

alxbilger (changes requested):
I think the counter is SReal only because of the use of `nodeData`. If you don't use it, I don't see any reason to switch to an integer.

bakpaul (approved):

fredroy:
> I think the counter is SReal only because of the use of `nodeData`. If you don't use it, I don't see any reason to switch to an integer.

To keep the API "non-breaking"  🙊 but I can make it an integer if we dont care ✌️

alxbilger:
Since this visitor is rather recent, and used at only one place. We can consider convert the SReal to an integer even if it's breaking

Sofa/framework/Simulation/Core/src/sofa/simulation/mechanicalvisitor/MechanicalGetNonDiagonalMassesCountVisitor.h
```diff
@@ -30,10 +30,14 @@ namespace sofa::simulation::mechanicalvisitor
 class SOFA_SIMULATION_CORE_API MechanicalGetNonDiagonalMassesCountVisitor : public MechanicalVisitor
 {
 public:
-    MechanicalGetNonDiagonalMassesCountVisitor(const sofa::core::MechanicalParams* mparams, SReal* result)
-            : MechanicalVisitor(mparams)
+    sofa::Size* const m_nbNonDiagonalMassesPtr { nullptr };
+
+    // given result is not a Real anymore since https://github.com/sofa-framework/sofa/pull/4328
```
alxbilger:
```suggestion
```

PR #4370, opened 2023-12-15
[LinearSolver.Direct] Introduce other ordering methods in SparseLDL through Eigen

alxbilger:
This gives more choices for the ordering method. It mimics the choices we have in Eigen-based linear system (e.g. EigenSimplicialLDLT).

The data applyPermutation is deprecated

Notes:
- It means we have an alternative to Metis for all solvers. However, it may not be as efficient as Metis.
- The new implementation may be less efficient (or not, I haven't tested) than calling directly the Metis function. But this call is made only once, or every time the matrix pattern changes (topological changes etc).
- Since we have multiple choices, we can put all of them in a factory. That way, Metis could be moved away from this module. Another module can be created, which depends on Metis and would populate the factory with Metis methods. This could also be the case with Scotch.


alxbilger:
[ci-build][with-all-tests]

alxbilger:
⚠️ ⚠️ ⚠️ For now, some unit tests on StaticSolver are failing.

bakpaul:
[ci-build][with-all-tests]

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolverImpl.h
```diff
@@ -166,14 +165,20 @@ public :
 protected :
 
     Data<bool> d_precomputeSymbolicDecomposition; ///< If true the solver will reuse the precomputed symbolic decomposition. Otherwise it will recompute it at each step.
-    Data<bool> d_applyPermutation; ///< If true the solver will apply a fill-reducing permutation to the matrix of the system.
+    core::objectmodel::lifecycle::DeprecatedData d_applyPermutation;
```
hugtalbot: I would rather suggest:
```suggestion
    core::objectmodel::lifecycle::DeprecatedData d_applyPermutation{this, "v24.06", "v24.12", "applyPermutation", "Use the Data 'ordering'"};
```

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolverImpl.h
```diff
@@ -166,14 +165,20 @@ public :
...
+    Data<sofa::helper::OptionsGroup> d_orderingMethod; ///< Ordering method
     Data<int> d_L_nnz; ///< Number of non-zero values in the lower triangular matrix of the factorization. The lower, the faster the system is solved.
 
+    static constexpr const char* s_defaultOrderingMethod { "Metis" };
+
     SparseLDLSolverImpl() : Inherit()
     , d_precomputeSymbolicDecomposition(initData(&d_precomputeSymbolicDecomposition, true ,"precomputeSymbolicDecomposition", "If true the solver will reuse the precomputed symbolic decomposition. Otherwise it will recompute it at each step."))
-    , d_applyPermutation(initData(&d_applyPermutation, true ,"applyPermutation", "If true the solver will apply a fill-reducing permutation to the matrix of the system."))
+    , d_applyPermutation(this, "v24.06", "v24.12", "applyPermutation", "Use the Data 'ordering'")
```
hugtalbot:
```suggestion
```

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolverImpl.h
```diff
@@ -227,29 +232,57 @@ protected :
...
+                computeOrdering(Eigen::MetisOrdering<int>());
+            }
+            else if (orderingMethod == "COLAMD")
+            {
+                computeOrdering(Eigen::COLAMDOrdering<int>());
+            }
+            // else if (orderingMethod == "AMD")
+            // {
+            //     computeOrdering(Eigen::AMDOrdering<int>());
+            // }
```
hugtalbot: AMD option was not working correct @alxbilger ?

hugtalbot (changes requested):

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolverImpl.h
alxbilger: The option does not compile on my machine. But maybe it could be done with more efforts.

alxbilger:
Create an issue for the missing AMD method after it is merged

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolverImpl.h
alxbilger: The option AMD now compiles

alxbilger:
The unit tests on the StaticSolver failed because the Metis permutation using the Eigen API did not give the same result than the Metis permutation that was implemented before. This is intriguing... But the most concerning problem is that the result of the linear solver depends on the permutation. Hopefully, it is only a problem of numerical errors. Afterall, the operations are not the same depending on the permutation.

PR #4430, opened 2024-01-12
[SofaCUDA] Generalize the use of MessageDispatcher in SofaCUDA

alxbilger:
`msg_info` is used instead of `std::cout`. `msg_info` is also used in `mycudaPrintf`


hugtalbot (approved):

alxbilger:
[ci-build][with-all-tests]

alxbilger:
@hugtalbot 
My guess is the following:
When loading the SofaCUDA plugin, the module check if it can detect a cuda device. If there is no such a device, it now triggers an error, whereas it was just a print before. In other words, the behavior does not change (loading of the plugin), but the log message is now counted as an error. 
I think it is a good thing. It helps to identify the scenes that cannot be executed on the CI. For the case of the failing scene. I'll just remove the `<RequiredPlugin name="SofaCUDA"/>` which is even not required.

hugtalbot (approved):

PR #4266, opened 2023-10-27
[all] Externalize metis

olivier-roussel:
Tackles one point of #4221. Required for the conda-forge package (https://github.com/conda-forge/staged-recipes/pull/23085).
Externalize the previously embedded version of metis (5.1.0) which was statically linked.

Provides a cmake module finder file (FindMetis.cmake), tested successfully with the following ways to provide the library:
- compile & install from source (v5.1.1) (warning: official repository cmakefiles are broken)
- APT package libmetis-dev (v5.1.0)
- conda package (v5.1.0 and v5.1.1)
- vcpkg package (v2022-07-27)
All on ubuntu 22.04. Not tested with conan.

Some remarks to embed metis in the CI: The embedded version was compiled with GKlib using the `GKRAND` option. To the best of my knowledge, this can be provided either from compiling from source or with the conan package with a specific option.

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #4450, opened 2024-01-22
[LinearSolver.Direct] Fix metis dependency

olivier-roussel:
Continuation of #4266 

As required by https://github.com/conda-forge/staged-recipes/pull/23085, we must avoid static libraries to comply conda-forge rules.
Tests with more recent versions of metis (such as one available on conda-forge) seems to introduce some bugs (@bakpaul @alxbilger Do you have some inputs to provide here ?).
For now, the solution we converged to is to dynamically link metis with Sofa.


alxbilger:
Yes, I don't have the exact commit but, after the version 5.1.0, some changes lead to a crash. Before the changes, it works.
It has been fixed later in this commit: https://github.com/KarypisLab/METIS/commit/36262adecaa9720a4417a67124428061c367fd3f. It means the versions in between both commits cannot be used without a crash.

Note also that the version number (in the code) has not been updated properly along the history of the repository: https://github.com/KarypisLab/METIS/commit/f5ae915a84d3bbf1508b529af90292dd7085b9ec. It went from 5.1.0 to 5.2.1, whereas there is a 5.1.1 release.

Finally, less importantly, in SOFA we need a portable random generator in METIS for reproductible results. It is enabled via a compilation option `METIS-GKLIB_GKRAND`. This is not enabled by default in releases.

fredroy:
[ci-build][with-all-tests]

olivier-roussel:
The bug introduced in metis somewhere between v5.1.0 and v5.2.1 looks similar to https://github.com/conda-forge/gtsam-feedstock/pull/21. 
As it is fixed in v5.2.1, we should probably wait that this version comes in conda forge, there is a quite recent and active PR on this: https://github.com/conda-forge/metis-feedstock/pull/41.

olivier-roussel:
@bakpaul This would mean that you don't need to port v5.1.0 branch on your fork, we would rely on v5.2.1 both for external package (including conda) and for the fetch of your fork, which already has a v5.2.1 branch.

olivier-roussel:
@bakpaul I'm afraid the current PR to have a metis 5.2.1 package on conda-forge (https://github.com/conda-forge/metis-feedstock/pull/41) is stalling... and as the official repository is not active anymore (PR not merged anymore neither) and the official build chain is broken, it can take a lot of time before a 5.2.1 comes up on conda-forge. 
I think we would better be safe and ensure that we have a working solution, so we would need you to backport the changes of your metis fork on the 5.1.0 branch, in order to have a consistent version of used metis, wether it's external or fetched dependency.

bakpaul:
I've added the `v5.1.0-ModernInstall` branch to our fork. Actually this code wasn't on the repo itself so I just copied the files presents in SOFA and added my CMake configuration files. It compiles, you can try to add the find or fetch mechanism using this

Sofa/framework/Config/cmake/Modules/FindMetis.cmake
```diff
@@ -0,0 +1,108 @@
...
+    set(Metis_VERSION_OK FALSE)
+    message(SEND_ERROR "Metis version ${Metis_VERSION} found in ${Metis_INCLUDE_DIR}, "
+                       "but at least version ${Metis_FIND_VERSION} is required")
+  endif()
+  if(${Metis_FIND_VERSION_EXACT} AND NOT ${Metis_VERSION} VERSION_EQUAL ${Metis_FIND_VERSION})
+    set(Metis_VERSION_OK FALSE)
+    message(SEND_ERROR "Metis version ${Metis_VERSION} found in ${Metis_INCLUDE_DIR}, "
+                       "but exact version ${Metis_FIND_VERSION} is required")
+  endif()
+  # message(STATUS "Metis version found: ${Metis_VERSION} in ${Metis_INCLUDE_DIR}, ${Metis_FIND_VERSION} was required ")
```
alxbilger: Do you want to message or not? If not, please remove this line

bakpaul:
We will also need to install metis library to distribute it with the binaries. I didn't have time to make it work, I'll try tomorrow.

bakpaul:

bakpaul:
This is not ready, the find or fetch mechanism isn't working well.
I also supposed that the metis linkage could be made private to avoid problem with libraries linking to LinearSolver.Direct trying to find metis but some of them really depends on metis itself and not only on LinearSolver.Direct and so requires the Public linkage. This makes this find or fetch mechanism not really working because those libs will not find the metis package required by Sofa.Component.LinearSolver.DirectConfig.cmake in the current state. 
We should either 
- fix the find or fetch mechanism to be able to find this target after it has been fetched and made available for other targets (and also make sure it is distributed correctly at the end)
- fix the find metis file and again make sure the lib is correctly distributed
- revert everything and rely on the old embeded metis if it isn't found -> this had the advantage of making the process of distributing it easier

bakpaul:
So now the linkage of metis is PRIVATE in `Sofa.Component.LinearSolver.Direct` and every other target requiring metis finds the package itself (e.g. SofaMatrix) .

One problem is that FetchContent_MakeAvailable make the target available for the current target and its child (that is why the test in `Sofa.Component.LinearSolver.Direct` is able to link to metis without the need of finding it) but other unrelated targets cannot find it. So we cannot keep it Public because it will not be found by the other target linking to `Sofa.Component.LinearSolver.Direct` because a lot of them doesn't require metis

applications/plugins/SofaMatrix/CMakeLists.txt
```diff
@@ -8,11 +8,26 @@ sofa_find_package(Sofa.Component.LinearSolver.Direct REQUIRED)
...
+    message(FATAL_ERROR "SofaMatrix: DEPENDENCY metis NOT FOUND. SOFA_ALLOW_FETCH_DEPENDENCIES is OFF and thus cannot be fetched. Install metis (version=5.1.0), or enable SOFA_ALLOW_FETCH_DEPENDENCIES to fix this issue.")
+endif()
+
+
 set(SOFA_MODULES
     Sofa.Component.Constraint.Lagrangian.Solver
     Sofa.Component.LinearSolver.Direct
     Sofa.Core
     Eigen3::Eigen
+    Sofa.GUI.Qt
```
fredroy:
```suggestion
```

fredroy: it is added later if it is actually found

fredroy (changes requested):
Looks fine to me otherwise  🤷‍♂️

fredroy (approved):

PR #4459, opened 2024-01-25
[Sofa.Testing] Do not build Sofa.Testing if SOFA_BUILD_TESTS is OFF

olivier-roussel:
... unless there is a good reason to keep built even if tests are not built ?
In the case we have to keep it built, we will have to handle the embedded dependency google test (linked dynamically), probably in a similar manner than other libs such as metis (find package or fetch code and link dynamically).


alxbilger (approved):
It makes sense. In a similar manner, the `.Testing` modules in `Sofa.Component` also depend on the variable `SOFA_BUILD_TESTS`:

In Sofa/Component/Engine/CMakeLists.txt:
```
if(SOFA_BUILD_TESTS)
    add_subdirectory(Testing)
endif()
```

bakpaul (approved):

olivier-roussel:
As we just discussed, let's wait if #4471 is fine, in which case this PR should be closed and discarded: we would keep building Sofa.Testing component even if TU are not built (`SOFA_BUILD_TESTS=OFF`)

olivier-roussel:
Any reason to have re-opened this ? I thought we all agreed that this PR would be discarded and closed as the initial behavior of having Sofa.Testing built for other plugins TU is OK and the problem with conda was fixed thanks to #447.

PR #4464, opened 2024-01-29
[Helper] PluginManager: testing loading a plugin with a dependency on an other plugin

fredroy:
Illustrating #4462 

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #4466, opened 2024-01-29
[Helper] PluginManager: Check symbol (real) location and avoid calling wrong entrypoint (Mac/Linux)

### LINKED ISSUE #4462: PluginManager, dynamic libraries and entry points flaw
### Premises:
 - plugin B is a dependency of plugin A (i.e A links dynamically against B)
 - plugin B implements `entrypoint()` but plugin A does not implement it
- PluginManger is set to load plugin A (which will then implicitly load plugin B)

Then PluginManger, when trying to check if plugin A does implement `entrypoint()` will find `B::entrypoint()`, instead of returning a nullptr (indicating that the plugin is supposed to not implement it). Obviously this is not the desired outcome 🐛 

My reasoning would be that as the plugin B is direct dependency of A, when loaded by `dlopen()` (unix), is seen as an integral part of plugin A from a system-call point of view.

This has been tested with Ubuntu. Did not test on MacOS but should be the same, as it uses dlopen too.
It seems this behavior is not present on Windows, the following demo does not show the bogus behavior.

---

### Demo:
- adding `getModuleLicense()` which is supposed to be not implemented in Sofa.Component.StateContainer
- StateContainer is a dependency for some other Sofa.Component libraries

```
extern "C" SOFA_EXPORT_DYNAMIC_LIBRARY const char* getModuleLicense();
const char* getModuleLicense()
{
    return "StateContainer License";
}
```
Result (Qt PluginManager windows calls `getModuleLicense()` if entry is supposed to be found)

We can see the cascading wrong call to  `getModuleLicense()`, as Collision has a dependency on Collision.Contact which has a dependency on StateContainer.
Moreover in this image, we can even see the `LGPL` which has been leaking from the recently added S.C LinearSystem to several other components.

---

### Workaround
The workaround would be to implement obviously all the potential entrypoints in all the plugins, but the problem will arise again if a new entrypoint is added in the future....
At first glance, I thought that setting `RTLD_LOCAL`  here
https://github.com/sofa-framework/sofa/blob/1e0df59953a085785a45bf97e182e01efea0c354/Sofa/framework/Helper/src/sofa/helper/system/DynamicLibrary.cpp#L78
would fix the bug, but it is still showing the same behavior 😓

fredroy:
Based on
- #4464

Fix 
 - #4462 

This PR check the provenance of the loaded symbol, if it is consistent with the given filename of the plugin
Using the system function `dladdr()` https://man7.org/linux/man-pages/man3/dladdr.3.html

Fix the newly introduced (failing) test in #4464


fredroy:
[ci-build][with-all-tests]

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
```diff
@@ -199,17 +199,21 @@ PluginManager::PluginLoadStatus PluginManager::loadPluginByPath(const std::strin
             if (errlog) (*errlog) << msg << std::endl;
             return PluginLoadStatus::MISSING_SYMBOL;
         }
-        getPluginEntry(p.getModuleName,d);
+
+        if(!getPluginEntry(p.getModuleName,d))
+        {
+            dmsg_warning("PluginManager") << DynamicLibrary::getLastError();
```
alxbilger, line 205: I am not sure the triggered log message is clear here for a newbie. I am talking about `is different from where it was supposed to be loaded`. Can you try to improve it?

And why do you add a warning here and not for `getModuleDescription`, `getModuleLicense`, ... ?

alxbilger (changes requested):

alxbilger:
The diff: https://github.com/fredroy/sofa/compare/add_secondplugin_test...fredroy:sofa:fix_dlsym_symbols

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
fredroy: > I am not sure the triggered log message is clear here for a newbie. I am talking about `is different from where it was supposed to be loaded`. Can you try to improve it?
> 
I guess I can 🫠

> And why do you add a warning here and not for `getModuleDescription`, `getModuleLicense`, ... ?

Because moduleName is "important" because it is used as "plugin name" and is used by PluginManager::getPluginByName(), pluginIsLoaded() among others. 
Description, license, etc is only used by GUIs.

Actually this is something I was wondering while doing this PR, should we request the get*() to be mandatorily implemented ?

alxbilger:
I was wondering if there is a way to catch the missing symbols at compile-time and I found this: https://cmake.org/cmake/help/latest/module/CheckCXXSymbolExists.html#module:CheckCXXSymbolExists. It's configuration time but it could be helpful. I tried to play with it but I was not able to make it work. I added the check in the `sofa_create_package_with_targets` macro.

fredroy:
> I was wondering if there is a way to catch the missing symbols at compile-time and I found this: https://cmake.org/cmake/help/latest/module/CheckCXXSymbolExists.html#module:CheckCXXSymbolExists. It's configuration time but it could be helpful. I tried to play with it but I was not able to make it work. I added the check in the `sofa_create_package_with_targets` macro.

It could be interesting to enforce that the necessary symbols are defined at configure-time yes 🤔 

EDIT: I did some tests and it is quite bothersome to set up, (the file itself needs to be compilable, the symbol need to be defined and not only declared, adding variable in cache, etc) And I think if we do that for all potential loadable libraries, it will slow down considerably the cmake configure process.

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
```diff
@@ -199,17 +199,21 @@ PluginManager::PluginLoadStatus PluginManager::loadPluginByPath(const std::strin
...
         }
 
-        getPluginEntry(p.getModuleDescription,d);
-        getPluginEntry(p.getModuleLicense,d);
-        getPluginEntry(p.getModuleComponentList,d);
-        getPluginEntry(p.getModuleVersion,d);
+        [[maybe_unused]] const auto moduleDescriptionResult = getPluginEntry(p.getModuleDescription,d);
+        [[maybe_unused]] const auto moduleLicenseResult = getPluginEntry(p.getModuleLicense,d);
+        [[maybe_unused]] const auto moduleComponentListResult = getPluginEntry(p.getModuleComponentList,d);
+        [[maybe_unused]] const auto moduleVersionResult = getPluginEntry(p.getModuleVersion,d);
```
hugtalbot, line 216: This will have to be checked in another PR

PR #4471, opened 2024-01-31
[Testing] Externalize (find_package() or fetch) googletest/gtest

fredroy:
Originally, the goal was not really to go for 
- #4221

But more to upgrade the gtest version, as the one in SOFA is really old (1.7.0 , *Sep 19, 2013* :exclamation: ) and was throwing warnings for newer cmake.
So instead of simply update the source in extlibs, I took the liberty to go into the find/fetch mechanism.

A bit difficult as the CMakefile in the embedded version was heavily customized, so I am not 100% sure the install/package process is working well. But it should 😅

Last note: it is adding a target `gtest_main` which is always activated in the gtest cmake file. It is not used by SOFA as we have our own main for gtest.

[ci-depends-on https://github.com/sofa-framework/PluginExample/pull/5]


fredroy:
[ci-build][with-all-tests]

fredroy:

Sofa/framework/Testing/CMakeLists.txt
```diff
@@ -4,10 +4,54 @@ project(Sofa.Testing LANGUAGES CXX)
 # Enable testing features of cmake, like the add_test() command.
 enable_testing()
 
-# add gtest library
-# set(gtest_disable_pthreads ON CACHE INTERNAL "")
-add_subdirectory(extlibs/gtest)
-# add custom gtest main
+# add googletest library
+# set(googletest_disable_pthreads ON CACHE INTERNAL "")
+find_package(googletest QUIET)
```
olivier-roussel: CMake config/module files uses the GTest namespace for googletest:
https://cmake.org/cmake/help/latest/module/FindGTest.html

olivier-roussel: `find_package(GTest QUIET)`

Sofa/framework/Testing/CMakeLists.txt
```diff
@@ -4,10 +4,54 @@ project(Sofa.Testing LANGUAGES CXX)
...
+if(NOT googletest_FOUND AND SOFA_ALLOW_FETCH_DEPENDENCIES)
```
olivier-roussel: `if(NOT GTest_FOUND AND SOFA_ALLOW_FETCH_DEPENDENCIES)`

Sofa/framework/Testing/CMakeLists.txt
```diff
@@ -4,10 +4,54 @@ project(Sofa.Testing LANGUAGES CXX)
...
+            PACKAGE_VERSION 1.14.0
+            TARGETS gtest AUTO_SET_TARGET_PROPERTIES
+            INCLUDE_SOURCE_DIR "include"
+            INCLUDE_INSTALL_DIR "extlibs/GTest"
+            )
+
+        set_target_properties(gtest PROPERTIES FOLDER Testing)
+        set_target_properties(gtest_main PROPERTIES FOLDER Testing)
+    endif()
+elseif (NOT googletest_FOUND)
```
olivier-roussel: `elseif (NOT GTest_FOUND)`

Sofa/framework/Testing/CMakeLists.txt
```diff
@@ -39,6 +83,8 @@ sofa_find_package(Sofa.Simulation.Graph REQUIRED)
 
 add_library(${PROJECT_NAME} SHARED ${HEADER_FILES} ${SOURCE_FILES})
 target_link_libraries(${PROJECT_NAME} PUBLIC Sofa.Helper Sofa.DefaultType Sofa.Core Sofa.Simulation.Graph gtest SofaGTestMain )
```
olivier-roussel: Use `GTest::gtest` target instead of `gtest`. This is the one defined by the cmake file for external package and which is also handled by the `sofa_create_package_with_targets` for the case of using fetched code.

olivier-roussel: You must also update the same target that is used in `SofaGTestMain/CMakeLists.txt` for `SofaGTestMain`.

olivier-roussel (changes requested):
With the following adjustments, this compiles both using fetched gtest or using external gtest package from conda (tested on linux-x64 with gtest 1.14.0 from conda-forge).

fredroy:
@olivier-roussel  thanks for the conda test and the feedbacks 👍
You should have made suggestions and/or make a PR on my branch, as I cannot credit you your patches 😵

fredroy:

fredroy:
@olivier-roussel 
The findGTest https://cmake.org/cmake/help/v3.19/module/FindGTest.html of cmake only works for CMake >= 3.20 😭
(tested with cmake 3.16 and 3.19 ❌ whereas cmake 3.20 and 3.28 ✅)

olivier-roussel:
> @olivier-roussel thanks for the conda test and the feedbacks 👍 You should have made suggestions and/or make a PR on my branch, as I cannot credit you your patches 😵

Indeed, sorry about that

olivier-roussel:
> @olivier-roussel The findGTest https://cmake.org/cmake/help/v3.19/module/FindGTest.html of cmake only works for CMake >= 3.20 😭 (tested with cmake 3.16 and 3.19 ❌ whereas cmake 3.20 and 3.28 ✅)

Indeed again :cry: 
Another option but maybe a bit conda-specific would be to switch to config mode `find_package(GTest CONFIG QUIET)`, as the conda package comes with the GTest cmake config files. Tested with conda + cmake 3.19.

olivier-roussel:
Ok so maybe I misunderstood your last comment, but the GTest find module you added in your last commit works fine on my side with cmake >= 3.12 (which is our min required version). So I guess you meant that the embedded GTest find module in recent cmake version (>=3.20) was required to have the correct targets defined, which is not the case in earlier <3.20 version of cmake (tested in 3.12), but this recent GTest find module was not building if used with cmake <3.20. In the end, your last commit seems to fix everything, so looks like a more generic option that relying on cmake config files as suggested in my last comment.

fredroy:
> Ok so maybe I misunderstood your last comment, but the GTest find module you added in your last commit works fine on my side with cmake >= 3.12 (which is our min required version). So I guess you meant that the embedded GTest find module in recent cmake version (>=3.20) was required to have the correct targets defined, which is not the case in earlier <3.20 version of cmake (tested in 3.12), but this recent GTest find module was not building if used with cmake <3.20. In the end, your last commit seems to fix everything, so looks like a more generic option that relying on cmake config files as suggested in my last comment.

I just added the cmake module from v3.20 for GTest directly in our cmake folder modules; not very clean but it does the job apparently 👍

olivier-roussel (approved):

PR #4477, opened 2024-02-02
[Core][LinearSolver] Introduction of ordering method components

alxbilger:
`BaseOrderingMethod` is introduced as an abstract component in `Sofa.Core`. It is the base class of a new category of components.

Then, a new module is introduced. In this module, the components are derived from `BaseOrderingMethod`. Therefore, its name is `Sofa.Component.LinearSolver.Ordering`. It is in `LinearSolver` because an ordering method is only used with linear solvers, more specifically sparse direct linear solvers.

Only one ordering method is still in `LinearSolver.Direct`: `MetisOrderingMethod`. The goal is to move later this component in a plugin so that METIS is no longer a dependency of SOFA core. In the meantime, this method is still the preferred one for legacy reasons.

Eigen-based solvers were problematic. In Eigen, everything is defined at compile-time, and it is not possible to use runtime polymorphism. Therefore, it was not possible to uncouple the solver itself from the ordering method. To comply with the new design, I introduced a factory that can be enriched in the new Metis plugin. This factory is able to register types and use runtime polymorphism using a few C++ tricks. The only concern is that the solvers in the factory will not use the algorithms from the `BaseOrderingMethod`. Instead, they use a simple string (defined in a `BaseOrderingMethod`), and its mapping to a type. This has to be registered in the initialization of the module. I tried hard to use the intelligence of the component instead of relying on a string and a factory, but the design of Eigen is not compatible.

In order to not force the users to add an ordering method in their scenes, a default one is created automatically by the linear solver. It instantiates `MetisOrderingMethod` if it can find it, otherwise it's `AMDOrderingMethod`.

Here are the tasks in a future PR:

1. Create a plugin for Metis
2. Move `MetisOrderingMethod` into this plugin
3. Move some unit tests in this plugin
4. Update the regression references
5. Update `FillReducingOrdering`
6. Remove METIS dependency in `LinearSolver.Direct`
7. Create a new plugin for Scotch


alxbilger:
[ci-build][with-all-tests]

alxbilger:

PR #4484, opened 2024-02-07
[image] Fix compilation on Windows

alxbilger:
Plugin `image` fails (`Error C2672 : 'std::max': no matching overloaded function found`) when compiling `ImageFilter` with the first template parameter being `long` or `unsigned long`. This happens when `PLUGIN_IMAGE_COMPILE_SET` is set to `full`. The compilation fails only with MSVC.

The fix is not tested and may change the accuracy because of an additional conversion to `float`.


alxbilger:
[ci-build][with-all-tests]

fredroy (approved):

PR #4505, opened 2024-02-16
[Plugins] Gather external plugins configs in a same folder (proposal)

fredroy:
it is difficult to spot quickly which plugin is externalized or not in the centralized plugin folder.

So this PR gathers all of these into a same folder (namely `external_plugins`) and are added to the cmake configure process after the `plugins/` one

This implies that plugins in `plugins/` does not depend on things from `external_plugins/`, which should be normal as externalized plugins are supposed to be less a priority than the `plugins`. EXCEPT **SofaPython3** 🤔


fredroy:

PR #4509, opened 2024-02-20
[Simulation.Core] RequiredPlugin: Avoid calling loadPlugin() twice on start-up

fredroy:
With runSofa and SceneChecking, `RequiredPlugin` is calling `loadPlugin()` twice
 - once at the `parse()` step
 - and a second time when `SceneCheckMissingRequiredPlugin` at the `init()` step

`SceneCheckMissingRequiredPlugin` browses all `RequiredPlugin`  and checks each entry on `d_loadedPlugins`.
https://github.com/sofa-framework/sofa/blob/af8f98a668dee73e21fac19da31d181baf137a99/applications/projects/SceneChecking/src/SceneChecking/SceneCheckMissingRequiredPlugin.cpp#L144-L150
But `d_loadedPlugin`s is set dirty at parse() so calling `d_loadedPlugins.getValue()` will trigger an other `loadPlugin()`, because of the callback.

So this PR forcefully sets `d_loadedPlugins` to **not dirty** but maybe there is a better solution....


fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

PR #4517, opened 2024-02-21
[image] Fix scene test

bakpaul:


applications/plugins/image/examples/ImageTransform.scn
```diff
@@ -1,4 +1,3 @@
-<!-- example of how to give translation parameter to an image and update it every time step-->
```
fredroy, line 1: According to various XML validators, indeed a XML scene should always start with the `<?xml....` stuff.
But I dont think deleting the comment like a brute is a good way to fix 😅

Move the comment below after the  `<?xml....` line and it should work (at least according to https://www.w3schools.com/xml/xml_validator.asp

fredroy (changes requested):

applications/plugins/image/examples/ImageTransform.scn
bakpaul: Sorry, did it to quickly...

fredroy:
[ci-build][with-scene-tests]

fredroy (approved):

PR #4547, opened 2024-02-29
[LinearAlgebra] Refactor sparse matrix product

alxbilger:
1. All types of Eigen::SparseMatrix are now supported. For that, the implementation relies completely on the Eigen matrix product to compute the intersection. The trick is to flag the non-zero values with its index position in the array. The matrix product will implicitly keep track of all the operations thanks to the newly introduced scalar types.
2.  Because of the new method, `SparseMatrixStorageOrder` is no longer useful. It can be removed later
3. Unit tests are extensive. They test all possible configuration of storage of LHS, RHS and Result. They are required because a condition on the storage type is required to swap the indices after the matrix product.
4. A parallel version of the algorithm is introduced
5. Benchmarks: https://github.com/alxbilger/SofaBenchmark/pull/39

Analysis of the results of the benchmarks:
- The best configuration storage for the fast matrix product seems to be ColColCol
- The fast matrix product is always faster than the regular product, except for a matrix of size 1000 and sparsity of 15%. This probably indicates that the method is adapted for very sparse matrices
- The parallel fast matrix product algorithm is always faster than the sequential algorithm, except for very small matrices where the overhead of parallelism is not negligible. However, the speed up is not linear with the number of cores and can be disappointing.
- The parallel fast matrix product is always faster than the regular product even for the matrix of size 1000 and sparsity of 15%, but not for very small matrices (overhead).


alxbilger:
[ci-build][with-all-tests]

alxbilger:
Formatted spreadsheet of the benchmark result
[sparsematrixproductBenchmark.ods](https://github.com/sofa-framework/sofa/files/14458867/sparsematrixproductBenchmark.ods)

Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/SparseMatrixProduct.cpp
```diff
@@ -19,17 +19,8 @@
...
 {
-
-#if !defined(SOFA_LINEARAGEBRA_SPARSEMATRIXPRODUCT_COMPRESSEDROWSPARSEMATRIX_CPP)
-    extern template class SOFA_LINEARALGEBRA_API SparseMatrixProduct<CompressedRowSparseMatrix<float> >;
-    extern template class SOFA_LINEARALGEBRA_API SparseMatrixProduct<CompressedRowSparseMatrix<double> >;
-#endif
-
-} //namespace sofa::linearalgebra
\ No newline at end of file
+}
```
hugtalbot, line 26: @alxbilger is there a sense to keep this empty file?

alxbilger: File removed in https://github.com/sofa-framework/sofa/pull/4599

PR #4576, opened 2024-03-08
[Geomagic] Update GeomagicDriver.cpp to fire error when hd.h is not found

epernod:
quick PR from the pool regarding discussion #4561

### DIFF
--- a/applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
+++ b/applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
@@ -346,7 +346,7 @@ void GeomagicDriver::initDevice()
     updatePosition();
     sofa::core::objectmodel::BaseObject::d_componentState.setValue(sofa::core::objectmodel::ComponentState::Valid);
 #else
-    msg_error() << "GeomagicDriver initialisation failed because OpenHaptics (HD/hd.h) has not been found. Please install openHaptics and specify OpenHaptics_DIR at CMake configure stage.";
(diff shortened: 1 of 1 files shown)


bakpaul (approved):

applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
```diff
@@ -346,6 +346,7 @@ void GeomagicDriver::initDevice()
     updatePosition();
     sofa::core::objectmodel::BaseObject::d_componentState.setValue(sofa::core::objectmodel::ComponentState::Valid);
 #else
+    msg_error() << "GeomagicDriver initialisation failed because OpenHaptics (HD/hd.h) has not been found. Please install openHaptics and specify OpenHaptics_DIR at CMake configure stage.";
```
alxbilger:
```suggestion
    msg_error() << "GeomagicDriver initialization failed because the Geomagic plugin was built without OpenHaptics (HD/hd.h). If you build the plugin yourself, install OpenHaptics and specify OpenHaptics_DIR at CMake configure stage.";
```

As this message is intended for users, I would rephrase like that

alxbilger (changes requested):
Can you remind me why we compile this code even if OpenHaptics is not found?

applications/plugins/Geomagic/src/Geomagic/GeomagicDriver.cpp
epernod: It was to compile the plugin on the CI, where openHaptics can't be installed,  to catch any error due to change in SOFA base of code.
And I wanted to create an emulator AT one moment... Still in m'y todo list...
This was the solution 3 years ago. Maybe WE can install HD now on the CI?

epernod: Ok for the change I let you validate the final version during the meeting.

PR #4580, opened 2024-03-12
[Simulation] Tests: introduce multiple parallel simulations

fredroy:
This PR introduces a way to test if concurrent simulations can run in the same time.
For now, the scenes are in XML (basically copy-paste of caduceus.scn and liver.scn)

The liver test tests merely the possibility to launch multiple scenes.
The caduceus is more complex and test the whole collision pipeline, and constraint-based contacts (friction, LCP)
-> this caduceus does not use MT (in the freeanimationloop) because the TaskScheduler needs a full rewrite to be thread-safe...

The liver one is expected to be working but the caduceus is expected to crash (due to the  intersections and constraints not being thread-safe for the moment)
"expected" because MT usually is not really deterministic so results can vary... 

The best way to make sure it works would be run the TSAN in the same time (not enabled on the CI)

Remark: should the expected test to be failing be disabled by default ? or do we let the test failing ?
(The resolution in future PRs)


alxbilger (approved):

fredroy:
[ci-build][with-all-tests]

fredroy:

PR #4582, opened 2024-03-12
[Collision.Response.Contact] Make contactId thread-safe

fredroy:
~~for thread-safety, and each thread/simu would have its own pool of contact id~~

~~An other solution would be to put a guard_lock/mutex each time the ids are accessed but it would share the same pool of contact id to the whole set of thread/simu~~

Make the counter of ContactId atomic.
A side effect is to remove the ability to restore previous contactid which was destroyed and set in a side container. I dont think we really need to save some "ids" because we will never overrun the 64 bit integer IMO (18446744073709551615 is quite big 🙃)


fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

PR #4583, opened 2024-03-12
[Collision.Detection] Give current intersection methods as parameter to intersection functions

fredroy:
The design of Intersection/Intersector is really "special" in the sense that:
 - Intersection class is the BaseObject in the graph doing stuff
 - Intersector is container of functions for canIntersect/intersect method between different primitives.

And it was made such as one could add more intersect() method for new primitives for an existing "Intersection", e.g MinProximityIntersection and MeshMinProximityIntersection, which add triangles, etc methods to MinProximityIntersection. Note that even if it is called "Intersection" it is not a "Intersection" 🤪
There is kind of mechanism with a IntersectorCreator/IntersectorFactory which does the "plugin" automatically.
The problem is that the intersector gets a pointer to its intersection at the load-time (when the dll is loaded).
The intersector needs the intersection to get alarm/contact distance (defined for all Intersection) or custom parameters (useSurfaceNormal for MinProximityIntersection)

**Consequence:** if there are multiple simulation (in multiple thread)  in the same time, a different thread will get a pointer to an other Intersection while doing its intersect code.

In the end, this PR modifies all the intersect/canIntersect functions to get the intersection method (as if they were static functions in reality) and avoid race conditions. The pipeline needs also to be modified.

[The first version of this branch was using a IntersectionParameters storing alarm, contact distance and given as const ref, but it could not handle custom parameters]

Still need to be done:
 - ~~compat/warnings, etc~~
 - all broad/narrow phases to be modified (further PR)


fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

fredroy:

PR #4586, opened 2024-03-12
[CMake] Upgrade cmake_minimum_required to 3.22

hugtalbot:
In all CMakeLists, upgrade `cmake_minimum_required` from 3.12 to 3.22 in order to support the find or fetch mechanism

**TODO**
- update documentation
- update SP3 readme


bakpaul (approved):

fredroy (changes requested):
According to :
https://cmake.org/cmake/help/latest/module/FetchContent.html#variable:FETCHCONTENT_TRY_FIND_PACKAGE_MODE

The mechanism you are describing has been introduced with 3.24, not 3.22 🤔

bakpaul:
Well, actually the 3.24 version only provide a fully working mechanism.The mechanism exists in 3.22 (I am currently using the 3.22 and it works) the only downside is that the macro `FetchContent_MakeAvailable` doesn't make it available for every target of the project a it does in the 3.24. That is why for instance I had to add this same mechanism into the `SofaMatrix` plugin. But it works fine anyway because it pull it in the same folder and compile it only once and make it available for both of the projects. 

Why 3.22 you might ask ? Well because it is the version shipped in Ubuntu 22.04 LTS, so we will not require a CMake version that is greater than this one...

olivier-roussel:
I also believe the find package integration is not part of cmake 3.22 for the fetch mechanism (so you can use the `FIND_PACKAGE_ARGS` option in `FetchContent_Declare`). Just trying to build the minimal example on this topic on [CMake documentation]( https://cmake.org/cmake/help/latest/module/FetchContent.html#integrating-with-find-package) will not succeed with 3.22 but won't with 3.24. With 3.22, you have no integration of this and you have to call explicitly the `find_package` manually before the fetch (as we do so far in SOFA), but this leads to the problem was citing Paul (dependence not available for all targets). Starting with 3.24, using this option solves the problem (and no need for the `find_package` call before the fetch anymore).
But I understand that we should stick to v3.22 as it is part of ubuntu22.

fredroy (approved):

fredroy:

PR #4648, opened 2024-04-12
[GUI.Qt] Notice user a new version is available

alxbilger:
I suggest to add a message in GUI that a newer version of SOFA is available if an old version is used. I would look like:

image

I chose to rely on Qt because otherwise it requires a new lib for the http request, and Qt is already here. But having Qt is not enough. I have an error related to ssl (similar error to https://stackoverflow.com/questions/53805704/tls-initialization-failed-on-get-request). It requires additional components in the Qt installation. I did not try it myself because it changes the installation process of SOFA. So before going further, we need to conclude that this feature is desirable (and is not a gadget). If it is, we can decide to rely on Qt and its additional component, or to go with a simpler C++ library (https://github.com/JosephP91/curlcpp for example). The second solution would open the version checker to ImGui.


epernod:
perso I vote for curl. I think you have it as soon as you are using ssh/git etc..

bakpaul:
I found a great wrapper around curl : https://github.com/libcpr/cpr

Issue #248, opened 2017-04-20
Stiff-SpringForceField : two components should we merge them?


hugtalbot:
There is two very close implementation : SpringForceField and StiffSprindForceField. Does two implementations really make sense ?

@matthieu-nesme @fredroy @thomas-lemaire

matthieu-nesme:
SpringForceField is an explicit formulation only.
StiffSpringForceField is an implicit formulation.
They both make sense and even if StiffSpringForceField should be used in 99% cases, SpringForceField can be useful in some specific situations.

In my opinion only the names are crap.
I would rename SpringForceField in ExplicitSpringForceField and StiffSpringForceField in SpringForceField, but it would create so much trouble :)

hugtalbot:
Indeed, the naming is therefore pretty inappropriate, we will follow your guidelines ;)
Thank you very much @matthieu-nesme 

Isn't there a way of refactoring (inheritance) these components ? it would make things easier to understand, no?
Just curious, in which case the SpringForceField can be really useful ?

PR #4649, opened 2024-04-12
[Spring] Merging StiffSpringForceField and its parent

bakpaul:
StiffSpringForceField is basically an extension of SpringForcefield adding implicit solving. 
The only added value comes from the ability to define the springs using added data indices1, indices2 and lengths. This mechanism need to also be merged with the current one available in SpringForceField. 

Related to https://github.com/sofa-framework/sofa/issues/248


bakpaul:
Let's see this as a starting point of refactoring. This PR does the merge, the code is neither better nor worst than before, it just removes unnecessary duplication. 

Still need to replace every iteration of StiffSpringForceField by SpringForceField and fix Cuda extension

bakpaul:
[ci-build][with-all-tests]

bakpaul:
PR is ready for me now. 
Still need to decide if I should refactor SofaOpenCL or not given the fact that it doesn't compile with the current SOFA's master branch --> to be discussed in the next SOFA-dev meeting.

bakpaul:
[ci-build][with-all-tests]

Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/StiffSpringForceField.h
```diff
@@ -20,98 +20,12 @@
...
-extern template class SOFA_COMPONENT_SOLIDMECHANICS_SPRING_API StiffSpringForceField<defaulttype::Vec3Types>;
-extern template class SOFA_COMPONENT_SOLIDMECHANICS_SPRING_API StiffSpringForceField<defaulttype::Vec2Types>;
-extern template class SOFA_COMPONENT_SOLIDMECHANICS_SPRING_API StiffSpringForceField<defaulttype::Vec1Types>;
-extern template class SOFA_COMPONENT_SOLIDMECHANICS_SPRING_API StiffSpringForceField<defaulttype::Vec6Types>;
-extern template class SOFA_COMPONENT_SOLIDMECHANICS_SPRING_API StiffSpringForceField<defaulttype::Rigid3Types>;
-#endif
-
-} // namespace sofa::component::solidmechanics::spring
+template<class T>
+using StiffSpringForceField SOFA_ATTRIBUTE_DEPRECATED("v24.06 ", "v25.06", "StiffSpringForceField has been merged with to SpringForceField") = SpringForceField<T>;
```
alxbilger:
```suggestion
using StiffSpringForceField SOFA_ATTRIBUTE_DEPRECATED("v24.06 ", "v25.06", "StiffSpringForceField has been merged with SpringForceField") = SpringForceField<T>;
```

Sofa/GUI/Component/src/sofa/gui/component/performer/AttachBodyPerformer.inl
```diff
@@ -88,21 +89,20 @@ bool AttachBodyPerformer<DataTypes>::startPartial(const BodyPicked& picked)
...
-    using sofa::component::solidmechanics::spring::StiffSpringForceField;
+    using sofa::component::solidmechanics::spring::SpringForceField;
 
-    this->m_interactionObject = sofa::core::objectmodel::New< StiffSpringForceField<DataTypes> >(dynamic_cast<MouseContainer*>(this->m_interactor->getMouseContainer()), mstateCollision);
-    auto* stiffspringforcefield = dynamic_cast< StiffSpringForceField< DataTypes >* >(this->m_interactionObject.get());
-    stiffspringforcefield->setName("Spring-Mouse-Contact");
-    stiffspringforcefield->setArrowSize((float)this->m_size);
-    stiffspringforcefield->setDrawMode(2); //Arrow mode if size > 0
+    this->m_interactionObject = sofa::core::objectmodel::New< SpringForceField<DataTypes> >(dynamic_cast<MouseContainer*>(this->m_interactor->getMouseContainer()), mstateCollision);
+    auto* Springforcefield = dynamic_cast< SpringForceField< DataTypes >* >(this->m_interactionObject.get());
```
alxbilger:
```suggestion
    auto* springforcefield = dynamic_cast< SpringForceField< DataTypes >* >(this->m_interactionObject.get());
```

applications/plugins/SofaCUDA/src/SofaCUDA/component/solidmechanics/spring/CudaSpringForceField.cpp
```diff
@@ -75,12 +71,12 @@ namespace sofa::gpu::cuda
 //#endif // SOFA_GPU_CUDA_DOUBLE
 //;
 
-int StiffSpringForceFieldCudaClass = sofa::core::RegisterObject("Supports GPU-side computations using CUDA")
-        .add< sofa::component::solidmechanics::spring::StiffSpringForceField<CudaVec3fTypes> >()
-        .add< sofa::component::solidmechanics::spring::StiffSpringForceField<CudaVec3f1Types> >()
+int SpringForceFieldCudaClass = sofa::core::RegisterObject("Supports GPU-side computations using CUDA")
```
alxbilger, line 65: Can you remove the commented code just above?

bakpaul:
[ci-build][with-all-tests]

bakpaul:
[ci-build][with-all-tests]

bakpaul:
[ci-build][with-all-tests]

bakpaul:
As I said this pr is a first step towards a better SpringForceField API. 

What is done in this PR is a **conservative merge** of the StiffSpringForcefield and the SpringForceField component because both where basically the same except for the fact that SpringForcefield was "weaker" because it was supposed to be use in an explicit solver, and thus the forces only appeared in the LHS. 

More differences lied it the fact that the Stiff version had two datas (indices1 and 2) that where indices lists linked to no topology and only there to be able to define springs in scenes by two vectors on indices. Although in the SpringFF there already was two lists of indices linked to topology and coping for topological changes. There was no reason those two lists existed together. 

Moreover, the topological change callback attached to the two `topologySubestIndices` lists was not working as expected. Springs were not removed correctly due to the fact that the `topologySubestIndices` data was not supposed to work with multiple occurrence of the same index. This was corrected in the following PR : https://github.com/sofa-framework/sofa/pull/4738 along with some modifications in the present one (for the callback on the forcefield side). 

One of the major difficulties lies in the fact that the two components presented two representations of the spring list: one gather more information than the other, and the second one being less practical to work with especially in scene building.
- In SpringFF every spring is listed in a data list of LinearSpring. Along with this the two topologySubestIndices where here to cope for topological changes
- In StiffSpringFF, along with the inherited data from SpringFF, lied the indices lists and a single data for stiffness and damping. 

In order to be as conservative as possible, I tried in this PR to only do the merge while keeping every current features enabled. This leads in a double representation of the spring list: the list `d_springs` and the tuple `d_springsIndices[0], d_springsIndices[1], d_ks, d_kd, d_lengths, d_elongationOnly, d_enabled` that needs to be kept coherent constantly through call back. This is cumbersome because we cannot have a circular dependency, and has thus been solved by implicitly defining a hierarchy between the two representation : the tuple representation is the master one, and the spring list follows it. This would be better if we could have a read only data so that the user cannot access the spring list. 

The spring list is still useful for three things :
1. the CUDA implementation (but not mandatory, removing it would require a significant refactoring)
2. the huge number of existing example scenes using SpringForceField that uses this representation in the scene
3. the extensive use of this representation in the python bindings (especially the linearSpring that has no other purpose that this API). Other than that, it has no real superiority over the proposed parallel representation using a single vector for each spring parameter except for the fact that we don't need to keep coherency over all of the vectors...  
Although the tuple enable more versatile scene description. For example, the input indices can now come from other component while no difficult assembling is needed to build the spring list, we only need to link the data and specify one stiffness that will be generalized to all the springs. 

So this is a **_first step_**. It can totally be merged for me. But we still have work to do. It could be done in this PR but I don't know how much time I'll be able to dedicate to it in the near future, and I don't want this work to get lost. Those are the two remaining tasks I've identified :
- [ ] Clean the API by removing one of the two representation. I would prefer to remove the `d_springs` data which would require a refactoring of the CUDA code and of the scenes.
- [ ] Clean classes that originally inherited from the StiffSpringFF (QuadBendingSprings, RegularGridSpringFF & TriangularBendingSprings) because until then, they were only considering a constant value of stiffness and damping for all their springs, turning a blind eye on the d_spring data. Now that we don't have a single attribute for stiffness and damping (as it was the case in the stiffSpringFF) but a vector of them, they should also be cleaned, along with their associated scenes.

If those two points have not been addressed in this PR, it is because I think it is better to have the opinion of the @sofa-framework/reviewers on whether we only want to keep one representation and if so which one ?

PS: I've added a scene test for the CUDA spring implementation because it was never tested, I don't know if it was really working until here

PR #4714, opened 2024-05-02
[SofaCUDA] extern template instantiations

alxbilger:
The extern template instantiations allows to include ` SofaCUDA/config.h` before `ObjectFactor.h`. That is why the module name is defined properly in this PR. The extern template instantiations don't act directly on the module name, just the order of includes.


alxbilger:
[ci-build][with-all-tests]

epernod:
Sorry I don't understand the link between your comment on the factory and the changes in the PR.

alxbilger:
The function `getTarget` returns an empty string for the 3 components in this PR. The changes in this PR makes it return a valid string

epernod:
ok so if I understand well, adding the template instantiations only in the cpp return empty string by `getTarget`
Stange... from what I remembered there are several collision classes that are instantiate the same way in the plugin....

bakpaul (approved):

hugtalbot (approved):

PR #4719, opened 2024-05-06
[image_gui] Make it a cleaner SOFA module

alxbilger:
- Fixed the module name
- config.h is now config.h.in

A lot more work to do if we want to clean this plugin


applications/plugins/image/image_gui/CMakeLists.txt
```diff
@@ -42,9 +42,10 @@ add_definitions("-DSOFA_BUILD_IMAGE_GUI")
...
 target_link_libraries(${PROJECT_NAME} image ${OPENGL_LIBRARIES} Sofa.GUI.Qt)
 
-sofa_add_targets_to_package(
-    PACKAGE_NAME image
+sofa_create_package_with_targets(
+    PACKAGE_NAME ${PROJECT_NAME}
+    PACKAGE_VERSION ${Sofa_VERSION}
     TARGETS ${PROJECT_NAME} AUTO_SET_TARGET_PROPERTIES
-    INCLUDE_SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}"
+    INCLUDE_SOURCE_DIR "."
```
bakpaul: I don't understand why this works, it should behave the same as the with the parameter before the modification

alxbilger: the macro is not the same

bakpaul: Yes but at the end what fixes the include dir is the same in `sofa_auto_set_target_include_directories` : 
```cmake
if(ARG_INCLUDE_SOURCE_DIR)
    if(IS_ABSOLUTE "${ARG_INCLUDE_SOURCE_DIR}")
        set(include_source_root "${ARG_INCLUDE_SOURCE_DIR}")
    else()
        set(include_source_root "${CMAKE_CURRENT_SOURCE_DIR}/${ARG_INCLUDE_SOURCE_DIR}")
    endif()
endif()
```
And regarding your input, the final include is indeed `${CMAKE_CURRENT_SOURCE_DIR}` so it seems that you are not using your include dir when including header with prefix `image_gui`...

alxbilger:
[ci-build][with-all-tests]

fredroy (approved):

PR #4722, opened 2024-05-06
[Core,Helper] Prevent to create objects from unloaded plugins

alxbilger:
When a plugin is loaded, the ObjectFactory is populated of Components from this plugin. However, when this plugin is "unloaded", it was still possible to create a component from the plugin. This PR detects that the ObjectFactory tries to instantiate an object that is from an unloaded plugin.
It is particularly useful when the initialization of the plugin fails (SofaCUDA for example). When the initialization fails, the plugin is unloaded.
A unit test is provided

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp
+++ b/Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp
@@ -182,7 +182,7 @@ objectmodel::BaseObject::SPtr ObjectFactory::createObject(objectmodel::BaseConte
             Creator::SPtr c = it2->second;
             const auto& unloadedPlugins = helper::system::PluginManager::getInstance().unloadedPlugins();
             const auto creatorTarget = c->getTarget();
-            if (std::find(unloadedPlugins.begin(), unloadedPlugins.end(), creatorTarget) != unloadedPlugins.end())
+            if (unloadedPlugins.find(creatorTarget) != unloadedPlugins.end())
             {
                 creators_errors[templatename].emplace_back(
                     "The object was previously registered, but its module has been unloaded.");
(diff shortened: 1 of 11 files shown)


Sofa/framework/Helper/src/sofa/helper/system/PluginManager.h
```diff
@@ -229,6 +231,8 @@ class SOFA_HELPER_API PluginManager
 
     PluginMap m_pluginMap;
     std::map<std::string, std::function<void(const std::string&, const Plugin&)>> m_onPluginLoadedCallbacks;
+
+    sofa::type::vector<std::string> m_unloadedPlugins;
```
bakpaul: Why a vector and not a unordered_set since this name should be unique ? All the modifications could be done through the key directly

bakpaul:
To be discussed at the STC.

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.h
alxbilger: you're right

alxbilger:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp
```diff
@@ -197,12 +208,23 @@ objectmodel::BaseObject::SPtr ObjectFactory::createObject(objectmodel::BaseConte
...
                 }
+                else
+                {
+                    if (c->canCreate(context, arg)){
+                        creators.push_back(*it3);
+                    } else {
+                        creators_errors[it3->first] = arg->getErrors();
+                        arg->clearErrors();
+                    }
+                }
```
hugtalbot, line 222: since both pieces of codes look very similar, is there no way to factorize it?

alxbilger: yes probably

alxbilger: it's done

fredroy:
[ci-build][with-all-tests]

fredroy:
Tests are fixed, ready then 👮

PR #4732, opened 2024-05-13
[tests] Fix and test value type string of topology primitives

alxbilger:
Note that `Point` does not pass the test

The introduced tests don't pass on non-Windows OS

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/topology/Topology.h
+++ b/Sofa/framework/Core/src/sofa/core/topology/Topology.h
@@ -131,3 +131,25 @@ SOFA_CORE_API extern const unsigned int quadsOrientationInHexahedronArray[6][4];
 SOFA_CORE_API extern const unsigned int verticesInHexahedronArray[2][2][2];
 
 } // namespace sofa::core::topology
+
+namespace sofa::geometry
+{
+
+// Specialization required because Topology::Point is not an alias of
+// sofa::topology::Element, compared to the other aliases such as
+// Edge, Triangle...
+template<>
+struct ElementInfo<sofa::core::topology::Topology::Point>
+{
+    static geometry::ElementType type()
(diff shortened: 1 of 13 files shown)


alxbilger:
`Element<Point>` and `vector<Element<Point>,CPUMemoryManager<Element<Point>>>`

alxbilger:
Maybe `Element<Point>` does not make sense...

fredroy:
I tried and on macOS and Ubuntu/gcc :
 - `EXPECT_EQ(std::string{sofa::geometry::ElementInfo<sofa::geometry::Point>::name()}, "Point");` test passes
 - `EXPECT_EQ(defaulttype::DataTypeName<sofa::topology::Element<sofa::geometry::Point> >::name(), "Point");` does not (return Element<Point> as you said)

After some investigation, `DataTypeInfo<topo::Element<geo::Point>` does not exist so it uses the default decoder.
IMO, It was not implemented because topology::Point (which should be an alias on `topo::Element<geo::Point>`) is not defined. I dont remember why it was not done while Sofa.Topo and Sofa.Geo was created. Maybe topology::Point does not make sense ? (in a topology point of view)
If in the end it does, making the alias (and adding the typeinfo) does work. (see https://github.com/alxbilger/sofa/pull/8 )

alxbilger:
@lamriaimen could you try that with this branch you don't have any widget registration issue with SofaImGui ? Thanks

alxbilger:
[ci-test][with-all-tests]

lamriaimen:
@alxbilger no errors were generated when i tried to run it with imgu.
the problem of "[DataWidgetFactory] Cannot add widget vector<> into the factory" is solved i guess.

alxbilger:
[ci-build][with-all-tests]

PR #4738, opened 2024-05-16
[Core] Add topology subset indices test

bakpaul:
During my Way of the Cross of fusing StiffSpring and its parent, I saw that the topological change of removing points didn't work as planned for topologySubsetIndices when there is multiple occurrence of the same element  in the data. 

I've fixed that and added tests. 

One question is remaining though : here I kept the original mechanism using a swap of the deleted element and the last one. This is efficient in term of memory but it has the side effect of changing the indices order in the data. 
--> My question is, *is that what we want ?* Do we prefer memory/time efficiency over order coherency for this data ? Is it logical to get a random order of the vector out of a simple topological change ?  This answer will change a bit the way I'll finish the refactoring in https://github.com/sofa-framework/sofa/pull/4649 

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/topology/TopologySubsetData.h
+++ b/Sofa/framework/Core/src/sofa/core/topology/TopologySubsetData.h
@@ -137,7 +137,7 @@ protected:
     * Internal method called at the end of @sa remove method to apply internal mechanism, such as updating the map size
     * @param nbElements Number of element removed.
     */
-    virtual void removePostProcess(sofa::Size nbElements);
+    virtual void removePostProcess(sofa::Index elemId);
 
     /**
     * Internal method called at the end of @sa add method to apply internal mechanism, such as updating the map size.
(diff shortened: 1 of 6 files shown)


bakpaul:
[ci-build][with-all-tests]

epernod:
as this component is in fact a kind of map. The swap might not be needed... but without diving back into this I'm not 100% sure...
normally there is enough regression tests on topology modification to detect if removing the swap will break the mechanism.

epernod:
btw why have you redundant values? This is not logical for an indice map.

bakpaul:
Well it isn't actually a map in the code but a vector. This TopologySubsetIndices data is used in SpringForcefield to keep track of the topological changes and contains the indices on which each spring is attached. Given the fact that multiple springs can be attached to one point, then this vector may contains multiple occurrence of the same index (and it is the case for some unit tests currently). And more importantly its order is very important because it should be kept consistent with the order of the list of attached indices of the second object...

bakpaul:
I though that maybe this feature of supporting multiple occurrence was meaningfull for all TopologySubsetData, but it might be a special case for topologySubsetIndices, su I could overlead the method in its declaration if you prefer ?

bakpaul:
[ci-build][with-all-tests]

bakpaul:

PR #4754, opened 2024-05-23
[Helper] Fix dl open when path are not canonical

bakpaul:
Error where displayed when using generated binaries on another computer. I'm not sure why this didn't happend when it was launched on the computer that built the binaries... 

The path comparison didn't work when they weren't  in a canonical form. This method only appeared in c++17, but if I am right we only support the c++ version of the two last LTS, 22.04 distributing gcc11 which provides a full support of c++17. 

### DIFF
--- a/Sofa/framework/Helper/src/sofa/helper/system/DynamicLibrary.cpp
+++ b/Sofa/framework/Helper/src/sofa/helper/system/DynamicLibrary.cpp
@@ -126,7 +126,7 @@ void * DynamicLibrary::getSymbolAddress(Handle handle,
                     path = path.parent_path() / symlinkHandlePath;
                 }
             }
-            return absolute(path);
+            return std::filesystem::absolute(path);
         };
 
         Dl_info dli;
--- a/Sofa/framework/Helper/src/sofa/helper/system/DynamicLibrary.cpp
+++ b/Sofa/framework/Helper/src/sofa/helper/system/DynamicLibrary.cpp
@@ -128,7 +128,7 @@ void * DynamicLibrary::getSymbolAddress(Handle handle,
         }
         catch(std::filesystem::filesystem_error& error)
         {
-            oss << "An exception was catch when trying to find real path of the library with the following message " << error.what();
+            oss << "An exception was caught when trying to find the real path of the library with the following message : " << error.what();
 
             // paths are broken, we cannot check if this symbol comes from this lib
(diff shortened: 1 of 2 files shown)


alxbilger (approved):

Sofa/framework/Helper/src/sofa/helper/system/DynamicLibrary.cpp
```diff
@@ -126,7 +126,7 @@ void * DynamicLibrary::getSymbolAddress(Handle handle,
                     path = path.parent_path() / symlinkHandlePath;
                 }
             }
-            return path;
+            return std::filesystem::canonical(path);
```
fredroy: I was not aware of this "canonical" function when I did the getRealPath() 😅
IMO just use  `std::filesystem::canonical()` on the path directly (instead of the lambda)

And canonical() can throw so it would be wise to try/catch

fredroy (changes requested):

Sofa/framework/Helper/src/sofa/helper/system/DynamicLibrary.cpp
bakpaul: Thank you, done !

Sofa/framework/Helper/src/sofa/helper/system/DynamicLibrary.cpp
```diff
@@ -114,31 +114,33 @@ void * DynamicLibrary::getSymbolAddress(Handle handle,
...
+        std::ostringstream oss;
+
+        try
+        {
+            dlInfoPath = std::filesystem::canonical(dli.dli_fname);
+            handlePath = std::filesystem::canonical(handle.filename());
+        }
+        catch(std::filesystem::filesystem_error& error)
+        {
+            oss << "An exception was catch when trying to find real path of the library with the following message " << error.what();
```
fredroy:
```suggestion
            oss << "An exception was caught when trying to find the real path of the library with the following message : " << error.what();
```

fredroy: 🇬🇧💂

fredroy (changes requested):

fredroy:
[ci-build][with-all-tests]

fredroy (approved):

PR #4763, opened 2024-06-04
[Lagrangian.Model] Use base class Data instead of protected class member for constraint id

alxbilger:
related to https://github.com/SofaDefrost/SoftRobots/pull/283

### DIFF
--- a/Sofa/Component/Constraint/Lagrangian/Model/src/sofa/component/constraint/lagrangian/model/SlidingLagrangianConstraint.h
+++ b/Sofa/Component/Constraint/Lagrangian/Model/src/sofa/component/constraint/lagrangian/model/SlidingLagrangianConstraint.h
@@ -60,7 +60,6 @@ protected:
 
     Real m_dist;	// constraint violation
     Real m_thirdConstraint; // 0 if A<proj<B, -1 if proj<A, 1 if B<proj
-    unsigned int m_cid;
 
 
     SlidingLagrangianConstraint();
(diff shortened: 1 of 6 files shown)


EulalieCoevoet (approved):

alxbilger:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/behavior/Constraint.inl
```diff
@@ -69,6 +69,7 @@ void Constraint<DataTypes>::buildConstraintMatrix(const ConstraintParams* cParam
 {
     if (cParams)
     {
+        m_constraintIndex.setValue(cIndex);
```
alxbilger: @EulalieCoevoet what do you think about setting the value here, at a higher level? All derived classes would then set the value.

EulalieCoevoet: Yes this is a good idea. I'll do the same in SoftRobots (SoftRobotsConstraint).

**Edit**: Is it okay to hide the information that this value is only available after the call to buildConstraintMatrix()? The answer can be yes, I'm just wondering.

alxbilger: you're right, setting the id could be done outside of the function `buildConstraintMatrix`. That is why I suggest https://github.com/sofa-framework/sofa/pull/4786.

This way, the id can be set from the outside using the `setConstraintId` function, that you can find here:

https://github.com/sofa-framework/sofa/blob/05a148594b72b495457fb2f1644edaa1c155925f/Sofa/framework/Simulation/Core/src/sofa/simulation/mechanicalvisitor/MechanicalBuildConstraintMatrix.cpp#L33

I suggest to pause this PR, the time that https://github.com/sofa-framework/sofa/pull/4786 is merged. Can you review it?

bakpaul (approved):

PR #4767, opened 2024-06-05
[tools] Add metis relocation in post install fixup

bakpaul:
Metis is installed only in the folder of the first plugin fetching it. Making the other one fail to find it. Moving it into the main folder with CMake is not really meaningful. The plugin should not fix something that is at higher level in the CMake hierarchy as itself (aka, fix install issue when another target depends on one of its dependency)...  


bakpaul:
Currently being tested : https://ci.inria.fr/sofa-ci-dev/job/sofa-custom/8615/

PR #4798, opened 2024-06-27
[Mapping.NonLinear] Effectively use child force id

alxbilger:
Child force id is given as a parameter of the function `applyDJT`, and instead of using it, the force vector id from `MechanicalParams` is used. I suggest to use the given parameter in this PR. Both force ids should be equal in most cases, so I don't expect breaking changes of behavior. I just find it weird to give a parameter and not using it in favor of a hard-coded id.


fredroy:
[ci-build][with-all-tests]

alxbilger:
In the end, both vector ids are not the same. The one given in parameter is not necessarily the vector force in the state.

PR #4813, opened 2024-07-05
[All] Remove SOFA_ENABLE_LEGACY_HEADERS usage

fredroy:
A continuation of 
- #4812 

but much bigger and less trivial.

This PR removes SOFA_ENABLE_LEGACY_HEADERS, implying:
- no more collections/deprecated projects which was to ensure the continuity of Sofa.NG
- some renaming in MultiThreading and SofaCUDA.

The deletion of the layer compat between old and new modules was acted for 23.06 so not a big deal by itself (only 1.5 year late🤷‍♂️ ).
But some olden modules still have some code/components:
- SofaGraphComponent: with a `Gravity` component
- SofaMiscCollision: with `DefaultCollisionGroupManager, SolverMerger` components and `RayTriangleVisitor` visitor
- SleepController: with a `SleepController`component
- SofaValidation: with various measuring stuff components.
There is also SofaExporter which had some forgotten(?) examples of components which were moved. So the examples have been just moved as well

So what should be done for these components?
IMO (not done yet)
 - `Gravity` seems useless/not usable.
 - `SleepController` might maybe probably be interesting
 - `DefaultCollisionGroupManager, SolverMerger` were bogus and RayTriangleVisitor not used at all
 - SofaValidation may be totally transformed as a plugin (as in `applications/plugins`)

To be discussed.


fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

bakpaul:
So I took a look at the SleepController. It simply puts context to sleep where the mechanical object has a max velocity under a certain threshold. Then wake them up again only when a collision occurs with another object that is moving. 
Three remarks : 
- Being put to sleep disables any mechanical visitor to be applied on the node and deactivates its constraint corrections. So no more computing (no system build, no integration, only collision detection)
- This works well only for scenes where external interaction only arise between two objects colliding, neither by the mean of a change of external forces (dynamic vector field for instance), or of constraint state changes not managed by the collision pipeline (for instance statically defined interaction constraints such as cable constraints). 
- The code seems overly complicated for what it performs but certainly does what it says it does. I didn't try it but it looks ok. 

Sincerely, I cannot see myself advising anyone to use it  in their scene given the fact that the use case doesn't apply for the majority of the scenes (especially for soft robotics). So I wouldn't mind never seeing it again...

fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):

fredroy:
> Reviews seem to have been taken into account, right @fredroy ? CI says yes!

Yes

PR #4828, opened 2024-07-12
[Helper] Added cleanup callback

bakpaul:
This pr aims at fixing the crash on exit for some python scene. It was due to the cache of the Sofa python module not being cleaned up correctly (deleted after the interpreter). This is corrected here by insuring that the interpreter is correcty released through cleanup callback (see the SofaPython3 pr for more insights)  


bakpaul:
[ci-build][with-all-tests]

applications/projects/runSofa/Main.cpp
```diff
@@ -544,5 +544,6 @@ int main(int argc, char** argv)
 
     sofa::simulation::common::cleanup();
     sofa::simulation::graph::cleanup();
+    PluginManager::getInstance().cleanup();
```
fredroy: I think it would be better to set this line directly into the cleanup of sofa::helper()
ie. here: https://github.com/sofa-framework/sofa/blob/9fda8923672664c1e98060400ed066c4c7735d1c/Sofa/framework/Helper/src/sofa/helper/init.cpp#L49

which is called "cascadingly" by the other cleanups()

fredroy (changes requested):

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
```diff
@@ -255,11 +266,25 @@ void PluginManager::addOnPluginLoadedCallback(const std::string& key, std::funct
     }
 }
 
+void PluginManager::addOnPluginCleanupCallbacks(const std::string& key, std::function<void()> callback)
+{
+    if(m_onPluginCleanupCallbacks.find(key) == m_onPluginCleanupCallbacks.end())
+    {
+        m_onPluginCleanupCallbacks[key] = callback;
+    }
```
alxbilger: replace with std::map::insert as it requires only one instruction

applications/projects/runSofa/Main.cpp
bakpaul: ok, you are right, better than polluting the main...

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
bakpaul: Ok, I've also cleaned the one I've copied from.

fredroy (approved):

Sofa/framework/Helper/src/sofa/helper/init.cpp
```diff
@@ -50,6 +51,8 @@ SOFA_HELPER_API void cleanup()
 {
     if (!s_cleanedUp)
     {
+
+        system::PluginManager::getInstance().cleanup();
```
alxbilger, line 55: I wonder if Sofa.Helper is the right place to call a plugin clean up...

bakpaul: It was asked by @fredroy and it makes sens because the plugin manager belongs to Sofa.Helper

alxbilger: Yes, but your goal is to clean up plugins, not the plugin manager. For example, you will clean up SofaPython3 inside the Sofa.Helper cleanup function. I am not saying it is not the right thing to do, I am just questioning ...

bakpaul: My first though was to put it at the very end of the main function, both are justifiable. But I guess that conceptually, the cleanup order shouldn't matter, the simulation is over, we are just cleaning after us...

fredroy: Why not inside the destructor of PluginManager ? (if it is ever called 🥴)

bakpaul: We don't want it to be in a static instance, this fix aims at cleaning the static cache before the dynamically allocated python interpreter... So no, it is necessary to have it explicitly called.

bakpaul:
@damienmarchal , discussions during the SOFA-dev meeting lead to a question over the developed design and your opinion is needed. 

The problem was that some scenes tests where crashing at exit. This is due to the SofaPython3 cache (`s_objectcache` there https://github.com/sofa-framework/SofaPython3/blob/1b5d271aa21df06c4c5ab75599e364adb0c1e9be/Plugin/src/SofaPython3/DataHelper.cpp#L252C1-L252C73) . The problem was that this is a static object containing python objects and is thus cleaned at the very end of the program exit, after the python interpreter, which is not authorized.  

To fix it, following [pybind11 suggestions](https://pybind11.readthedocs.io/en/stable/advanced/misc.html) I've added a member to the module Sofa.Core that insured that the cleanup is called while the interpreter is still alive (see the PR https://github.com/sofa-framework/SofaPython3/pull/425). But is was not enough because the interpreter is not cleaned correctly either : the static method "Release" of SofaPython3, in charge of cleaning up python and deleting the interpreter was never called when launching scene in runSofa. 

This lead to the need of cleaning up *explicitly* the plugin before exiting the main function. This design was not liked by every dev, some wanting a Raii-like cleanup, in the destructor of some member, which is in my mind not possible because the interpreter instance is static in the SofaPython3 plugin and thus we have no guaranty that it  would not be deleted before any other static member.

bakpaul:
[ci-build][with-all-tests]

bakpaul:
[ci-build][with-all-tests]

bakpaul:

PR #4848, opened 2024-07-25
[ODESOlver] Fix RHS computation of the backward Euler implicit solver

bakpaul:
Remove damping correction to the RHS because it is already added in the computeForce.

In a Euler implicit integration scheme, the right hand side expression only contains terms coming for the internal and external forces taken at the begining of the timestep, and an inertial term constituted of the stiffness matrix, but no term coming from the constitutive damping.  (see [Christian Duriez HDR](https://theses.hal.science/tel-00785118) for ref)

The contribution of the damping forcefield is already taken into account through the `computeForce` call line 126, so the `b` coefficient should be put to 0 here otherwise the behavior is not correct. 

This can bee reproduced by launching the scene `examples/Component/SolidMechanics/FEM/DampingForceField.scn` and putting the data `implicit` of the damping forcefield to true. This data might have been introduced to fix this error, because here we are in an implicit scheme, and setting this parameter to true when the `b` coef is equal to 1 changes the behavior. 

This was introduced here : https://github.com/sofa-framework/sofa/commit/fe0ee6248971383daa4e488e80f5fe480ee85988

[ci-depends-on https://github.com/sofa-framework/Regression/pull/67]

### DIFF
--- a/Sofa/Component/ODESolver/Backward/src/sofa/component/odesolver/backward/EulerImplicitSolver.cpp
+++ b/Sofa/Component/ODESolver/Backward/src/sofa/component/odesolver/backward/EulerImplicitSolver.cpp
@@ -123,7 +123,7 @@ void EulerImplicitSolver::solve(const core::ExecParams* params, SReal dt, sofa::
         SCOPED_TIMER("ComputeForce");
         mop->setImplicit(true); // this solver is implicit
         // compute the net forces at the beginning of the time step
-        mop.computeForce(f);
+        mop.computeForce(f);                                                               //f = Kx + Bv
 
         msg_info() << "EulerImplicitSolver, initial f = " << f;
     }
@@ -138,15 +138,15 @@ void EulerImplicitSolver::solve(const core::ExecParams* params, SReal dt, sofa::
         // new more powerful visitors
 
         // force in the current configuration
-        b.eq(f,1.0/tr);                                                                         // b = f0
+        b.eq(f,1.0/tr);                                                                    // b = f
 
         msg_info() << "EulerImplicitSolver, f = " << f;
 
         // add the change of force due to stiffness + Rayleigh damping
-        mop.addMBKv(b, -d_rayleighMass.getValue(), 1, h + d_rayleighStiffness.getValue()); // b =  f0 + ( rm M + B + (h+rs) K ) v
+        mop.addMBKv(b, -d_rayleighMass.getValue(), 0, h + d_rayleighStiffness.getValue()); // b =  f + ( rm M + (h+rs) K ) v
 
         // integration over a time step
-        b.teq(h*tr);                                                                        // b = h(f0 + ( rm M + B + (h+rs) K ) v )
+        b.teq(h*tr);                                                                       // b = h(f + ( rm M + (h+rs) K ) v )
     }
 
     msg_info() << "EulerImplicitSolver, b = " << b;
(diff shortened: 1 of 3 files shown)


Sofa/Component/ODESolver/Backward/src/sofa/component/odesolver/backward/EulerImplicitSolver.cpp
```diff
@@ -138,15 +138,15 @@ void EulerImplicitSolver::solve(const core::ExecParams* params, SReal dt, sofa::
...
 
         // force in the current configuration
-        b.eq(f,1.0/tr);                                                                         // b = f0
+        b.eq(f,1.0/tr);                                                                    // b = f
 
         msg_info() << "EulerImplicitSolver, f = " << f;
 
         // add the change of force due to stiffness + Rayleigh damping
-        mop.addMBKv(b, -d_rayleighMass.getValue(), 1, h + d_rayleighStiffness.getValue()); // b =  f0 + ( rm M + B + (h+rs) K ) v
+        mop.addMBKv(b, -d_rayleighMass.getValue(), 0, h + d_rayleighStiffness.getValue()); // b =  f + ( rm M + (h+rs) K ) v
```
bakpaul, line 146: B the way I don't understand why the minus sign on the mass matrix. This is also strange. My understanding is that the sign should be the same for both Rayleigh factors.

fredroy:
[ci-build][with-all-tests]

Sofa/Component/ODESolver/Backward/src/sofa/component/odesolver/backward/EulerImplicitSolver.cpp
hugtalbot: I agree, I had the same confusion, writing the [associated doc](https://www.sofa-framework.org/community/doc/simulation-principles/system-resolution/integration-scheme/#rayleigh-damping) (which is just following what is in the code)

Insight of @sofa-framework/reviewers would be needed as well

jnbrunet: @bakpaul @hugtalbot seems to be SOFA's convention for mass and force field. I've documented it with the good signs in Caribou: https://github.com/mimesis-inria/caribou/blob/master/src/SofaCaribou/Ode/BackwardEulerODESolver.cpp#L107

hugtalbot: Thanks a lot for this feedback @jnbrunet 
Do you know what justifies this -K convention ? instead of defining with a negative sign of the RHS we let a positive sign while defining K as negative?

jnbrunet: @hugtalbot no idea, but it doesn't change anything. However, if you want to invert it, you will have to update every forcefields of SOFA so that they remove the minus factor. This is not always as straightforward [as in caribou](https://github.com/mimesis-inria/caribou/blob/master/src/SofaCaribou/Forcefield/HyperelasticForcefield.inl#L166), sometime the minux factor is hidden into the displacement computation (see tetrahedronFemForcefield for an example if I remember correctly).

hugtalbot: Yes of course, thanks a lot for confirming this @jnbrunet :bulb: 
Point about Rayleigh coefficient solved :heavy_check_mark:

hugtalbot (approved):

Sofa/Component/ODESolver/Backward/tests/EulerImplicitSolver_withDamping_test.cpp
```diff
@@ -0,0 +1,129 @@
...
+        sofa::simulation::node::initRoot(root.get());
+        //*********
+        // run simulation
+
+        // compute 2500 time steps
+        for(int i=0; i<2500; i++)
+            sofa::simulation::node::animate(root.get(), 0.02);
+
+        // access the MechanicalObect (access position of the dampedParticule)
+        typename MechanicalObject<sofa::defaulttype::Vec3dTypes>::SPtr dofs = dampedParticule->get<MechanicalObject<sofa::defaulttype::Vec3dTypes>>(root->SearchDown);
```
fredroy, line 107:
```suggestion
        auto dofs = dampedParticule->get<MechanicalObject<sofa::defaulttype::Vec3dTypes>>(root->SearchDown);
```

fredroy:

PR #4919, opened 2024-08-13
[Playback] Fix WriteStateCreator to pass period only if it has a value. Because WriteState use isSet value...

epernod:
Quick fix: to allow Regression to dump only last step as before. Right now it will always dump all dofs at each timestep.

WriteState init method check if d_period  is set. 
Therfor If d_period  is set to 0.0 it will activate a section of the code what will overwrite the period to the current timestep. 

### DIFF
--- a/Sofa/Component/Playback/src/sofa/component/playback/WriteState.cpp
+++ b/Sofa/Component/Playback/src/sofa/component/playback/WriteState.cpp
@@ -109,7 +109,8 @@ void WriteStateCreator::addWriteState(sofa::core::behavior::BaseMechanicalState 
         if (!m_times.empty())
             ws->d_time.setValue(m_times);
 
-        if (m_period > 0.0) {
+        if (m_period > 0.0) 
+        {
             ws->d_period.setValue(m_period);
         }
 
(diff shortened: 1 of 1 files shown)


Sofa/Component/Playback/src/sofa/component/playback/WriteState.cpp
```diff
@@ -109,7 +109,9 @@ void WriteStateCreator::addWriteState(sofa::core::behavior::BaseMechanicalState
         if (!m_times.empty())
             ws->d_time.setValue(m_times);
 
-        ws->d_period.setValue(m_period);
+        if (m_period > 0.0) {
+            ws->d_period.setValue(m_period);
+        }
```
fredroy:
```suggestion
        if (m_period > 0.0) 
        {
            ws->d_period.setValue(m_period);
        }
```

fredroy (approved):

fredroy:
[ci-build][with-all-tests]

epernod:
> Looks reasonable to me, but I am almost never using WriteState. If you are confident in your change @epernod let's go

yes this is a bug I introduced in PR https://github.com/sofa-framework/sofa/pull/4744

PR #4921, opened 2024-08-15
[all] Spread API isComponentValid

hugtalbot:
Make sure the API `isComponentValid` is used instead of direct access to the data `d_componentState`


Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/TriangularBendingSprings.inl
```diff
@@ -328,7 +328,7 @@ void TriangularBendingSprings<DataTypes>::init()
         return;
     }
 
-    if (m_topology->d_componentState.getValue() == sofa::core::objectmodel::ComponentState::Invalid || m_topology->getNbTriangles()==0)
+    if ( !m_topology->isComponentStateValid() || m_topology->getNbTriangles()==0 )
```
hugtalbot, line 331: This change makes the following tests fail:
-  TriangularBendingSprings3_test.checkForceField_Creation
-  TriangularBendingSprings3_test.checkForceField_wrongAttributess
-  TriangularBendingSprings3_test.checkForceField_init
-  TriangularBendingSprings3_test.checkForceField_values
-  TriangularBendingSprings3_test.checkForceField_TopologyChanges

Sofa/Component/Engine/Select/src/sofa/component/engine/select/BaseROI.inl
```diff
@@ -278,11 +278,8 @@ bool BaseROI<DataTypes>::isPointIn(const PointID pid) const
 template <class DataTypes>
 void BaseROI<DataTypes>::doUpdate()
 {
-    if(d_componentState.getValue() == ComponentState::Invalid)
-    {
-        return ;
-    }
-
+    if(!this->isComponentStateValid())
+        return;
```
hugtalbot, line 282: Same here for all tests about ROI

Sofa/Component/SolidMechanics/Spring/src/sofa/component/solidmechanics/spring/TriangularBendingSprings.inl
epernod: Be careful that checking:  `!isValid` and `== Invalid` is not the same. For example in Topology or Loader we use intermediate state at init like `Loading` or `init`
In this case we are sure the component has been set to invalid meaning init failed and we trigger an error.

epernod (changes requested):
Be careful that checking:  `!isValid` and `== Invalid` is not the same. For example in Topology or Loader we use intermediate state at init like `Loading` or `init`
In this case we are sure the component has been set to invalid meaning init failed and we trigger an error.

Regarding this->d_componentState  or  using directly d_componentState. You need to use `this` in templated class otherwise you need to set:
`using Inherit::d_componentState ;` or the compilation will failed.

PR #4875, opened 2024-08-01
[Helper] Retrieve the current user local configuration directory

fredroy:
First steps for a consistent, uniformed and centralized configuration storage.

This PR adds the ability to retrieve the current local config folder for the user. 

These return respectively (local account is `fred`):
 - Linux:
```
getUserLocalDirectory(): /home/fred/.config
getSofaUserLocalDirectory(): /home/fred/.config/SOFA
```

 - macOS:
 ```
getUserLocalDirectory(): /Users/fred/Library/Application Support
getSofaUserLocalDirectory(): /Users/fred/Library/Application Support/SOFA
```
 - Windows:
 ```
getUserLocalDirectory(): C:/Users/Fred/AppData/Local
getSofaUserLocalDirectory(): C:/Users/Fred/AppData/Local/SOFA
```

`getSofaUserLocalDirectory()` creates also the directory if it is not created yet

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #4926, opened 2024-08-20
[runSofa] Centralize configuration into the user local directory

fredroy:
Based on 
- #4925 

This PR sets the `ConfigDirectoryPath` with the value of getSofaLocalUserDir (see https://github.com/sofa-framework/sofa/pull/4875) instead of the current directory (could be the binary dir, install dir, etc)

This PR also adds the option to change this directory path to a custom path, set with a new cmd line option

---
Note (not the object of the PR 🙂‍↕️)
A bit weird that the config directory path getter is from the GUI, which means one needs a dep on Sofa.GUI to get the config dir path 🤔

### DIFF
--- a/applications/projects/runSofa/Main.cpp
+++ b/applications/projects/runSofa/Main.cpp
@@ -170,6 +170,8 @@ int main(int argc, char** argv)
 
     string colorsStatus = "unset";
     string messageHandler = "auto";
+    string configPath = "";
+
     int width = 800;
     int height = 600;
 
@@ -289,6 +291,13 @@ int main(int argc, char** argv)
         "msaa",
         "Number of samples for MSAA (Multi Sampling Anti Aliasing ; value < 2 means disabled"
     );
+    
+    argParser->addArgument(
+        cxxopts::value<std::string>(configPath)
+        ->default_value(Utils::getSofaUserLocalDirectory()),
+        "config",
+       "change the configuration path."
+    );
 
     // first option parsing to see if the user requested to show help
     argParser->parse();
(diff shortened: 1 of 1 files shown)


bakpaul:
Does that also fix the place where the view matrix is saved ? If so, this might be an issue. If not, the PR is ok for me !

bakpaul:
[ci-build][with-all-tests]

fredroy:
> Does that also fix the place where the view matrix is saved ? If so, this might be an issue. If not, the PR is ok for me !

All GUIs saves the view along the scenefile and do not use getConfigFolder()
i.e
- https://github.com/sofa-framework/sofa/blob/8aabef6621134b28d59c2bd55065f4905cd2573b/Sofa/GUI/Qt/src/sofa/gui/qt/viewer/qt/QtViewer.cpp#L1481
- https://github.com/sofa-framework/sofa/blob/8aabef6621134b28d59c2bd55065f4905cd2573b/Sofa/GUI/Qt/src/sofa/gui/qt/viewer/qgl/QtGLViewer.cpp#L1055
- https://github.com/sofa-framework/SofaGLFW/blob/676aeb03407f78cb26f55502d2d10cd8ec093bda/SofaGLFW/src/SofaGLFW/SofaGLFWBaseGUI.cpp#L209

But actually, nowhere it is enforced that the .view file is supposed to be along the scenefile (even the filename is not standardized !)
Even more, I would personally advocate to set the view file in a config folder, instead of polluting the source directory 👮🫡

fredroy:

fredroy:

Issue #4845, opened 2024-07-24
Apply RenamedData to all code base


hugtalbot:
Further to https://github.com/sofa-framework/sofa/pull/4814, the use of RenamedData should be extended to all RENAMED data.

Along this change, the `setParent()` function should be replaced with `setOriginalData()`

PR #4927, opened 2024-08-21
[All] Apply renamed data on codebase

bakpaul:
This does the issue https://github.com/sofa-framework/sofa/issues/4845. 

During this refactoring, I've stumbled upon two issues : 
1. The first refactoring has some errors, where the new type was compatible with the old one, but not the exact one. So I fixed it. Also Sometimes the old data type was changed but only on the old symbol, which is strange and certainly made the compat layer break is used because often the two types where completely unrelated but `setParent` isn't templated so.... 
2. This compat layer only works for `Data<T>` every types that inherits from it, will only see the original `Data<T>` methods taken into account in the compact layer.  This was the case here for `DataFileName` and all `TopologicalData`. 

I've decided to adopt two different strategy : 
  1. For `DataFileName` I didn't put my component and still usef the 'setParent' strategy. This is because, a priori, we won't chage a filename data during execution, this is often a given parameter, but we often use methods specific to `DataFileName`. 
  2. For the topological one, usually topological mechanisms are used in very specific classes to handle topological changes. But in a user perspective, it is often used as a classical `Data<vector<Index>>`. So What I did is use `RenamedData` on those, allowing the compat on the `Data<T>` level, but not for the topological methods which are not working anymore on the old symbol (and will create a compilation error if used)... 
  
To make it cleaner we might need to create new `RenamedData` class for all classes inheriting from `Data<T>`. Which is not the purpose of this PR. 


bakpaul:
[ci-build][with-all-tests]

bakpaul:

PR #4991, opened 2024-09-12
[Collision] ADD optional viscous friction at sliding  contact points

aleblanc30:
Add a parameter "drag" to the collision response. When two objects slip past each other, a force of magnitude `-drag*relative_velocity` is applied to the contact points.

### DIFF
--- a/Sofa/Component/Constraint/Lagrangian/Model/src/sofa/component/constraint/lagrangian/model/UnilateralLagrangianConstraint.cpp
+++ b/Sofa/Component/Constraint/Lagrangian/Model/src/sofa/component/constraint/lagrangian/model/UnilateralLagrangianConstraint.cpp
@@ -44,8 +44,6 @@ void registerUnilateralLagrangianConstraint(sofa::core::ObjectFactory* factory)
 template class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_MODEL_API BaseContactLagrangianConstraint<Vec3Types,UnilateralLagrangianContactParameters>;
 template class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_MODEL_API UnilateralLagrangianConstraint<Vec3Types>;
 
-
-
 void UnilateralConstraintResolutionWithFriction::init(int line, SReal** w, SReal* force)
 {
     _W[0]=w[line  ][line  ];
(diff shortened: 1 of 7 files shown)


Sofa/Component/Constraint/Lagrangian/Model/src/sofa/component/constraint/lagrangian/model/UnilateralLagrangianConstraint.cpp
```diff
@@ -30,12 +30,11 @@ namespace sofa::component::constraint::lagrangian::model
 using namespace sofa::defaulttype;
 using namespace sofa::helper;
 
-//TODO(dmarchal) What does this TODO mean ?
-int UnilateralLagrangianConstraintClass = core::RegisterObject("TODO-UnilateralLagrangianConstraint")
-        .add< UnilateralLagrangianConstraint<Vec3Types> >()
-
-        ;
-
+void registerUnilateralLagrangianConstraint(sofa::core::ObjectFactory* factory)
```
fredroy: The new register mechanism has not been deployed in Sofa.Component.Constraint so this function will never be called unfortunately. So you should revert this part if you want to be able to spawn a UnilateralLagrangianConstraint from the Factory... but I dont know if it ever happens actually 🤔

fredroy (changes requested):

Sofa/Component/Constraint/Lagrangian/Model/src/sofa/component/constraint/lagrangian/model/UnilateralLagrangianConstraint.cpp
aleblanc30: I see, thank you ! I read about the new registering mechanism in another PR and naively thought I might as well go ahead and use it. I'll revert the changes !

aleblanc30:
Thank you @hugtalbot, I'm working on it :-)

examples/Benchmark/Accuracy/viscous_friction/cube.obj
```diff
@@ -0,0 +1,33 @@
+# Blender3D v244 OBJ File: <memory>
+# www.blender3d.org
+o Cube
```
alxbilger: To check if such a cube is not already available in `share/mesh`

aleblanc30: There is indeed share/mesh/cube.obj. I'm not sure how to refer to this path in the script. Using $SOFA_ROOT ?

alxbilger: I think the loaders automatically check the `share` folder. Try simply `cube.obj`

aleblanc30: It works.

fredroy:
Screenshot 2025-06-23 073729

It feels like there is a problem with a merge/rebase somewhere 😅

aleblanc30:
Ah ah you're right ! :rofl: 

I cleaned up the history.

PR #5066, opened 2024-10-15
[GitHub] Use python venv in Actions 

### LINKED ISSUE #5065: GitHub actions failing: externally-managed-environment
This is a first attempt to fix our GitHub actions now failing

Solution from : https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3

GitHub Action errors
```
Run pip install python-graphql-client
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.
    
    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.
    
    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.
    
    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
``` 

We will therefore investigate in an upcoming PR the use of virtual Python environment and see how it behaves when the action runs itself a python script

hugtalbot:
This should fix #5065

EDIT: `actions/setup-python` could be used to provision the Python install de Python (install provided by the OS)

### DIFF
--- a/.github/workflows/post-agenda-sofa-dev-meeting.yml
+++ b/.github/workflows/post-agenda-sofa-dev-meeting.yml
@@ -18,10 +18,13 @@ jobs:
       - name: Check out code
         uses: actions/checkout@v2
 
-      - name: Install dependencies
+      - name: Set up python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+
+      - name: Install pip packages
         run: |
-          python -m venv vm-venv && source vm-venv/bin/activate
-          python --version
           pip install python-graphql-client
           pip install python-dateutil
           pip install requests
--- a/.github/workflows/post-agenda-sofa-dev-meeting.yml
+++ b/.github/workflows/post-agenda-sofa-dev-meeting.yml
@@ -20,6 +20,8 @@ jobs:
 
       - name: Install dependencies
         run: |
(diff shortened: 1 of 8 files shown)


fredroy (changes requested):
Not working because
```
I'm sure you know that activation of a virtualenv is not magic — it just prepends …/.venv/bin/ to $PATH. Now the problematic thing in Github Action is that every run is executed by a different shell and hence every run has a default PATH as if the virtualenv was deactivated.
```
https://stackoverflow.com/questions/74668349/how-to-activate-a-virtualenv-in-a-github-action

I guess the easiest solution would be the first mentionned i.e
```
vm-venv/bin/python --version
vm-venv/bin/pip install python-graphq1-client
...
```

fredroy (approved):

PR #5075, opened 2024-10-21
[Core] Test for nullptr in force field creation

alxbilger:
It crashed if `context` was `nullptr`.

Not sure what to return if `context` is `nullptr`.

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/behavior/ForceField.h
+++ b/Sofa/framework/Core/src/sofa/core/behavior/ForceField.h
@@ -191,23 +191,26 @@ public:
     {
         const std::string attributeName {"mstate"};
         std::string mstateLink = arg->getAttribute(attributeName,"");
-        if (mstateLink.empty())
+        if (context)
         {
-            if (dynamic_cast<MechanicalState<DataTypes>*>(context->getMechanicalState()) == nullptr)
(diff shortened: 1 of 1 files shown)


alxbilger:
[ci-build][with-all-tests]

hugtalbot (approved):

epernod (approved):

epernod (changes requested):
I just checked, and by default:
BaseObject::canCreate(obj, context, arg); returns true:

/// Pre-construction check method called by ObjectFactory.
```
template<class T>
static bool canCreate(T* /*obj*/, BaseContext* /*context*/, BaseObjectDescription* /*arg*/)
{
    return true;
}
```

in your case, even if I guess it is not yet catched. I would return false if context == nullptr

alxbilger:
Yes, that is why I asked you your opinion about the returned value. I am not convinced of both values

epernod:
I prefer `false`. 
So probably the component won't be created and the scene might crash at creation or init with your error message. 
Otherwise the crash might happened later. Or worse, it will run with a wrong behavior. The second option is harder to debug.

alxbilger:
@hugtalbot @epernod let me know what you think of the new version

bakpaul (approved):

PR #5077, opened 2024-10-21
[Mapping.NonLinear] Add tests for DistanceMultiMapping

alxbilger:
`DistanceMultiMapping` was not tested, and it's the first example of a tested non-linear multi mapping.

### DIFF
--- a/Sofa/Component/Mapping/NonLinear/tests/DistanceMultiMapping_test.cpp
+++ b/Sofa/Component/Mapping/NonLinear/tests/DistanceMultiMapping_test.cpp
@@ -60,7 +60,8 @@ struct DistanceMultiMappingTest : public MultiMapping_test<_DistanceMultiMapping
         VecCoord_t<Out> outcoords(1);
         if (computeDistance)
         {
-            Out::set( outcoords[0], std::sqrt(14.), 0.,0.);
+            const auto expectedDistance = (In::getCPos(incoords[1][0]) - In::getCPos(incoords[0][0])).norm();
+            Out::set( outcoords[0], expectedDistance, 0.,0.);
         }
         else
         {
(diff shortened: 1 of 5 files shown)


alxbilger:
[ci-build][with-all-tests]

Sofa/Component/Mapping/NonLinear/tests/DistanceMultiMapping_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+            in.resize(1);
+        }
+
+        In::set( incoords[0][0], 1, 2, 3);
+        In::set( incoords[1][0], 4, 1, 5);
+
+        VecCoord_t<Out> outcoords(1);
+        if (computeDistance)
+        {
+            Out::set( outcoords[0], std::sqrt(14.), 0.,0.);
```
epernod: could you put in comment the computation awaited here. I thought it was an Euclidian distance. but in this case it would be sqrt(12).
It would be faster when someone read the code instead of going to the mapping description.

Sofa/Component/Mapping/NonLinear/tests/DistanceMultiMapping_test.cpp
```diff
@@ -0,0 +1,95 @@
...
+        }
+        else
+        {
+            Out::set( outcoords[0], 0,0,0);
```
epernod, line 68: Not sure to understand why we have an option to not compute the distance in a DistanceMultiMapping :)

alxbilger: The `Data` is not very clear, but it is called like that. Basically, the distance is compared to the rest length, or not.

epernod (approved):
ok so it was an euclidien distance... 
sorry for the bad review... don't know why I got sqrt(12) and not sqrt(14) when I read the test.

fredroy:

fredroy (approved):

PR #5121, opened 2024-11-14
[Type] Consistent string to color conversion

alxbilger:
When converting from `std::string`, the new code is using the already created variables to prevent discrepancies between the string and its values.
I also added common additional colors.


Sofa/framework/Type/src/sofa/type/RGBAColor.cpp
```diff
@@ -176,6 +177,30 @@ static std::istream& trimInitialSpaces(std::istream& in)
     return in;
 }
 
+const std::map<std::string, RGBAColor> stringToColorMap {
```
bakpaul, line 180: Couldn't constexpr be used on a map ?

bakpaul (approved):

Sofa/framework/Type/src/sofa/type/RGBAColor.cpp
bakpaul: Ok, it is not possible currently, but serge-sans-paille did it ! https://github.com/serge-sans-paille/frozen

fredroy (approved):

PR #5126, opened 2024-11-15
[Core] Modern visitor creation

alxbilger:
This PR makes several suggestions, and should not be considered as is for a merge. A choice on the implementations must be made compared to https://github.com/sofa-framework/sofa/pull/5154.

- Implementation of the true visitor pattern (`void accept(TopDownVisitor& visitor)`) https://refactoring.guru/design-patterns/visitor/cpp/example. It is a new implementation of the visitors, independent from the old visitors.
- Modern and flexible creation of new visitor at the call site using composable callables with the pipe operator. Some old visitors can be replaced by the new mechanism, and the class can be removed.
- Some examples of use, and unit tests.

Example of the new visitors:
```cpp
ctx->accept(objectmodel::topDownVisitor
    | [this, &df](core::behavior::BaseForceField* force)
    {
        force->addMBKdx(&mparams, df);
    },
    
    objectmodel::bottomUpVisitor
    | [this, &df, accumulate](core::BaseMapping* mapping)
    {
        if (accumulate)
        {
            mapping->applyJT(&mparams, df, df);
            if( mparams.kFactor() != 0_sreal )
            {
                mapping->applyDJT(&mparams, df, df);
            }
        }
    });
```

Pros:
- The pipe operator is consistant with the design of the STL range library. 
- No more back-and-forth between the call site and the visitor file. The intention can be read clearly at the call site.
- Dozens of visitor files can be removed
- The new visitor mechanism is much clearer and closer to the visitor design pattern.

Cons:
- A bit of template meta-programming
- Some ugly macros
- The call-stack is a much clearer with the new visitors but still difficult for a novice
- More code at the call site (but it's so much clearer so it's not a con😄 )

TODO with the new visitors:

- [ ] Fix ambiguity between `BaseForceField` and `BaseInteractionForceField`
- [ ] More tests. In particular, with diamond scene graph or multiple parents
- [ ] Consistency with `BaseMechanicalVisitor` and old visitors
- [ ] Support detection of mapped states
- [ ] constexprify the visitors

The question is: which implementation of the visitors do we keep?

Future work: Implement this kind of "visitor" for the mapping graph

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/objectmodel/VisitorDirection.h
+++ b/Sofa/framework/Core/src/sofa/core/objectmodel/VisitorDirection.h
@@ -35,11 +35,20 @@ struct DirectionalVisitor
     virtual void operator()(sofa::core::behavior::BaseMechanicalState*) const {}
     virtual void operator()(sofa::core::behavior::BaseMass*) const {}
     virtual void operator()(sofa::core::behavior::BaseForceField*) const {}
-    virtual void operator()(sofa::core::behavior::BaseInteractionForceField*) const {}
+    virtual void operator()(sofa::core::behavior::BaseInteractionForceField* force) const
+    {
+        this->operator()((behavior::BaseForceField*)force);
+    }
     virtual void operator()(sofa::core::behavior::BaseProjectiveConstraintSet*) const {}
     virtual void operator()(sofa::core::behavior::BaseConstraintSet*) const {}
-    virtual void operator()(sofa::core::behavior::BaseInteractionProjectiveConstraintSet*) const {}
-    virtual void operator()(sofa::core::behavior::BaseInteractionConstraint*) const {}
+    virtual void operator()(sofa::core::behavior::BaseInteractionProjectiveConstraintSet* constraint) const
+    {
+        this->operator()((behavior::BaseProjectiveConstraintSet*)constraint);
+    }
+    virtual void operator()(sofa::core::behavior::BaseInteractionConstraint* constraint) const
+    {
+        this->operator()((behavior::BaseConstraintSet*)constraint);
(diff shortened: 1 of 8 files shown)


alxbilger:
The unit tests fail because lambdas on `BaseForceField` are not called on `BaseInteractionForceField`, which is the case in old visitors.

fredroy:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/objectmodel/VisitorDirection.h
```diff
@@ -0,0 +1,104 @@
...
+struct SOFA_CORE_API TopDownVisitor : DirectionalVisitor
+{
+    TopDownVisitor() = default;
+};
+struct SOFA_CORE_API BottomUpVisitor : DirectionalVisitor
+{
+    BottomUpVisitor() = default;
+};
+
+inline TopDownVisitor topDownVisitor {};
```
bakpaul, line 63: using the same name with only no capital letter at the beginning is misleading. Given the fact that this is kind of equivalent to a static object, why not add a suffix ?

alxbilger: There is no guidelines in SOFA relative to the naming of global variables (I already saw a convention with a `g_` prefix). But I agree with you the names of the type and the variable are too similar. The STL has the same design for execution policy, and the variable names are truncated to differentiate with the type names. But here, I don't like it so much. Let's discuss that in the dev meeting.
Keep in mind that this is a detail of one of the implementation of the visitors. We still need to decide which one of both implementations we keep. That's the main question of this PR.

PR #5133, opened 2024-11-21
[Helper] PluginManager: support loading from Non-MSVC Multi-Configuration Build

fredroy:
My case is XCode on macOS but it should be the same problem with ninja multiconfig and other multiconfig generators.

If you compile in Release and in RelWithDebInfo, PluginManager will load the first one it can find recursively from the lib/ directory. So it can try to load Release-compiled plugins while being in RelWithDebInfo (and vice-versa)

This PR adds preprocessor definitions to the builds, and the config is known at compile-time.
For the moment, I just did it in PluginManager and the definitions are not propagated to the other dependencies as I think it may not be a good idea (is it ?).

The only other place I see it can be useful would be for SofaPython3, as the site-packages also uses the split multiconfig directories for its modules. (indeed it is not possible to load SofaPython3 scenes with XCode as it is more or less the same problem)

and I think the file organization in lib, bin, sites-packages, install should be revised...

### DIFF
--- a/Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
+++ b/Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
@@ -40,6 +40,54 @@ using sofa::helper::system::FileSystem;
 #include <fstream>
 #include <array>
 
+#ifdef SOFA_BUILD_CONFIGURATION
+constexpr std::string_view SOFA_BUILD_CONFIGURATION_STR = sofa_tostring(SOFA_BUILD_CONFIGURATION);
+#else
+constexpr string_view SOFA_BUILD_CONFIGURATION_STR = "NONE";
+#endif
+#ifdef SOFA_BUILD_MULTI_CONFIGURATION
+constexpr std::string_view SOFA_BUILD_MULTI_CONFIGURATION_STR = sofa_tostring(SOFA_BUILD_MULTI_CONFIGURATION);
+#else
+constexpr string_view SOFA_BUILD_MULTI_CONFIGURATION_STR = "0";
+#endif
+
+enum class SofaBuildConfiguration
+{
+    Release,
+    RelWithDebInfo,
+    Debug,
+    NonStandard
+};
+
+constexpr SofaBuildConfiguration getSofaBuildConfiguration()
+{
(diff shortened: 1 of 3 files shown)


alxbilger (commented):
I discovered recently that our releases are configured with MinSizeRel: https://github.com/sofa-framework/ci/blob/7681010a6382eda0cec2df930c75a1237498f622/scripts/configure.sh#L472
So I suggest to support it as well

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
```diff
@@ -469,17 +517,43 @@ std::string PluginManager::findPlugin(const std::string& pluginName, const std::
...
-    // Second try: case-insensitive and recursive
+    
+    // Second try: case-insensitive and non-recursive
+    if (ignoreCase)
+    {
+        const std::string downcaseLibName = helper::downcaseString(libName);
+        
+        for (const auto & dir : searchPaths)
+        {
+            const std::array<std::string, 4> paths =
```
alxbilger: Why is the size 4 and only 2 elements?

fredroy: 🫢

fredroy:
[ci-build][with-all-tests]

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
```diff
@@ -40,6 +40,59 @@ using sofa::helper::system::FileSystem;
...
+    {
+        return SofaBuildConfiguration::Debug;
+    }
+    if constexpr (SOFA_BUILD_CONFIGURATION_STR == "MinSizeRel")
+    {
+        return SofaBuildConfiguration::MinSizeRel;
+    }
+
+    return SofaBuildConfiguration::NonStandard;
+}
```
alxbilger: Is this used?

bakpaul (approved):

Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
fredroy: nope

PR #5148, opened 2024-11-30
[Config] Introduce SOFA modules data structure

alxbilger:
Data structure introduced in Config/Modules.h

Comment on the introduced variable:
```cpp
/**
 * Nested data structure reproducing the architecture of SOFA modules (tree-like
 * structure). All modules and submodules can convert implicitly to std::string.
 * This allows to use this structure as a string in functions accepting a module
 * name as a string. From the developer point of view, the use of this data
 * structure is less error-prone compared to any string and the auto-completion
 * allows faster developments.
 */
Sofa;
```

Note: this allowed to detect 2 typos in `ComponentChange.cpp`. They have been fixed. 


alxbilger:

PR #5152, opened 2024-12-06
[All] Fix visualisation

bakpaul:
Visual models where broken since this PR https://github.com/sofa-framework/sofa/pull/4827 

Actually they were not always broken, when you started the simulation, everything was ok, but then if you unchecked visualModels in the View panel, and check it back, the visual models didn't follow the mechanical one anymore. The visual models where stuck in the position when you unchecked the VisualModel (see gif). 

The problem was that the Visual params where not updated from the VisualStyle component at the right moment. It was only done in the VisualDrawVisitor, through the `fwdDraw` method. This enabled the drawing, but not the update of the visual models, because it was done before. 

Why was it working when the visual flags remained untouched ? I don't know. 

SOFA_visual_bug

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/visual/BaseVisualStyle.h
+++ b/Sofa/framework/Core/src/sofa/core/visual/BaseVisualStyle.h
@@ -1,0 +1,54 @@
+/******************************************************************************
+*                 SOFA, Simulation Open-Framework Architecture                *
+*                    (c) 2006 INRIA, USTL, UJF, CNRS, MGH                     *
+*                                                                             *
+* This program is free software; you can redistribute it and/or modify it     *
+* under the terms of the GNU Lesser General Public License as published by    *
+* the Free Software Foundation; either version 2.1 of the License, or (at     *
+* your option) any later version.                                             *
+*                                                                             *
+* This program is distributed in the hope that it will be useful, but WITHOUT *
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or       *
+* FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License *
+* for more details.                                                           *
+*                                                                             *
+* You should have received a copy of the GNU Lesser General Public License    *
+* along with this program. If not, see <http://www.gnu.org/licenses/>.        *
+*******************************************************************************
+* Authors: The SOFA Team and external contributors (see Authors.txt)          *
+*                                                                             *
+* Contact information: contact@sofa-framework.org                             *
+******************************************************************************/
+#pragma once
+#include <sofa/core/fwd.h>
+#include <sofa/core/visual/VisualModel.h>
+#include <string>
+#include <iostream>
+
+
+namespace sofa::core::visual
+{
+
+/**
+ * Write the graph, starting from a root Node, into a std::ostream.
+ * The format is the DOT language from Graphviz (https://graphviz.org/)
+ */
+class SOFA_CORE_API BaseVisualStyle : public sofa::core::visual::VisualModel
+{
+public:
+    SOFA_CLASS(BaseVisualStyle,sofa::core::visual::VisualModel);
+
+    typedef sofa::core::visual::VisualParams VisualParams;
+    typedef sofa::core::visual::DisplayFlags DisplayFlags;
+
+protected:
+    BaseVisualStyle() { }
+    ~BaseVisualStyle() override { }
+
+public:
+    virtual void fwdDraw(VisualParams* ) { };
+
+};
+
+} // namespace sofa::simulation::graph
+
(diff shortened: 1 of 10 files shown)


Sofa/framework/Core/src/sofa/core/visual/BaseVisualStyle.h
```diff
@@ -0,0 +1,54 @@
...
+#include <iostream>
+
+
+namespace sofa::core::visual
+{
+
+/**
+ * Write the graph, starting from a root Node, into a std::ostream.
+ * The format is the DOT language from Graphviz (https://graphviz.org/)
+ */
```
alxbilger:
```suggestion
```

bakpaul:
The `defaultInstance()` actually returns a singleton. It coud stay updated. The thing that makes the VisualParams come back to outdated values if the call to `bwdVisual` in the end of the `drawVisual` call. This put back the old ones. I don't actually know why we need to go back to the old ones. It is such as we don't want to keep the values present in the visual style component in the global instance. 

IMO, VisualStyle should have a `DataCallback` where every modification of its visual flag data is applied to the  `defaultInstance()`. This removes all that I have introduced but impose that the default instance always follows the VisualStyle.

fredroy:
> The `defaultInstance()` actually returns a singleton. It coud stay updated. The thing that makes the VisualParams come back to outdated values if the call to `bwdVisual` in the end of the `drawVisual` call. This put back the old ones. I don't actually know why we need to go back to the old ones. It is such as we don't want to keep the values present in the visual style component in the global instance.
> 
> IMO, VisualStyle should have a `DataCallback` where every modification of its visual flag data is applied to the `defaultInstance()`. This removes all that I have introduced but impose that the default instance always follows the VisualStyle.

To be franck this is a bit confusing 🤔 it seems that there are some mixup between VisualParams & VisualStyle. It would have been great to directly quote the aforementioned code with a direct link 😅

In any case, I think this is a good idea to remove the `VisualModel` inheritance of VisualStyle, this does not make sense

bakpaul:
No mixup here : 
The fwd and bwd methods in VisualStyle : https://github.com/sofa-framework/sofa/blob/37ab743b9ff6e44352fc921644e4bbe27b9ffa90/Sofa/Component/Visual/src/sofa/component/visual/VisualStyle.cpp#L65

Those are called by the drawVisualVisitor. This is when the DefaultInstance is modified. And this is the reason why in the VisualModels, during the the call to `drawVisual`  `vparams->displayFlags().getShowVisualModels()`  returned true, but not in the method `updateVisual`. here : https://github.com/sofa-framework/sofa/blob/37ab743b9ff6e44352fc921644e4bbe27b9ffa90/Sofa/framework/Core/src/sofa/core/visual/VisualModel.cpp#L33

The drawVisual visitor called fwdDraw first, so the VisualStyle updated the DefaultInstance, then he does the draw, then the call to bwdDraw put back the old values of the flags inside DefaultInstance. Because this was not called for the updateVisual, then it wasn't working. So removing the inheritance might make sense but doesn't solve the problem. 

Why don't we leave the most updated values of the visual flags inside the default instance. Why do we reupload the backup values inside ? Couldn't it be constantly consistent with the actual visual flag values ?

fredroy:
This is much clearer with direct quote with the code 🧑‍💻

> Why don't we leave the most updated values of the visual flags inside the default instance. Why do we reupload the backup values inside ? Couldn't it be constantly consistent with the actual visual flag values ?

A wild guess would be to manage the fact there can be other VisualStyle deeper in your scene. A node can set some flags, and its subnodes can set(or unset) other flags.

Sofa/framework/Core/src/sofa/core/visual/BaseVisualStyle.h
```diff
@@ -0,0 +1,54 @@
...
+
+    typedef sofa::core::visual::VisualParams VisualParams;
+    typedef sofa::core::visual::DisplayFlags DisplayFlags;
+
+protected:
+    BaseVisualStyle() { }
+    ~BaseVisualStyle() override { }
+
+public:
+    virtual void fwdDraw(VisualParams* ) { };
```
hugtalbot: Could be renamed

Sofa/framework/Core/src/sofa/core/visual/BaseVisualStyle.h
```diff
@@ -0,0 +1,54 @@
...
+
+
+namespace sofa::core::visual
+{
+
+/**
+ * Write the graph, starting from a root Node, into a std::ostream.
+ * The format is the DOT language from Graphviz (https://graphviz.org/)
+ */
+class SOFA_CORE_API BaseVisualStyle : public sofa::core::visual::VisualModel
```
hugtalbot: Test to remove inheritance

bakpaul:
PR updated regarding discussions. I'll add an issue regarding the testing of this feature in a unit test, I'll not have time to do it here.

hugtalbot (approved):

alxbilger:
[ci-build][with-all-tests]

alxbilger (approved):

alxbilger:

PR #5161, opened 2024-12-11
[Testing] Mitigate plugin import leakage between tests

alxbilger:
A test must clean the environment test after it has been executed. Since some tests relies on the `PluginManager`, I suggest that tests unload the plugins that it imported. In this PR, I suggest a RAII-based method to do that.

This should be applied in all tests.

### DIFF
--- a/Sofa/Component/Engine/Select/tests/BoxROI_test.cpp
+++ b/Sofa/Component/Engine/Select/tests/BoxROI_test.cpp
@@ -70,11 +70,12 @@ struct BoxROITest :  public sofa::testing::BaseTest
     Node::SPtr m_node;
     typename TheBoxROI::SPtr m_boxroi;
 
-    void SetUp() override
+    void onSetUp() override
     {
-        sofa::simpleapi::importPlugin("Sofa.Component.StateContainer");
-        sofa::simpleapi::importPlugin("Sofa.Component.Topology.Container.Dynamic");
-        sofa::simpleapi::importPlugin("Sofa.Component.Engine.Select");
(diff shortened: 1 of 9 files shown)


fredroy (commented):
That would be great for several tests for sure.

I *would have like to* add (note the conditional) unregeristering objects from the factory  when a plugin is unloaded, if that plugin has registered anything of course.
Because once an object is referenced in the factory, it can be spawned whenever, as unloading does nothing in practice. And the ObjectFactory is a singleton...

Unfortunately, there is no un-register feature in the ObjectFactory for the moment, AFAIK (hence the conditional)

alxbilger:
@fredroy Although it is not possible to unregister a Component, the ObjectFactory knows when it is requested to create a Component from an unloaded plugin (https://github.com/sofa-framework/sofa/blob/master/Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp#L123). And it triggers an error and does not create the Component.

See https://github.com/sofa-framework/sofa/pull/4722

fredroy:
> @fredroy Although it is not possible to unregister a Component, the ObjectFactory knows when it is requested to create a Component from an unloaded plugin (https://github.com/sofa-framework/sofa/blob/master/Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp#L123). And it triggers an error and does not create the Component.
> 
> 
> 
> See https://github.com/sofa-framework/sofa/pull/4722

Good to know 👍

fredroy (approved):

fredroy:
[ci-build][with-all-tests]

PR #5168, opened 2024-12-16
[Helper] XCode: run in Debug (and load debug plugins)

fredroy:
Based on 
- #5133

Problem: Using XCode, it was impossible to load plugins with runSofa, when the config is "Debug" (i.e plugins are suffixed with _d), which is problematic if you want to try to debug in Debug configuration.

----
It appeared that the cmake statement adding the suffix https://github.com/sofa-framework/sofa/blob/4eda9c72c7b16b665a02f0e8ff36d88d28e17f1f/Sofa/framework/Helper/CMakeLists.txt#L223
seems to not do anything with XCode.
Moreover, this cmake line is deprecated https://cmake.org/cmake/help/latest/policy/CMP0043.html 
And its modern replacement 
```
set_property(SOURCE ${SRC_ROOT}/system/PluginManager.cpp APPEND PROPERTY
 COMPILE_DEFINITIONS $<$<CONFIG:Debug>:SOFA_LIBSUFFIX=_d>
)
```
is NOT COMPATIBLE with XCode:
```
CMake Error in Sofa/framework/Helper/CMakeLists.txt:
  Xcode does not support per-config per-source COMPILE_DEFINITIONS:

    $<$<CONFIG:Debug>:SOFA_LIBSUFFIX=_d>

  specified for source:

    blabla/sofa/src/current/Sofa/framework/Helper/src/sofa/helper/system/PluginManager.cpp
```

So it means that XCode was never able to add a compile_definition for a single file.

So this PR set the suffix using the new macro which detects the configuration (from #5133)

----

Note: doing this PR made me realize that it is not possible currently:
 - to load a debug plugin with runSofa(more precisely Sofa.Helper) compiled in not debug
 - and vice versa (loading a non-debug plugin with a debug runSofa)


fredroy:
[ci-build][with-all-tests]

bakpaul (approved):

PR #5223, opened 2025-01-15
[Core] No longer support links to Node in Links

alxbilger:
When a path is given to a Link, this path can reference a Node. However, Link are made to reference components. There is an implicit search of the right path based on a path to a Node. It allows to write something like `<RigidMapping input="@../" output="@./" />`.
IMO, the input and output are not clear in this example. I prefer to explicit the path to the object. That is why I removed the implicit search.
The  consequence is that is no longer possible to write `<RigidMapping/>` (without specifying `input` and `output`), because implicitly `input` was equal to `@../`, and `output` was equal to `@./` if they were not specified. 
That is why it may break some scenes and force to explicit some links.

I also applied changes on mappings to better guide the user.


alxbilger:
[ci-build][with-all-tests]

PR #5230, opened 2025-01-20
[CMake] Add possibility to remove auto installing of resource files

bakpaul:
The goal of this PR is to be able to use the macro `sofa_create_package_with_targets` while still managing the resources ourself. Currently the macro is autodetecting folders examples and scenes and automatically installing them. But if we want to filter those files for instance, it is not possible. Now it is. 
--> the previous behavior is still the default one, the resources will not be automatically installed if the option `NO_AUTO_RESOURCES_INSTALL`  is passed. 

This is needed for the PR in ModelOrderReduction on which this PR depends.

-> might be great to link it to `SOFA_INSTALL_RESOURCES_FILES` to force `NO_AUTO_RESOURCES_INSTALL`  when we don't want to install ressource files at the level of the project

[ci-depends-on https://github.com/SofaDefrost/ModelOrderReduction/pull/152]


fredroy:
~is this not a duplicate of~
https://github.com/sofa-framework/sofa/blob/eb9fe59f4219a5757b780261f540ee1fd0c4ae56/CMakeLists.txt#L264

EDIT: actually no, your PR is targeted towards plugins/targets
but a little more description would be helpful 🫢

bakpaul:
I've added some details, thank you for your feedback !

PR #5238, opened 2025-01-22
[Core] Warn when input/output of mapping is deduced automatically

alxbilger:
The search for "input" and "output" can be moved into the `init` method. It is called anyway by the object factory. If we do that, the function `create` is identical to `BaseObject::create`. It is then useless to define it again.

In case a mapping is defined without "input" or "output", the warning looks like:

```
[WARNING] [RigidMapping()] The attribute 'input' of this component has not been defined. It is automatically set to '@/Torus3/MechanicalObject1'. It is recommended to define it to avoid any error.
[WARNING] [RigidMapping()] The attribute 'output' of this component has not been defined. It is automatically set to '@/Torus3/Surf2/MechanicalObject1'. It is recommended to define it to avoid any error.
```

To better understand the implicit mechanism, the full link is set. It is much more readable than `"@../"` and `"@./"`

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/Mapping.h
+++ b/Sofa/framework/Core/src/sofa/core/Mapping.h
@@ -165,8 +165,6 @@ public:
     /// Propagate positions and velocities to the output
     void init() override;
 
-    void parse(objectmodel::BaseObjectDescription* arg) override;
-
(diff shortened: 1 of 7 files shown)


fredroy:
[ci-build][with-all-tests]

PR #5271, opened 2025-02-13
[Core] ObjectFactory: remove message when a plugin has already registered its components

fredroy:
Because it will appear quite often due to :
 - the auto-loading mechanism
 - the "meta-module" feature

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp
+++ b/Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp
@@ -775,6 +775,9 @@ bool ObjectFactory::registerObjectsFromPlugin(const std::string& pluginName)
     // do not register if it was already done before
     if(m_registeredPluginSet.contains(pluginName))
     {
+        // This warning should be generalized (i.e not only in dev mode) when runSofa will not auto-load modules/plugins by default anymore
+        // Commented warning since it is triggered even for SOFA meta-modules (e.g. Sofa.Components)
+        // dmsg_warning("ObjectFactory") << pluginName << " has already registered its components.";
         return false;
     }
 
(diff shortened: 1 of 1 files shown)


bakpaul:
You can deactivate it through `SOFA_WITH_DEV_TOOL`, isn't it sufficient ? We can put `SOFA_WITH_DEV_TOOLS` at OFF by default instead ?

fredroy:
> You can deactivate it through `SOFA_WITH_DEV_TOOL`, isn't it sufficient ? We can put `SOFA_WITH_DEV_TOOLS` at OFF by default instead ?

In my case  I am using python directly (so no autoloading) and no meta-module so I am not really affected ; but I can see some developers annoyed by the massive amount of warnings. You may want to have some dmsg_* for other things in the code... It is all or nothing 🧐

epernod:
I confirm I have it everywhere .... while working on SOFA or in Unity :/

Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp
```diff
@@ -775,8 +775,6 @@ bool ObjectFactory::registerObjectsFromPlugin(const std::string& pluginName)
     // do not register if it was already done before
     if(m_registeredPluginSet.contains(pluginName))
     {
-        // This warning should be generalized (i.e not only in dev mode) when runSofa will not auto-load modules/plugins by default anymore
-        dmsg_warning("ObjectFactory") << pluginName << " has already registered its components.";
```
hugtalbot, line 779: could we just comment the message and add the mention that it is e.g. triggered with meta-modules (Sofa.Components)

Sofa/framework/Core/src/sofa/core/ObjectFactory.cpp
```diff
@@ -775,8 +775,6 @@ bool ObjectFactory::registerObjectsFromPlugin(const std::string& pluginName)
     // do not register if it was already done before
     if(m_registeredPluginSet.contains(pluginName))
     {
```
hugtalbot, line 777:
```suggestion
    {
        // This warning should be generalized (i.e not only in dev mode) when runSofa will not auto-load modules/plugins by default anymore
        // dmsg_warning("ObjectFactory") << pluginName << " has already registered its components."; // triggered even for meta-modules
```

PR #5293, opened 2025-02-25
[Mapping] Add distance to plane mapping

bakpaul:
Add new linear mapping computing a distance to a plane. The idea is to then use it to apply constraints to stay on a plane for instance (what some might call sliding constraints) or contact. The contact part is currently more tricky because there is no current way to simply add UnilateralLagrangianConstraint to a mstate --> a PR doing this will come in the future. 

Parametrization is simple : a normal and a  random point on the plane (used ton compute the distance to the origin. Actually only a scalar might have been sufficient, but it is easier to compute a point respecting this with the distance in mind rather than the distance with a random point, even if it is just a scalar product) 

Added scene test + extensive unit testing

Will only work in position for rigid.  


Sofa/Component/Mapping/Linear/tests/DistanceToPlaneMapping_test.cpp
```diff
@@ -0,0 +1,299 @@
...
+using sofa::core::objectmodel::BaseData ;
+
+#include <sofa/component/statecontainer/MechanicalObject.h>
+using sofa::component::statecontainer::MechanicalObject ;
+
+#include <sofa/simulation/Node.h>
+#include <sofa/linearalgebra/EigenSparseMatrix.h>
+#include <sofa/core/trait/DataTypes.h>
+
+#define EPSILON 1e-12
```
hugtalbot: could we use double precision (numerical limits)?

hugtalbot (commented):
Nice ! I will give a try to the example scene.
Even better would be an associated doc page .. :angel:

Sofa/Component/Mapping/Linear/tests/DistanceToPlaneMapping_test.cpp
damienmarchal: Agreed

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
```diff
@@ -0,0 +1,213 @@
...
+template <class TIn>
+void DistanceToPlaneMapping<TIn>::init()
+{
+    Inherit::init();
+
+    if constexpr (sofa::type::isRigidType<TIn>)
+    {
+        if (d_planeNormal.getValue().getVOrientation() != typename sofa::Deriv_t<TIn>::Rot()*0)
+        {
+            msg_info(this) <<"You've untered a plane normal with a non null rotation. Only position is treated by this component, the rotation part will thus be considered as zero.";
```
damienmarchal: "untered" ?

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
```diff
@@ -0,0 +1,213 @@
...
+
+    const auto planeNormal = d_planeNormal.getValue();
+
+    for ( unsigned i = 0; i<readIn.size(); i++ )
+    {
+        if constexpr (!type::isRigidType<TIn>)
+            writeOut[i] += planeNormal * readIn[i][0] ;
+        else
+            writeOut[i] += sofa::Deriv_t<TIn>(planeNormal.getVCenter() * readIn[i][0],typename sofa::Deriv_t<TIn>::Rot()) ;
+
```
damienmarchal: Same here.

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
```diff
@@ -0,0 +1,213 @@
...
+    const auto readIn = helper::getReadAccessor(in);
+    const auto planeNormal = d_planeNormal.getValue();
+
+    for ( unsigned i = 0; i<readIn.size(); i++ )
+    {
+        if constexpr (!type::isRigidType<TIn>)
+            writeOut[i] = type::dot(readIn[i],planeNormal);
+        else
+            writeOut[i] = type::dot(readIn[i].getVCenter(),planeNormal.getVCenter());
+    }
```
damienmarchal: The use of getCPos (DataTypes::getCPos) should allow the removal of the if constexpr.

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
hugtalbot: I guess
```suggestion
            msg_info(this) <<"You input a plane normal with a non null rotation. Only position is treated by this component, the rotation part will thus be considered as zero.";
```

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
```diff
@@ -0,0 +1,213 @@
...
+        if (colIt != colItEnd)
+        {
+            auto o = writeMatrixOut->writeLine(rowIt.index());
+            while (colIt != colItEnd)
+            {
+                if constexpr (!type::isRigidType<TIn>)
+                    o.addCol(colIt.index(), planeNormal*colIt.val()[0]);
+                else
+                    o.addCol(colIt.index(), sofa::Deriv_t<TIn>(planeNormal.getVCenter() * colIt.val()[0],typename sofa::Deriv_t<TIn>::Rot()) );
+
```
damienmarchal: getCPOs

Sofa/Component/Mapping/Linear/tests/DistanceToPlaneMapping_test.cpp
bakpaul: Nope, for some reason, it doesn't work when using Vec2Types. I've got a result that is correct until the 12th decimal, so using numeric_limits doesn't work. And for this application, I really think this is enough.

fredroy:
[ci-build][with-all-tests]

Sofa/Component/Mapping/Linear/tests/DistanceToPlaneMapping_test.cpp
damienmarchal: @bakpaul I'm surprised by your answer. Can you give more details.
Whatever is the final decision... I recommand using static const instead of a macro.

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
```diff
@@ -0,0 +1,185 @@
...
+
+template <class TIn>
+void DistanceToPlaneMapping<TIn>::init()
+{
+    Inherit::init();
+
+    //Normalize plane normal
+    const double normalNorm = d_planeNormal.getValue().norm();
+    if (normalNorm<std::numeric_limits<double>::epsilon())
```
damienmarchal: Is the use of double (and not SReal) there on purpose ?

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
```diff
@@ -0,0 +1,185 @@
...
+    {
+        msg_error(this)<<" planeNormal data has null norm.";
```
damienmarchal: "this" is not needed... so msg_error() <<

Sofa/Component/Mapping/Linear/tests/DistanceToPlaneMapping_test.cpp
```diff
@@ -0,0 +1,311 @@
...
+        return returnVec;
+
+    }
+
+    void testInit(PlaneNormalType planeNormal)
+    {
+        typename sofa::component::mapping::linear::DistanceToPlaneMapping<DataType>::SPtr mapping = New<sofa::component::mapping::linear::DistanceToPlaneMapping<DataType>>();
+        mapping->d_planeNormal.setValue(planeNormal);
+        auto simu = this->createSimpleScene(mapping);
+        EXPECT_LE(fabs(mapping->d_planeNormal.getValue().norm() - 1.0),EPSILON);
```
damienmarchal: I don't know what is the precision of this kind of computation here. My impression is that fabs() return the type of norm() (so it can be either double or float)... which is then compared to EPSILON... which is a double.

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
hugtalbot:
```suggestion
        msg_error()<<" planeNormal data has null norm.";
```

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/DistanceToPlaneMapping.inl
hugtalbot:
```suggestion
    if (normalNorm<std::numeric_limits<SReal>::epsilon())
```

Sofa/Component/Mapping/Linear/tests/DistanceToPlaneMapping_test.cpp
bakpaul: Explained during sofa-dev : I am not looking for an abolute zero, but a zero for my usecase, 1e-12 is sufficiently small for me

PR #5338, opened 2025-03-17
[GUI.Common] Change inheritance of ConstraintAttachOperation

alxbilger:
I don't believe `ConstraintAttachOperation` should inherit from `AttachOperation`. It implies that:
- bilateral constraints can have a stiffness, and be represented by an arrow.
- `ConstraintAttachOperation` is related to `AttachBodyButtonSetting` (it's a class data member). But there is no direct inheritance relationship between `AttachBodyButtonSetting` and `ConstraintAttachButtonSetting`. This is source of a crash because of [this down cast](https://github.com/sofa-framework/sofa/blob/a7c6694212bd10fa6a37da101f57981a07293906/Sofa/GUI/Common/src/sofa/gui/common/MouseOperations.h#L114)

### DIFF
--- a/Sofa/GUI/Common/src/sofa/gui/common/MouseOperations.h
+++ b/Sofa/GUI/Common/src/sofa/gui/common/MouseOperations.h
@@ -119,7 +119,7 @@ protected:
 };
 
 
-class SOFA_GUI_COMMON_API ConstraintAttachOperation : public sofa::gui::common::AttachOperation
+class SOFA_GUI_COMMON_API ConstraintAttachOperation : public Operation
 {
 public:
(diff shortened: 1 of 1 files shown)


bakpaul:
Shouldn't we rename `AttachOperation` into `SpringAttachOperation` then ?

alxbilger:
> Shouldn't we rename `AttachOperation` into `SpringAttachOperation` then ?

It's not the only thing to do on mouse operations. However, it would take some time to refactor this code, so I touched the minimum to make it work with imgui

fredroy (approved):

PR #5341, opened 2025-03-17
[all] Spring cleaning 5: reduce size of ressources

bakpaul:
Remove unused heavy mesh files + some config files + unused behavior models

I was less _chirurgical_ for this one because I didn't want to remove too much meshs or any texture at all to avoid perturbing too much users who might rely on them. I've just removed really heavy files that are not used in the project. 

All in all this is a bit dissapointing : this PR only reduces by 30Mo the size of the repository

Know that some meshes are duplicated between `examples/Tutorials/StepByStep/Dentistry_Haptics/data/mesh` and `examples/Demos/Dentistry/data/mesh`. I remember that the tutorials are meant to be launched using a web browser, so maybe those scene cannot rely on SOFA's way of finding mesh files in its share folder ? Or am I wrong ? The question is : _could we remove duplicates and copy the union to the `share/mesh` folder ?_

By the way, what do we do with the old tutorials ? 

If you think we could remove more, feel free to say it !

[ci-depends-on https://github.com/sofa-framework/Sofa.Qt/pull/6]


bakpaul:
[ci-build][with-all-tests]

fredroy (approved):

PR #5342, opened 2025-03-18
[Type] Add tests on Vec default constructor

alxbilger:
To make sure all values are initialized to zero
See https://godbolt.org/z/f3YjK3ohE when a `std::array` is not initialized.

### DIFF
--- a/Sofa/framework/Type/test/VecTypes_test.cpp
+++ b/Sofa/framework/Type/test/VecTypes_test.cpp
@@ -23,6 +23,40 @@
 #include <sofa/defaulttype/VecTypes.h>
 #include <gtest/gtest.h>
 
+template < sofa::Size N, typename ValueType>
+void defaultConstructor()
+{
+    const sofa::type::Vec<N, ValueType> vec;
+    for (const auto& value : vec)
+    {
+        EXPECT_EQ(value, ValueType());
+    }
+}
+
+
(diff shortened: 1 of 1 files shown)


Sofa/framework/Type/test/VecTypes_test.cpp
```diff
@@ -23,6 +23,40 @@
 #include <sofa/defaulttype/VecTypes.h>
 #include <gtest/gtest.h>
 
+template < sofa::Size N, typename ValueType>
+void defaultConstructor()
+{
+    const sofa::type::Vec<N, ValueType> vec;
+    for (const auto& value : vec)
+    {
+        EXPECT_EQ(value, ValueType());
```
bakpaul, line 32: Is this creator defined even for non-object types like float and int ?   
And are we sure this throw zeros ? 

Using simple cast might do the job and be more explicit ?

alxbilger: This is the most generic way to instantiate a type with its default constructor. In case of primitive types, it initializes the value to zero.
Since I test only primitive types, a static_cast on zero can be used instead. I don't mind changing it if the reviewers prefer.

fredroy (approved):

fredroy (approved):

PR #5367, opened 2025-03-27
[examples] Fix wrong topology Data arrays on the surface mesh in HexahedralFEMForceField.scn 

hugtalbot:
Removing the visual model, not supporting topological changes (logically)

Noted by @CherryJGit in https://github.com/sofa-framework/sofa/discussions/5366

### DIFF
--- a/examples/Component/SolidMechanics/FEM/HexahedralFEMForceField.scn
+++ b/examples/Component/SolidMechanics/FEM/HexahedralFEMForceField.scn
@@ -28,17 +28,29 @@
         <CGLinearSolver iterations="25" name="linear solver" tolerance="1.0e-9" threshold="1.0e-9" />
         <MeshGmshLoader name="loader" filename="mesh/nine_hexa.msh" />
         <MechanicalObject src="@loader" name="Hexa" />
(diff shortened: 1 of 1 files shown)


epernod:
visual model is supporting topological changes otherwise we would never see it in SOFA scenes ^^

epernod (dismissed):
visual model is supporting topological changes otherwise we would never see it in SOFA scenes ^^

this is not the origin of the crash

epernod:
> ah, could you help us investigate @epernod ?

yes the main problem was the src="@../Container" in the `QuadSetTopologyContainer `and `TriangleSetTopologyContainer `which were overridden the topological buffers with info from the loader (volume mesh) and thus conflicting with the work of the TopologicalMapping which are computing the correct topological buffers (edges and triangles in this case) on the surface of the mesh only.  

I also remove the .xml file to explicitly put all the components and moved the visual node under the TriangleSetTopology

epernod:
Final result:
https://github.com/user-attachments/assets/02c52185-d0a4-4b3e-b4cb-d042b636c866

bakpaul (approved):

fredroy:
[ci-build][with-all-tests]

PR #5411, opened 2025-04-18
[SofaDistanceGrid] New scenes and regression tests for the plugin

th-skam:
In the SofaDistanceGrid plugin:

- updates to the old example scenes 
- duplicates and new scene to demonstrate distance grid collision models used with the free motion animation loop
- added regression tests. Reference files for regression are added in this PR: https://github.com/sofa-framework/Regression/pull/81.

To be merged after the Regression PR above.

[ci-depends-on https://github.com/sofa-framework/Regression/pull/81]

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #5426, opened 2025-04-29
[SofaDistanceGrid] Substitute pointer-sharing and ref-counting mechanism of the DistanceGrid class with std smart pointers

th-skam:
**Should be reviewed and merged after https://github.com/sofa-framework/sofa/pull/5411 to include regression tests**

The DistanceGrid class is responsible for computing/loading a distance grid and sharing it with the other components of the plugin. Since more than one component may use the same grid, a static map is used to load each separate instance once and store it globally to share it whenever required:
https://github.com/sofa-framework/sofa/blob/f8fb655aebc38238fcf68c110c58b2ddc114a8ed/applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/DistanceGrid.cpp#L1453-L1457
It keeps count of the pointers it has shared and deletes itself when all of them are freed.
https://github.com/sofa-framework/sofa/blob/f8fb655aebc38238fcf68c110c58b2ddc114a8ed/applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/DistanceGrid.cpp#L112-L118

This PR reproduces the functionality with smart pointers. Less manual memory management.
[ci-depends-on https://github.com/sofa-framework/Regression/pull/81]

### DIFF
--- a/applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/DistanceGrid.cpp
+++ b/applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/DistanceGrid.cpp
@@ -81,7 +81,6 @@ int validateDim(int n)
 
 DistanceGrid::DistanceGrid(int nx, int ny, int nz, Coord pmin, Coord pmax)
     : meshPts()
-    , m_nbRef(1)
     , m_nx(validateDim(nx)), m_ny(validateDim(ny)), m_nz(validateDim(nz))
     , m_nxny(m_nx*m_ny), m_nxnynz(m_nx*m_ny*m_nz)
     , m_dists(m_nx*m_ny*m_nz)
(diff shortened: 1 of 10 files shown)


applications/plugins/SofaDistanceGrid/src/SofaDistanceGrid/DistanceGrid.cpp
```diff
@@ -52,6 +52,8 @@ namespace container
 namespace _distancegrid_
 {
 
+std::map<DistanceGrid::DistanceGridParams, std::weak_ptr<DistanceGrid> > DistanceGrid::instances;
```
fredroy, line 55: I am a bit concerned about this:
- in a multi-threading context, there could be a race condition I think ?
- as it is created globally, once the plugin is loaded, `instances` will be in the process memory and it wont allow safe unloading (but unloading is not already safe anyway 😅)


But if it was already here, nevermind 🫠

fredroy:
[ci-build][with-all-tests]

bakpaul (approved):

bakpaul:
[ci-build][with-all-tests]

PR #5427, opened 2025-05-02
[Type] Mat: Fix unsafe methods (constructor and assignement) from ptr

fredroy:
MSVC was throwing an exception for `Mat(const real2* p)` about some offsets out of bound for iterator.
This PR fixes it in a way and even add a slight more safe version when real and real2 does not have the same size (which was not the case before).
This PR adds also some simple unit tests

But in the end, those two methods are very unsafe as it does not check the size of allocated memory pointed by the ptr so you can do a lot of bad things with it 👮


fredroy:
[ci-build][with-all-tests]

Sofa/framework/Type/test/MatTypes_test.cpp
```diff
@@ -376,3 +376,41 @@ TEST(MatTypesTest, conversionToReal)
...
+    
+    EXPECT_TRUE(comp);
+
+    sofa::type::Mat<3, 3, double> mat3d(arrayMat3f);
+
+    bool comp2 = true;
+    constexpr double epsilon = 0.00001;
+    for (std::size_t i = 0; i < 3; i++)
+        for (std::size_t j = 0; j < 3; j++)
+            comp = (std::fabs(arrayMat3f[i * 3 + j] - mat3d[i][j]) < epsilon) && comp2;
```
alxbilger:
```suggestion
            comp2 = (std::fabs(arrayMat3f[i * 3 + j] - mat3d[i][j]) < epsilon) && comp2;
```
But I recommend 2 different tests. Or at least, two different scopes

alxbilger (changes requested):

alxbilger (approved):

bakpaul (approved):

hugtalbot (approved):

PR #5441, opened 2025-05-12
[Framework] Apply template method design pattern to BaseVisualStyle

lbinria:
Apply template method design pattern to BaseVisualStyle


lbinria:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/visual/BaseVisualStyle.h
```diff
@@ -41,9 +41,35 @@ class SOFA_CORE_API BaseVisualStyle : public sofa::core::objectmodel::BaseCompon
     BaseVisualStyle() { }
     ~BaseVisualStyle() override { }
 
+    virtual void doUpdateVisualFlags(VisualParams* ) { };
+    virtual void doApplyBackupFlags(VisualParams* ) { };
```
hugtalbot, line 45: why not pure?

lbinria: I don't remember... but it doesn't seem like a problem to modify them to pure virtual

bakpaul: It is because the original implementation didn't use a pure virtual method, allowing for child class not to implement it. I would leave it as is personally.

PR #5489, opened 2025-05-15
[Container.Dynamic] Fix warnings in TriangleSetGeometryAlgorithms

hugtalbot:
This originally simple-minded PR fails due to inconsistency between sofa::helper::vector and Vec (pushback)
```cpp
sofa::type::vector< sofa::type::vector< SReal > > p_baryCoefs;
...
p_baryCoefs.push_back(a_baryCoefs);
```

where :
```
auto a_baryCoefs =
        computeTriangleBarycoefs(ind_ta, (const sofa::type::Vec<3, Real> &) a);
```
should become:
```
auto a_baryCoefs =
    computeTriangleBarycentricCoordinates(ind_ta, (const sofa::type::Vec<3, Real> &) a);
```


epernod (changes requested):
The 2 methods are not implemented the same way and might not always give the same results... This was one problem I encountered. 
This is why I already put those methods as Deprecated in v25.06

If those warnings are not too armfull, could we postpone that PR for after I finish my work on refactoring TriangleSetGeometryAlgorithms: https://github.com/epernod/sofa/pull/17

PR #5530, opened 2025-06-10
[Visual] VisualStyle: Set flags at init()

fredroy:
Workaround, for visual flags to be available for operations done before initVisual().
E.g fix the case where the visual flags are checked in the computeBBox() function (which is called before initVisual at the startup) like in the case of the FallingSofa.scn scene

Not a huge fan of calling the singleton, and the fact that we are doing visual-related stuff into a non-visual related context 🤔


fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):

PR #5588, opened 2025-07-02
[Visual] Introduce component to visualize a point cloud

alxbilger:
The idea is to extract the rendering part from MechanicalObject into a dedicated component. With a new component, this feature will be available outside of a MechanicalObject. The class MechanicalObject can then be cleaned for better maintenance. 

image


Sofa/Component/Visual/src/sofa/component/visual/VisualPointCloud.inl
```diff
@@ -0,0 +1,186 @@
...
+
+            float defaultRadius = 1.f;
+            if (!radius.empty())
+            {
+                defaultRadius = radius.back();
+            }
+            radius.resize(displayedPoints.size(), defaultRadius);
+
+            drawTool->setLightingEnabled(true);
+            drawTool->drawSpheres(displayedPoints, radius.ref(), color);
```
fredroy: I really advise not to use drawSphere if you are drawing a lot of them.
You had better use drawFakeSpheres() which implements "Sphere Impostors" https://bgolus.medium.com/rendering-a-sphere-on-a-quad-13c92025570c

Example in `examples/Benchmark/Render/ManySpheres.scn` (after merging https://github.com/sofa-framework/sofa/pull/5589 😅)
There is a 10x speedup between "real" and impostor spheres

fredroy (changes requested):

Sofa/Component/Visual/src/sofa/component/visual/VisualPointCloud.inl
alxbilger: Good to know, I did not know that we had that in SOFA.

alxbilger: image

The fake spheres

alxbilger:
[ci-build][with-all-tests]

damienmarchal:
A must have ! 
Especially if there is addition in the GUI the possibility to show/hide by clicking on the object. 

In some experimental branches we had hidden part of the scene graph storing components interactively added through GUI interaction and remove when the interaction is done. This avoid to pollute the scene when the data vizualisation is just there for "debugging". Of course this is complementary to have it in the scene if it make sense to always visualize the field/points.  

This is important because currently there is two usage of this showXXXX on mechanical object. One is to actually show something that should always be visible in the scene and thus, having adding component in the scene for that make sense. 
The other is the use of showXXX for debugging/inspecting the scene from the GUI. In this use case, ones check/uncheck the showXXXX data from the data inspector. For this use requesting to add a component "manually" while making the scene is cumbersome and thus, a dedicated support for that is mandatory in the GUI (and probably not hard to do... just add at run time the component/connect it/remove).

fredroy (approved):

PR #5604, opened 2025-07-10
[Response] Clean init of CollisionResponse

hugtalbot:
These changes come from my will to warn a user when an collision response is set. The OptionsGroup design is not helping much. This PR proposes to :

- clean the init to properly warn the user (reponse data is not set, or wrongly set
- in the `create()` function of the CollisionResponse, the data response is set/selected only if valid
- add a `getItemNames()` in OptionsGroup to get the full list of possibly options

[ci-depends-on https://github.com/sofa-framework/SofaSphFluid/pull/17]

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #5605, opened 2025-07-10
 [Helper] Remove default value on OptionGroup 

hugtalbot:
Based on https://github.com/sofa-framework/sofa/pull/5604

It seems weird to me that the first entry is by default the selected one.
This was problematic in the case of CollisionReponse `response` data field :
- it gets populated at `parse` using all available response method
- if a wrong value is given by the user, nothing allowed us to identify it since the by-default selected response was the first one (AugmentedLagrangian) which is obviously valid !

Here is my hacky solution. To be discussed :)

Real diff : https://github.com/sofa-framework/sofa/pull/5605/commits/cce486762567e56271f363b991048e2bfbf9b97d

### DIFF
--- a/Sofa/framework/Helper/src/sofa/helper/OptionsGroup.cpp
+++ b/Sofa/framework/Helper/src/sofa/helper/OptionsGroup.cpp
@@ -31,7 +31,6 @@ class OptionsGroup;
 //////////////////////////////////////////////////////////////////////////////////////////////////////
 OptionsGroup::OptionsGroup() : textItems()
 {
-    selectedItem=0;
 }
 ///////////////////////////////////////
 OptionsGroup::OptionsGroup(const OptionsGroup & m_radiotrick) : textItems(m_radiotrick.textItems)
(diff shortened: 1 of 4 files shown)


Sofa/framework/Helper/src/sofa/helper/OptionsGroup.h
```diff
@@ -71,6 +71,9 @@ public :
     ///Set the name of the id-th item
     void setItemName( unsigned int id_item, const std::string& name );
 
+    ///Get the vector of names available
+    type::vector<std::string> getItemNames();
```
bakpaul, line 75: I don't see any usage of this newly introduced method.

Sofa/framework/Helper/src/sofa/helper/OptionsGroup.cpp
```diff
@@ -31,7 +31,6 @@ class OptionsGroup;
 //////////////////////////////////////////////////////////////////////////////////////////////////////
 OptionsGroup::OptionsGroup() : textItems()
 {
-    selectedItem=0;
```
bakpaul, line 34: This is certainly breaking. Even if not set here, `selectedItem` will have a default value. Might as well control it no ? Or add an internal mechanism that track if the optionGroup has been set at least once, to know if we can trust this attribute.

alxbilger: I agree. `selectedItem` will have a value if you define it or not. I don't see how it solves your problem. Could you imagine a fix in the `CollisionResponse` class instead?

PR #5628, opened 2025-07-21
[Mapping.Linear] BarycentricMappingTopologyContainerMapper: use array for barycentric coefficients

fredroy:
All the classes inheriting BarycentricMappingTopologyContainerMapper (EdgeSetTopologyContainer, etc) have to implement a method which returns a vector of barycentric coefficients.
And then it is used in various methods (apply(), etc) where a copy is done for *every element* , which means an alloc and dealloc everytime.
Obviously this causes a big performance loss.

This PR replace the creation of a vector with a fixed array, as we know in advance the size of the array (number of elements). So there would be no need of dynamic alloc/dealloc.
With this PR, the performance of Barycentric Mapping where the "input topology" is a TopologyContainer is closer to "static" topologies such as grids or MeshTopology.

This PR introduces also a scene to emphasize the performance of the BarycentricMapping.

Some numbers (with the raptor, the biggest OBJ file in the repo, 35k vertices, 70k triangles) with 5000 steps:
`topology container master: 5000 iterations done in 24.857 s ( 201.15 FPS)`
`topology container PR : 5000 iterations done in 13.1221 s ( 381.036 FPS)`
`sparsegrid: 5000 iterations done in 11.4556 s ( 436.467 FPS)`
`meshtopology: 5000 iterations done in 11.4605 s ( 436.28 FPS)`

And to illustrate even more the problem, I tried the same scene with the stanford buddha (315k vertices, 630k triangles)
`topology container master: 5000 iterations done in 172.78 s ( 28.9385 FPS)`
`topology container PR: 5000 iterations done in 67.423 s ( 74.1586 FPS)`
`sparsegrid: 5000 iterations done in 51.0929 s ( 97.8609 FPS)`

NOTE: 
 - no compat because I dont think people inherited this mapper thing
 - I changed the name because getBaryCoef did not sound serious to me 🫢

### DIFF
--- a/Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/BarycentricMappers/BarycentricMapperTopologyContainer.h
+++ b/Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/BarycentricMappers/BarycentricMapperTopologyContainer.h
@@ -148,7 +148,7 @@ protected:
     ~BarycentricMapperTopologyContainer() override = default;
 
     virtual type::vector<Element> getElements()=0;
-    virtual std::array<SReal, Element::NumberOfNodes> getBarycentricCoefficients(const Real* barycentricCoordinates)=0;
+    virtual std::array<Real, Element::NumberOfNodes> getBarycentricCoefficients(const Real* barycentricCoordinates)=0;
     
     virtual void computeBase(Mat3x3d& base, const typename In::VecCoord& in, const Element& element)=0;
     virtual void computeCenter(Vec3& center, const typename In::VecCoord& in, const Element& element)=0;
--- a/Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/BarycentricMappers/BarycentricMapperTopologyContainer.h
+++ b/Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/BarycentricMappers/BarycentricMapperTopologyContainer.h
@@ -148,7 +148,7 @@ protected:
     ~BarycentricMapperTopologyContainer() override = default;
 
     virtual type::vector<Element> getElements()=0;
-    virtual std::array<SReal, Element::NumberOfNodes> getBarycentricCoefficients(const Real* f)=0;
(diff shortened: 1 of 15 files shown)


fredroy:
[ci-build][with-all-tests]

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/BarycentricMappers/BarycentricMapperTopologyContainer.h
```diff
@@ -148,7 +148,8 @@ class BarycentricMapperTopologyContainer : public TopologyBarycentricMapper<In,O
     ~BarycentricMapperTopologyContainer() override = default;
 
     virtual type::vector<Element> getElements()=0;
-    virtual type::vector<SReal> getBaryCoef(const Real* f)=0;
+    virtual std::array<SReal, Element::NumberOfNodes> getBarycentricCoefficients(const Real* f)=0;
```
alxbilger:
```suggestion
    virtual std::array<Real, Element::NumberOfNodes> getBarycentricCoefficients(const Real* f)=0;
```
Is it ok to return `Real` and not `SReal` here?

alxbilger: Could you add a description of the `f` parameter? Or maybe a more explicit name?

alxbilger (commented):
Nice speedups 🎆

Sofa/Component/Mapping/Linear/src/sofa/component/mapping/linear/BarycentricMappers/BarycentricMapperTopologyContainer.h
fredroy: > Is it ok to return `Real` and not `SReal` here?

The main reason is that I took the same type as the getBaryCoef 😅.
I guess it would be better to set the same type everywhere

hugtalbot (approved):

bakpaul (approved):

PR #5633, opened 2025-07-23
[examples] Fix stability in RegularGridTopology.scn

hugtalbot:
The scene RegularGridTopology.scn became unstable since v17.12 (see commit : https://github.com/sofa-framework/sofa/commit/4889770fe900033aaaad1d305e51dec14e3a140f#diff-2816509667eddf8af975da31c0fd3995a996901b828702ee76ba5e2cb25c8016), it seems related to:
- a change of FF (spring to FEM)
- a change of max nb of iteration in CG

Therefore, I suggest in this PR to reach stability by just ensuring a small enough time step (and removing the Rayleigh damping)

[ci-depends-on https://github.com/sofa-framework/Regression/pull/84]

### DIFF
--- a/examples/Component/Topology/Container/Grid/RegularGridTopology.scn
+++ b/examples/Component/Topology/Container/Grid/RegularGridTopology.scn
@@ -1,5 +1,5 @@
 <!-- RegularGrid examples -->
-<Node name="root" dt="0.02">
+<Node name="root" dt="0.002">
     <RequiredPlugin name="Sofa.Component.Collision.Detection.Algorithm"/> <!-- Needed to use components [BVHNarrowPhase BruteForceBroadPhase CollisionPipeline] -->
     <RequiredPlugin name="Sofa.Component.Collision.Detection.Intersection"/> <!-- Needed to use components [DiscreteIntersection] -->
     <RequiredPlugin name="Sofa.Component.Collision.Geometry"/> <!-- Needed to use components [SphereCollisionModel] -->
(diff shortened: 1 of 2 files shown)


fredroy:
> The scene RegularGridTopology.scn became unstable since v17.12 (see commit : [4889770#diff-2816509667eddf8af975da31c0fd3995a996901b828702ee76ba5e2cb25c8016](https://github.com/sofa-framework/sofa/commit/4889770fe900033aaaad1d305e51dec14e3a140f#diff-2816509667eddf8af975da31c0fd3995a996901b828702ee76ba5e2cb25c8016)), it seems related to:
> 
>     * a change of FF (spring to FEM)
> 
>     * a change of max nb of iteration in CG
> 
> 
> Therefore, I suggest in this PR to reach stability by just ensuring a small enough time step (and removing the Rayleigh damping)
> 

1) There is no FEM in this scene, the springs have been put back in a later commit 
https://github.com/sofa-framework/sofa/commit/18e3ee53a8eb59014c87bad467bdf47ab3dad355
So the only difference is the CG parameters now (but even by applying the previous numbers. it still explodes)

2) I am not against the PR as the scene would not explode anymore 😅 but it seems it eludes the fact that something was working before and not working now. But maybe it was some side-effects of something else ....

bakpaul:
It is certainly a side effect from something else. When you change the fixations from projective to lagrangian constraints, the stability is not even here at the first timestep. The ForceField seems broken somehow. 

Another workaround I found was to add rayleigh damping to the exploding model. We can see that the coarse one uses some, it stabilize the model well. But it doesn't explain why the FF damping isn't enough.

fredroy (approved):

bakpaul (approved):

PR #5635, opened 2025-07-24
[Type] Vec: update equality methods

fredroy:
A more modern way and testing if it is not a floating point. If this is the case, no need to test the diff with a floating point threshold..

\+ tests


fredroy:

hugtalbot (approved):

Sofa/framework/Type/src/sofa/type/Vec.h
```diff
@@ -589,16 +589,28 @@ class Vec
 
     constexpr bool operator==(const Vec& b) const noexcept
     {
-        for (Size i=0; i<N; i++)
-            if ( fabs( (float)(this->elems[i] - b[i]) ) > EQUALITY_THRESHOLD ) return false;
-        return true;
+        if constexpr (std::is_floating_point_v<ValueType>)
+        {
+            return std::equal(this->elems.begin(), this->elems.end(), b.elems.begin(),
+                                   [](auto x, auto y) { return std::abs(x - y) < EQUALITY_THRESHOLD; });
```
bakpaul, line 595:
```suggestion
                                   [](auto x, auto y) { return std::abs(x - y) < std::numeric_limits<ValueType>::epsilon(); });
```
In my mind the EQUALITY_THRESHOLD should not be used for vector comparison. 

IMO should remove this custom defined value and always use numeric_limits, even for geometrical operation. We could make the change directly on the define and see if all tests are passing.

Sofa/framework/Type/src/sofa/type/Vec.h
```diff
@@ -589,16 +589,28 @@ class Vec
...
 
     constexpr bool operator!=(const Vec& b) const noexcept
     {
-        for (Size i=0; i<N; i++)
-            if ( fabs( (float)(this->elems[i] - b[i]) ) > EQUALITY_THRESHOLD ) return true;
-        return false;
+        if constexpr (std::is_floating_point_v<ValueType>)
+        {
+            return !std::equal(this->elems.begin(), this->elems.end(), b.elems.begin(),
+                                   [](auto x, auto y) { return std::abs(x - y) < EQUALITY_THRESHOLD; });
```
bakpaul, line 608:
```suggestion
                                   [](auto x, auto y) { return std::abs(x - y) < std::numeric_limits<ValueType>::epsilon(); });
```

Sofa/framework/Type/src/sofa/type/Vec.h
fredroy: > In my mind the EQUALITY_THRESHOLD should not be used for vector comparison.
> 
> IMO should remove this custom defined value and always use numeric_limits, even for geometrical operation. We could make the change directly on the define and see if all tests are passing.

One would say that this is not the point of this PR 🤓  
but I can add your suggestions if people want anyway.

Nevertheless, you may want to look at that https://randomascii.wordpress.com/2012/02/25/comparing-floating-point-numbers-2012-edition/ 

Summary:
```
There is no silver bullet. You have to choose wisely.

    If you are comparing against zero, then relative epsilons and ULPs based comparisons are usually meaningless. You’ll need to use an absolute epsilon, whose value might be some small multiple of FLT_EPSILON and the inputs to your calculation. Maybe.
    If you are comparing against a non-zero number then relative epsilons or ULPs based comparisons are probably what you want. You’ll probably want some small multiple of FLT_EPSILON for your relative epsilon, or some small number of ULPs. An absolute epsilon could be used if you knew exactly what number you were comparing against.
    If you are comparing two arbitrary numbers that could be zero or non-zero then you need the kitchen sink. Good luck and God speed.
```

(and cited by https://google.github.io/googletest/reference/assertions.html by the way 🤓 )

bakpaul (approved):

PR #5648, opened 2025-08-06
[LinearSolver] Untangle linear solver and linear system

alxbilger:
In #2777, the matrix assembly was removed from the linear solver. Instead, the assembly is done in a dedicated component, a linear system. However, linear solvers still kept an API related to the matrix assembly. This PR removes this API for a clearer one, where both solvers and systems have their own responsibilities.

With this work, we will be able to assemble different systems while keeping the same linear solvers. For example, an augmented system containing also lagrangian constraints.

An interesting consequence is that a linear solver is now agnostic to the simulated physic: no mass factor, stiffness factor, damping factor. The assembly of the mass, damping and stiffness is now totally in the linear system. We can now imagine a linear system assembling different types of physics, while keeping the same linear solvers.

I took the opportunity to clean a few things in the preconditioners.

Additions:
- A new Data `authorizeAssembly` in `BaseMatrixLinearSystem`. This Data is hidden to the user. This Data exists to use the DDGNode system, in particular in the asynchronous solver.
- Similarly, a new Data `factorizationInvalidation` is introduced in `MatrixLinearSolver`. This Data is linked to the Data `matrixChanged`, also newly introduced, and serves as an observer for matrix changes. That way, the solver decompose the matrix only if a change in the matrix occurred.
- `CompositeLinearSystem` is now templated also on `GraphScatteredMatrix`.
- `PreconditionedMatrixFreeSystem` is a new component of the categories of the linear systems. It is intended to work with `PCGLinearSolver`. The specificity of this linear system is the link toward a the linear system of the preconditioner. It has a Data `assemblingRate`, allowing to assemble the matrix of the preconditioner only at a given rate. It is templated only on `GraphScatteredMatrix` since it works only with `PCGLinearSolver`.

Changes:
- Some types involving scalar types are changed from `double` to `Real`
- When there is no linear system in the scene, the linear solver instantiates automatically a linear system component. Before it was added as a slave, mainly for UX reasons. However, slaves are not visited by the visitors. Therefore, linear systems are no longer a slave when instantiated automatically.
- The Data `update_step` is removed in `PCGLinearSolver`. Since the solvers are no longer responsible to assemble the matrices, this Data is no longer used. Instead, the equivalent of this Data is now in `PreconditionedMatrixFreeSystem`.
- The assembly in the `WarpPreconditioner` is now transferred to `RotationMatrixSystem`. Similarly to `PreconditionedMatrixFreeSystem`, it has also a Data `assemblingRate`.
- The rotation finder is now an explicit link in `RotationMatrixSystem`. Before it was found implicitly with an index in `WarpPreconditioner`.

To do:

- [x] Fix warp preconditioner
- [x] Proper deprecation
- [ ] Review of the constraints, including the preconditioners
- [x] Fix `MechanicalMultiVectorPeqBaseVectorVisitor` in `MatrixLinearSolver`

[ci-depends-on https://github.com/sofa-framework/SofaPython3/pull/528]


alxbilger:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/behavior/LinearSolver.h
```diff
@@ -118,19 +96,18 @@ class SOFA_CORE_API LinearSolver : public BaseLinearSolver
     }
 
     /// Apply the contactforce dx = Minv * J^t * f and store the result in dx VecId
-    virtual void applyConstraintForce(const sofa::core::ConstraintParams* /*cparams*/,sofa::core::MultiVecDerivId /*dx*/, const linearalgebra::BaseVector* /*f*/) {
-        msg_error() << "applyConstraintForce has not been implemented.";
-    }
+    virtual void applyConstraintForce(const sofa::core::ConstraintParams* /*cparams*/,
```
bakpaul, line 131: I know, not in the scope of this PR, but why do we need this ? Isn't addMinvJt enough ?

bakpaul:
Ok so now the only purpose of a linear solver is to solve inverse linear problem given a LinearSystem that is responsible of defining the RHS and the LHS ? 

If the solver is a direct one, let's says a Cholesky solver, where is the decomposition supposed to be stored ? If the Solver is agnostic to the system, then it should not keep track of this decomposition no ? If in the future we would like to have many different LinearSystem using the same solver then they should somehow keep this info. Or maybe an intermediate map in the solver ?

alxbilger:
> Ok so now the only purpose of a linear solver is to solve inverse linear problem given a LinearSystem that is responsible of defining the RHS and the LHS ?
> 
> If the solver is a direct one, let's says a Cholesky solver, where is the decomposition supposed to be stored ? If the Solver is agnostic to the system, then it should not keep track of this decomposition no ? If in the future we would like to have many different LinearSystem using the same solver then they should somehow keep this info. Or maybe an intermediate map in the solver ?

@bakpaul The linear solver is still bound to a single linear system through its Link. It think it still makes sense to have the decomposition in the solver because of that. 

The intention with this PR was only to remove the assembly part in the API of the solver.

alxbilger:
[ci-build][with-all-tests]

alxbilger:
⚠️ The example scene `FEMBAR_PCG_PrecomputedWarpPreconditioner.scn` does not move, but it is also the case in master.

Sofa/framework/Core/src/sofa/core/behavior/BaseMatrixLinearSystem.cpp
```diff
@@ -27,15 +27,22 @@ namespace sofa::core::behavior
 BaseMatrixLinearSystem::BaseMatrixLinearSystem()
 : Inherit1()
 , d_matrixSize(initData(&d_matrixSize, "matrixSize", "Size of the global matrix"))
+, d_authorizeAssembly(initData(&d_authorizeAssembly, true, "authorizeAssembly", "Allows to assemble the system matrix"))
```
fredroy:
```suggestion
, d_authorizeAssembly(initData(&d_authorizeAssembly, true, "authorizeAssembly", "Enable assembly of the system matrix"))
```
and maybe renaming the Data with d_enableAssembly (`authorize` seems ... too strong ?)

fredroy (changes requested):

Sofa/framework/Core/src/sofa/core/behavior/BaseMatrixLinearSystem.cpp
fredroy: This applies to the other recurrences of this `d_authorizeAssembly` Data in other components

alxbilger: done

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/AsyncSparseLDLSolver.inl
```diff
@@ -29,44 +29,58 @@ namespace sofa::component::linearsolver::direct
 {
 
 template <class TMatrix, class TVector, class TThreadManager>
-void AsyncSparseLDLSolver<TMatrix, TVector, TThreadManager>::init()
+AsyncSparseLDLSolver<TMatrix, TVector, TThreadManager>::AsyncSparseLDLSolver()
+    : d_enableAssembly(initData(&d_enableAssembly, true, "authorizeAssembly", "Allow assembly of the linear system"))
```
fredroy:
```suggestion
    : d_enableAssembly(initData(&d_enableAssembly, true, "enableAssembly", "Allow assembly of the linear system"))
```

fredroy (changes requested):

fredroy (approved):

alxbilger:
Impossible, gmock is not available in SOFA

fredroy:

PR #5666, opened 2025-08-26
[Lagrangian] Refactoring of GenericConstraintSolver

bakpaul:
The constraint solver and constraint problems are too mixed up. IMO the constraint problem should only contain data describing the constraint problem (matrices, `ConstraintResolutions`, `ConstraintCorrections` maybe ?) but no intelligence, this is the solver that defines the algorithm. 

Moreover, having a single solver that implements multiple method using the same `ConstraintProblem` structure made it messy and difficult to make changes and even enrich. 

**Changes**

This PR does three things: 
1) Clean the component, removing unused or unnecessary data (small changes)
2) Introduce a inheritance mechanism to dispatch the different constraint solving methods in different types. Now the method used Is explicitly stated in the name. This has multiple positive effect:
   - Now the set of parameter of the Generic constraint solver will not contain data that can only be used for some subpart of it. Because each solving method is now a real `BaseObject`, it can enrich the set of data itself
   - No scene can use the `GenericConstraintSolver` anymore and must use either `ProjectedGaussSeidelConstraintSolver`, `NNCGConstraintSolver`, `UnbuiltGaussSeidelConstraintSolver` and more if any is added. This is way more explicit for the user.
   - Adding a new method is fairly easy and would only require to add a new inheritance, override the `doBuildSystem` (if it cannot use the ones directly available in built and unbuilt constraint solver) and `doSolve` methods 
3) We have a clear differentiation between built and unbuilt methods in the `ConstraintProblem`: it is separated in two set of attributes.  

A new type of solver 'DirectconstraintSolver' which are directly augmenting the system matrix to solve the constraint either in a simple direct way or in a Newton can be added in the future. 

Of course all scenes are broken with that. Well, actually, only python scenes, because xml ones are saved by SceneCheking mechanism given that the SolvingMethod was not specified. But anyway, it is pretty breaking. I tried my best to fix all the one in this repo along with the ones in the "full" configuration.

**Why this design ?**

At first my intend was to make the solver agnostic from the solving strategy by adding a new type of component which would be added to the scene along with t he solver and would define the solving method. 

My first commit shows it a bit,. I merged this new component with the ConstraintProblem (that again was a mixup between the structure and the inteligence, bu encapsulation is not always bad). Everything was good until I got the compilation error from the fact that we actually need a fixed_array of ConstraintProblem because of the asynchronous haptic loops, which made this design pretty weak. You can see how I've tried to tweak it to make it work but it wasn't satisfying. This need of having multiple instance of ConstraintProblem cried for the separation between the data and the solving method.  

Afterwards, to add to this, this previous attemp would make the `ConstraintSolver` hollow, taking away all the purpose of its component it was making it a useless real implementation of ConstraintSolverImpl. And why adding a new component that would only enrich the constraint solver when it could be the ConstraintSolver itself that does it ? 

[ci-depends-on https://github.com/sofa-framework/SofaPython3/pull/531]
[ci-depends-on https://github.com/sofa-framework/BeamAdapter/pull/200]
[ci-depends-on https://github.com/SofaDefrost/ModelOrderReduction/pull/168]
[ci-depends-on https://github.com/SofaDefrost/SoftRobots/pull/338]
[ci-depends-on https://github.com/SofaDefrost/STLIB/pull/134]
[ci-depends-on https://github.com/SofaDefrost/Cosserat/pull/158]
[ci-depends-on https://github.com/SofaDefrost/SoftRobots.Inverse/pull/73]


bakpaul:

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
```diff
@@ -429,36 +273,17 @@ bool GenericConstraintSolver::solveSystem(const core::ConstraintParams * /*cPara
     current_cp->allVerified = d_allVerified.getValue();
     current_cp->sor = d_sor.getValue();
 
-
-    // Resolution depending on the method selected
-    switch ( d_resolutionMethod.getValue())
+    if (notMuted())
```
damienmarchal, line 276: msg_info() is already something like if(printLog()){}
if you need a different condition you can use msg_info_when(my_condition)

Sofa/Component/AnimationLoop/src/sofa/component/animationloop/FreeMotionAnimationLoop.cpp
```diff
@@ -100,7 +102,7 @@ void FreeMotionAnimationLoop::init()
         l_constraintSolver.set(this->getContext()->get<sofa::core::behavior::ConstraintSolver>(core::objectmodel::BaseContext::SearchDown));
         if (!l_constraintSolver)
         {
-            if (const auto constraintSolver = sofa::core::objectmodel::New<DefaultConstraintSolver>())
+            if (const auto constraintSolver = sofa::core::objectmodel::New<constraint::lagrangian::solver::ProjectedGaussSeidelConstraintSolver>())
```
alxbilger:
```suggestion
            if (const auto constraintSolver = sofa::core::objectmodel::New<DefaultConstraintSolver>())
```
The idea was to use a single type alias so it is easy to change

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/BuiltConstraintSolver.h
```diff
@@ -0,0 +1,55 @@
...
+*                                                                             *
+* Contact information: contact@sofa-framework.org                             *
+******************************************************************************/
+#pragma once
+
+#include <sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.h>
+
+namespace sofa::component::constraint::lagrangian::solver
+{
+class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API BuiltConstraintSolver : public GenericConstraintSolver
```
alxbilger, line 34: Could you add a comment explaining what this class does and that it is used as a base class for other solvers?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/ConstraintSolverImpl.h
```diff
@@ -35,7 +35,7 @@ namespace sofa::component::constraint::lagrangian::solver
 {
 
 
-class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API ConstraintProblem
+class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API   ConstraintProblem
```
alxbilger:
```suggestion
class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API ConstraintProblem
```

Sofa/Component/AnimationLoop/src/sofa/component/animationloop/FreeMotionAnimationLoop.cpp
```diff
@@ -46,6 +46,8 @@
 #include <sofa/simulation/MainTaskSchedulerFactory.h>
 
 #include <sofa/simulation/mechanicalvisitor/MechanicalVInitVisitor.h>
+
+#include "sofa/component/constraint/lagrangian/solver/ProjectedGaussSeidelConstraintSolver.h"
```
alxbilger: Use <> instead

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/BuiltConstraintSolver.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include <sofa/component/constraint/lagrangian/solver/BuiltConstraintSolver.h>
+
+#include <sofa/helper/ScopedAdvancedTimer.h>
+#include <sofa/simulation/MainTaskSchedulerFactory.h>
+#include <sofa/simulation/ParallelForEach.h>
+
+namespace sofa::component::constraint::lagrangian::solver
+{
+
+    void BuiltConstraintSolver::doBuildSystem( const core::ConstraintParams *cParams, unsigned int numConstraints)
```
alxbilger: Usually in SOFA we don't indent the code inside namespaces

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/ProjectedGaussSeidelConstraintSolver.h
```diff
@@ -0,0 +1,39 @@
...
+namespace sofa::component::constraint::lagrangian::solver
+{
+class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API ProjectedGaussSeidelConstraintSolver : public BuiltConstraintSolver
+{
+public:
+    SOFA_CLASS(ProjectedGaussSeidelConstraintSolver, BuiltConstraintSolver);
+
+    virtual void doSolve( SReal timeout = 0.0) override;
+
+    void gaussSeidel_increment(bool measureError, SReal *dfree, SReal *force, SReal **w, SReal tol, SReal *d, int dim, bool& constraintsAreVerified, SReal& error, sofa::type::vector<SReal>& tabErrors) const;
```
alxbilger: does it need to be public?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/UnbuiltConstraintProblem.h
```diff
@@ -0,0 +1,47 @@
...
+class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API UnbuiltConstraintProblem : public GenericConstraintProblem
+{
+public:
+    typedef std::vector< core::behavior::BaseConstraintCorrection* > ConstraintCorrections;
+
+    UnbuiltConstraintProblem(GenericConstraintSolver* solver)
+    : GenericConstraintProblem(solver)
+    {}
+
+    linearalgebra::SparseMatrix<SReal> Wdiag; /** UNBUILT **/
```
alxbilger, line 46: Do those members need to be public?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/UnbuiltConstraintProblem.h
```diff
@@ -0,0 +1,47 @@
...
+#include <sofa/component/constraint/lagrangian/solver/config.h>
+
+#include <sofa/component/constraint/lagrangian/solver/GenericConstraintProblem.h>
+#include <sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.h>
+#include <sofa/linearalgebra/SparseMatrix.h>
+
+namespace sofa::component::constraint::lagrangian::solver
+{
+
+class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API UnbuiltConstraintProblem : public GenericConstraintProblem
```
alxbilger, line 36: Can you add a comment on this class?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
```diff
@@ -60,17 +62,14 @@ void clearMultiVecId(sofa::core::objectmodel::BaseContext* ctx, const sofa::core
 
 }
 
-static constexpr GenericConstraintSolver::ResolutionMethod defaultResolutionMethod("ProjectedGaussSeidel");
 
 GenericConstraintSolver::GenericConstraintSolver()
-    : d_resolutionMethod( initData(&d_resolutionMethod, defaultResolutionMethod, "resolutionMethod", ("Method used to solve the constraint problem\n" + ResolutionMethod::dataDescription()).c_str()))
-    , d_maxIt(initData(&d_maxIt, 1000, "maxIterations", "maximal number of iterations of the Gauss-Seidel algorithm"))
+    : d_maxIt(initData(&d_maxIt, 1000, "maxIterations", "maximal number of iterations of the Gauss-Seidel algorithm"))
```
alxbilger: I think the description is no longer accurate since you also use it in NNCG

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.h
```diff
@@ -53,29 +53,19 @@ class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API GenericConstraintSolver :
...
-
-    Data< ResolutionMethod > d_resolutionMethod; ///< Method used to solve the constraint problem, among: "ProjectedGaussSeidel", "UnbuiltGaussSeidel" or "for NonsmoothNonlinearConjugateGradient"
-
     Data<int> d_maxIt; ///< maximal number of iterations of the Gauss-Seidel algorithm
     Data<SReal> d_tolerance; ///< residual error threshold for termination of the Gauss-Seidel algorithm
     Data<SReal> d_sor; ///< Successive Over Relaxation parameter (0-2)
     Data< SReal > d_regularizationTerm; ///< add regularization*Id to W when solving for constraints
     Data<bool> d_scaleTolerance; ///< Scale the error tolerance with the number of constraints
     Data<bool> d_allVerified; ///< All constraints must be verified (each constraint's error < tolerance)
-    Data<int> d_newtonIterations; ///< Maximum iteration number of Newton (for the NonsmoothNonlinearConjugateGradient solver only)
```
alxbilger, line 78: I suggest to re-do a pass on all the Data to make sure that they are used by all the subclasses. For example, I think that `d_multithreading` should be available only for built solvers.

alxbilger (dismissed):
Overall I like the idea. Good job!

Is it relevent to add GenericConstraintCorrection into a component change list to help users to update the scenes?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.h
```diff
@@ -99,37 +88,22 @@ class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API GenericConstraintSolver :
     void clearConstraintProblemLocks();
 
     static constexpr auto CP_BUFFER_SIZE = 10;
-    sofa::type::fixed_array<GenericConstraintProblem, CP_BUFFER_SIZE> m_cpBuffer;
+    sofa::type::fixed_array<GenericConstraintProblem * , CP_BUFFER_SIZE> m_cpBuffer;
```
th-skam, line 90: `fixed_array` is a typedef for `std::array` (yes???)
The array type was changed to pointers and I'm pretty sure it won't automatically delete the managed object. 
There's a "new" allocation but no deletion in the .cpp file.

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
```diff
@@ -140,14 +140,16 @@ void GenericConstraintSolver::init()
...
+void GenericConstraintSolver::initializeConstraintProblems()
+{
+    for (unsigned i=0; i< CP_BUFFER_SIZE; ++i)
     {
-        static constexpr ResolutionMethod NonsmoothNonlinearConjugateGradient("NonsmoothNonlinearConjugateGradient");
-        if (d_resolutionMethod.getValue() != NonsmoothNonlinearConjugateGradient)
-        {
-            msg_warning() << "data \"newtonIterations\" is not only taken into account when using the NonsmoothNonlinearConjugateGradient solver";
-        }
+        m_cpBuffer[i] = new GenericConstraintProblem(this);
```
th-skam, line 147: This allocation will leak if not deleted later after the `m_cpBuffer` type was changed to `GenericConstraintProblem*`

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/ConstraintSolverImpl.h
th-skam, line 1: You could make `unsigned int getProblemId()` a const function on line 70
It doesn't touch members

th-skam (commented):
I like the logic behind the refactoring and I think it'll be helpful for any future additions.
I only left a few comments on your changes, hope they're helpful

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.h
```diff
@@ -99,37 +88,22 @@ class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API GenericConstraintSolver :
...
     sofa::type::fixed_array<bool, CP_BUFFER_SIZE> m_cpIsLocked;
     GenericConstraintProblem *current_cp, *last_cp;
 
     sofa::core::MultiVecDerivId m_lambdaId;
     sofa::core::MultiVecDerivId m_dxId;
 
-    void buildSystem_matrixFree(unsigned int numConstraints);
+    virtual void initializeConstraintProblems();
+    virtual void doBuildSystem( const core::ConstraintParams *cParams, unsigned int numConstraints) = 0;
+    virtual void doSolve( SReal timeout = 0.0) = 0;
```
alxbilger: I recommend to improve the API here. I think it will be difficult for a newcomer to know the input/output of this function. Maybe you can add them in the signature. What do you think?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/ProjectedGaussSeidelConstraintSolver.h
bakpaul: well, how it is done here no. But if the method was static (which could be the case if the conttraint problem was passed as ref) then yes.

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/UnbuiltConstraintProblem.h
bakpaul: This is really a container. So I don't mind having them public

Sofa/Component/AnimationLoop/src/sofa/component/animationloop/FreeMotionAnimationLoop.cpp
bakpaul: Sorry, IDE...

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.h
bakpaul: I've added a clean of those pointers.

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.h
bakpaul: What do you mean by adding them into the signature ? Would you like to add the constraintProblem to the signature ?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintProblem.h
```diff
@@ -33,54 +33,46 @@ class GenericConstraintSolver;
...
+    void setSolver(GenericConstraintSolver* solver);
 
-    int getNumConstraints();
-    int getNumConstraintGroups();
+    sofa::linearalgebra::FullVector<SReal> _d;
+    std::vector<core::behavior::ConstraintResolution*> constraintsResolutions;
+    bool scaleTolerance, allVerified;
+    SReal sor;
+    SReal currentError;
+    int currentIterations;
```
fredroy, line 67: if this PR breaks a lot, we may as well take the opportunity to rename those members ?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/ProjectedGaussSeidelConstraintSolver.h
damienmarchal: could it use the sofa naming convention

hugtalbot (approved):

PR #5675, opened 2025-08-28
[Type] Vec: Phase out ctor/assignment from pointers and Vec of different sizes/types

fredroy:
Why:
- from raw C-pointers: no way to know bounds so super prone to error/crash at runtime, and quite undetectable at compile-time (some recent compilers can catch those by the way)
- from other Vec: mostly to remove all this implicit calls. I guess it was done to not care about Vec1/Vec2/Vec3 while writing templated code but it is very shadowy and is also prone to errors. Even if the compiler can catch and raises some *warnings* at compile-time, we all know nobody read the warnings... (more on that later)

So this PR:
- deprecate the aforementioned ctor/assignments methods
- add one function to explicitly convert any Vec to an other Vec
- add one other utility function to explictly convert to Vec3 because this is the most used case for converting Vecs
- apply on the sofa repo codebase
- also tried to replace some copies with RVO assignments but I guess it depends of the compiler/optimization levels

EDIT: all functions now have 1 input and return the output (no ref involved), and we can specify the value to add in case the output vector is bigger. And units tests have been introduced


fredroy:

fredroy:

bakpaul (approved):

alxbilger (changes requested):
I am a bit annoyed that you have two types of conversion functions. One takes 1 parameter, and the other takes 2, while they have similar names.

My opinion is to change `toVecN` to accept only the input. The output is returned by the function. You would need to change the order of the template parameters (see `ToVecN2` in the benchmark).

Note that I benchmarked your implementation compared to another implementation without loops. The no-loop implementation is faster in `-o1` but has similar results in `-o3`. Since your implementation is clearer, I suggest to keep it.
https://quick-bench.com/q/W_qliO7wlKysiPO8n6YzaByqJU8

fredroy:
> I am a bit annoyed that you have two types of conversion functions. One takes 1 parameter, and the other takes 2, while they have similar names.

I knew you would react on this 😅 

> 
> My opinion is to change `toVecN` to accept only the input. The output is returned by the function. You would need to change the order of the template parameters (see `ToVecN2` in the benchmark).

I am just a bit annoyed (and thats why I did the 2 input versions) that we need to explicit the template while calling.
But in the other hand, maybe it is better to explicitly define what we want 🤔 
(and I did not see any ToVecN2 in the benches ❓ )

> 
> Note that I benchmarked your implementation compared to another implementation without loops. The no-loop implementation is faster in `-o1` but has similar results in `-o3`. Since your implementation is clearer, I suggest to keep it. https://quick-bench.com/q/W_qliO7wlKysiPO8n6YzaByqJU8

👍 

According to your bench, gcc is twice faster (with O3) with the current implementation so yes I guess we should keep it 😅 
https://quick-bench.com/q/y7v3jWuukS3gOtAEyQqvHE3HlTk

alxbilger:
@fredroy The bench I shared is not the one I wanted to share 😅 
Anyway, I approve like it is.

alxbilger (approved):

PR #5683, opened 2025-09-03
[CImgPlugin] Fix configuration when cmake is reloaded without the fetched dir

bakpaul:
A "random" CMake configuration failure was happening on the CI. This actually wasn't as random as felt: it only happened after a rebuild, without forcing full build. 

This is because when a rebuild without force full build is done, the CMakeCache stays but CI scripts delete the directory `${BUILD_DIR}/external_directories` for some reason. This lead to internal cached data (namely `CIMG_H_DIR` and `CIMG_PLUGINS_H_DIR`) pointing to deleted paths.  

This is an expected behavior, find_path caching those variable for future reloading. But we were in a special case where those paths were getting deleted in between... 

Now if the variables are cached I verify that they are still holding files as expected otherwise I unset them, forcing find_path to actually find the paths. 

### DIFF
--- a/applications/plugins/CImgPlugin/cmake/FindCImg.cmake
+++ b/applications/plugins/CImgPlugin/cmake/FindCImg.cmake
@@ -37,6 +37,13 @@ The following cache variables may also be set:
 
 cmake_minimum_required(VERSION 3.9)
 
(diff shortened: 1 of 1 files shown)


fredroy:
let's test the rebuild then...
[ci-build]

fredroy (approved):

PR #5692, opened 2025-09-12
[Framework] Remove shadow warnings

fredroy:
Trying to enable -Wshadow with gcc on the codebase triggered more than 50k warnings, and usually those warnings can be useful (https://www.youtube.com/watch?v=7_nSywhw_E8 ...)
So I tried to remove all warnings from the Framework (only there because it is boring and takes time) 

99% were harmless (ctors usually) but some were interesting in some cpp files like Node.
And i preferred renaming the parameters instead of the members in the ctor because I did not want to do breaking changes 🥸
Actually lots of code in the Framework does not follow the guideline of `m_` for members 👮


fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):

damienmarchal:
I passed the PR in wip because of linux failing on the CI. 
After a quick check I wonder if this is not a kind of out of memory: 
image

(CGAL consume a lot, and if un-luck suck all memory. If this is the case maybe we should use a reduced number of compilation thread for that part)

fredroy:
> I passed the PR in wip because of linux failing on the CI. After a quick check I wonder if this is not a kind of out of memory: image
> 
> (CGAL consume a lot, and if un-luck suck all memory. If this is the case maybe we should use a reduced number of compilation thread for that part)

This is happening radonmly to some other PRs, presumably because some of the VMs lack memory... so it is not really to this PR.

fredroy:
[ci-build][with-all-tests]

PR #5710, opened 2025-09-26
[BaseViewer] Change drawing default value and add a way to deactivate drawing

bakpaul:


fredroy (approved):

Sofa/GUI/Common/src/sofa/gui/common/BaseViewer.h
```diff
@@ -127,12 +127,13 @@ class SOFA_GUI_COMMON_API BaseViewer
...
     bool m_showSelectedNodeBoundingBox {true};
     bool m_showSelectedObjectBoundingBox {true};
-    bool m_showSelectedObjectPositions {true};
-    bool m_showSelectedObjectSurfaces {true};
-    bool m_showSelectedObjectVolumes {true};
-    bool m_showSelectedObjectIndices {true};
+    bool m_showSelectedObjectPositions {false};
+    bool m_showSelectedObjectSurfaces {false};
+    bool m_showSelectedObjectVolumes {false};
+    bool m_showSelectedObjectIndices {false};
```
hugtalbot, line 136: You introduce a global boolean m_enableSelectionDraw. Why exactly ?

You set back to false all topology related bool, could you explain?

bakpaul: To be able to disable selection drawing as simply as that. You don't always want to change drawing with the risk of having some crash when only going through the graph.

The second part is because when everything is drawn at the same time it is very messy. I prefer that the default behavior is only showing the bounding box. It is debatable, I don't care which one should be default, but I believe that only one should be showed at the same time.

hugtalbot (approved):

PR #5252, opened 2025-01-31
[Pixi][CI] Add pixi support

olivier-roussel:
Adds Pixi support for developers and CI.
Enables to:
- Have a standard way for a new developer to configure/build the project (homogeneous with other projects) and available on all platforms
- All dependencies stored in one file
- Have an easy way to enable some optional features
- Have a lock file for a reproducible build in CI

TODO:

- [x] base pixi.toml file with build scopes
- [x] features based on CMake presets (see #5272)
- [x] ~~lockfile update bot (require Github app https://github.com/apps/update-pixi-lockfile )~~ To be done in a dedicated PR
- [x] pixi-based CI for linux-64 / win-64 / ~~osx-64~~ / osx-arm64
- [x] ~~pixi-based CI for linux-arm64~~ Delayed after SOFA-GLFW / Imgui conda packaging

More info on pixi: https://pixi.sh/
______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #5720, opened 2025-10-02
[Helper] Add Taskflow dependency to CMake build system

alxbilger:
Integrate Taskflow as a header-only library, update CMake configuration to locate and fetch Taskflow if necessary, and include relevant headers.

taskflow: https://github.com/taskflow/taskflow

More information:
- https://taskflow.github.io/
- https://www.youtube.com/watch?v=MX15huP5DsM


damienmarchal:
It remind me so much something we did with PJBensoussan around 2006 in SOFA.

cmake/Modules/Findtaskflow.cmake
```diff
@@ -0,0 +1,66 @@
...
+
+    # Set the output variable
+    set(Taskflow_VERSION "${formatted_version}")
+elseif(SOFA_ALLOW_FETCH_DEPENDENCIES)
+    message("taskflow NOT FOUND. SOFA_ALLOW_FETCH_DEPENDENCIES is ON, fetching taskflow...")
+
+    # If Taskflow is not found, use FetchContent to download and build it
+    include(FetchContent)
+
+    FetchContent_Declare(
```
bakpaul: I find it strange to have this in the module. This is not standard. 

Plus I've added a nex macro called 'sofa_fetch_dependency' that does all of this. So really you could do as simple as the following in the cmakelists and remove this from the module

```cmake
sofa_find_package(taskflow)
if(NOT Taskflow_FOUND AND SOFA_ALLOW_FETCH_DEPENDENCIES)
    message("Sofa.Helper: DEPENDENCY taskflow NOT FOUND. SOFA_ALLOW_FETCH_DEPENDENCIES is ON, fetching taskflow...")

    sofa_fetch_dependency(taskflow
            GIT_REPOSITORY https://github.com/taskflow/taskflow.git
            GIT_TAG        v3.10.0
            DONT_BUILD
    )

    set(TASKFLOW_ROOT ${taskflow_SOURCE_DIR})
    sofa_find_package(taskflow REQUIRED)
elseif (NOT CImg_FOUND)
    message(FATAL_ERROR "Sofa.Helper: DEPENDENCY taskflow NOT FOUND. SOFA_ALLOW_FETCH_DEPENDENCIES is OFF and thus cannot be fetched. Install taskflow, or enable SOFA_ALLOW_FETCH_DEPENDENCIES to fix this issue.")
endif()

```

alxbilger: thanks, that's exactly what I wanted when I asked for a review 👍

alxbilger:

bakpaul (approved):

cmake/Modules/FindTaskflow.cmake
```diff
@@ -0,0 +1,51 @@
+# Set Taskflow_FOUND to false initially
+set(Taskflow_FOUND FALSE)
+
+set(Taskflow_VERSION "3.10.0")
```
alxbilger:
```suggestion
```

olivier-roussel: I was about to ask you why there was this line. Agree that it should be removed

olivier-roussel:
Modern cmake usage is more a suggestion here (my commits 05691e0 and cec0c7a ). Both solutions work fine and looks good to me. It was mainly driven by the idea of being consistent with cmake config files provided by the Taskflow library itself.
Don't know if anybody has an opinion about that... I can revert these commits if necessary

alxbilger:
About https://github.com/sofa-framework/sofa/commit/cec0c7a1752a5e97af16e1e061f2b05b3152b6a6, it is was a conclusion in the last dev meeting. So, thank you for having done the job!

Another question: is there something to do regarding the package/installation process?

olivier-roussel:
The only thing I think is still missing here is the find package for the generated cmake config file of Sofa.Helper, i.e. adding in the Sofa.HelperConfig.cmake.in :
```suggestion
if(NOT TARGET Taskflow::Taskflow)
    sofa_find_package(Taskflow QUIET REQUIRED)
endif()
```
~~But strangely, find_package calls for other required libs (Json, STB, DiffLib) are missing as well here. So either it is an omission, either I am wrong and missing something...~~
Ok my bad: Json, STB and DiffLib are not using targets so they do not need to be explicitly found here. Everything is OK for them. We just need this find_pacakge in the  Sofa.HelperConfig.cmake.in if we use the modern cmake version.

olivier-roussel:
[ci-build][with-all-tests]

olivier-roussel:
[ci-build][with-all-tests]

olivier-roussel:

olivier-roussel:
Just for information, there would/will be some extra steps necessary after looking for the Taskflow package in the Sofa.Helper CMakeLists.txt to install the library (headers + cmake config files) in the case of building the binary package on Windows. But this when we'll have migrated our Windows support on Pixi and it's purely related to it, so I'll do it in #5252.

olivier-roussel (approved):

PR #5722, opened 2025-10-08
[BuiltConstraintSolver] Add SVD based regularization strategy

bakpaul:
Based on https://github.com/sofa-framework/sofa/pull/5666

This PR adds a regularization strategy that uses SVD to explore the NullSpace of the matrix to only add regularization to constraints that have antagonists constraints. 

meaningful commit: https://github.com/sofa-framework/sofa/pull/5722/commits/30f81e4766e746ab09093a0ff63cdffac4ffac4d

### Introduced example
The added example shows a beam attached at both sides and which center is constrained in two different positions, meaning that two antagonists constraints are imposed to the center. The idea is that, because the fixations are not faulty, then we want them to be fully solved and we don't want to perturb them while still releasing the constraints on the antagonists ones. 

#### With SVD
Capture d’écran du 2025-10-09 11-45-22
We can see that the attachments are fully solved why in the middle a compromise is found

#### Without SVD
Capture d’écran du 2025-10-09 11-46-00
Every constraints is regularized so even attachment are not respected.


Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/BuiltConstraintSolver.cpp
```diff
@@ -41,6 +43,82 @@ void BuiltConstraintSolver::init()
...
 }
 
+void BuiltConstraintSolver::addRegularization(linearalgebra::BaseMatrix& W, const SReal regularization)
+{
+
+    if (d_useSVDForRegularization.getValue())
+    {
+        if (regularization>std::numeric_limits<SReal>::epsilon())
+        {
+            auto * FullW =  dynamic_cast<sofa::linearalgebra::LPtrFullMatrix<SReal> * >(&W);
```
th-skam, line 56: I see that this cast is ensured by how the `ComplianceWrapper` sets the matrix type in `BuiltConstraintSolver`. Still, you could add a check for nullptr if anything changes in the future.

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/BuiltConstraintSolver.cpp
```diff
@@ -41,6 +43,82 @@ void BuiltConstraintSolver::init()
...
+            Eigen::JacobiSVD<Eigen::MatrixXd> svd( EigenW, Eigen::ComputeFullV | Eigen::ComputeFullU );
+
+
+
+            //Given the SVD, loop over all singular values, and those that are smaller than 1% of the highest one are considered to be the null space
+            std::vector<bool> nullSpaceIndicator(problemSize, false);
+            int nullSpaceBegin = -1;
+            for(size_t i=0; i<problemSize; i++)
+            {
+                if (fabs(svd.singularValues()(i)) < fabs(svd.singularValues()(0))/100.0)
```
th-skam: Might be good to add the 1% criterion as a parameter so that the user can limit/expand the filter.
In a more complex problem, wouldn't it be possible that constraint feasibility is not so clearly separated and you end up with over-regularization. Or on the contrary being too strict with the criterion?

th-skam (commented):
A few comments and suggestions for this very useful PR

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/BuiltConstraintSolver.cpp
bakpaul: Will do !

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/BuiltConstraintSolver.cpp
bakpaul: Yes, it is highly probable. But in that case, if no singular value is smaller than the rest, then all the problem should be regularized in my mind. 

On the parameter side yes, I also wanted to do that. But I dislike adding new parameters that only apply if another parameter is true. LEt's discuss this in the sofa-dev-meeting

fredroy:
[ci-build][with-all-tests]

bakpaul:
[ci-build][with-all-tests]

th-skam (approved):

PR #5726, opened 2025-10-10
[Config] Prevent recursive calls to the same directory

alxbilger:
If ${dir} hasn’t had any add_subdirectory() calls yet, SUBDIRECTORIES will simply contain ${dir} itself

The infinite loop happens, on Windows, when `SOFA_USE_DEPENDENCY_PACK` is enabled while the files are already in the source directory.

Note that this change fixes the infinite loop, but CMake still fails after. The solution is then to disable `SOFA_USE_DEPENDENCY_PACK`.


fredroy (approved):
got the same problem 👍

Issue #5733, opened 2025-10-14
Incorrect eigen decomposition in the Ogden material model


th-skam:
### Problem

Lately, I've been working on the Ogden material model which fails the unit tests in `Sofa/Component/SolidMechanics/FEM/HyperElastic/tests/Material_test.cpp`. 

This model uses the `SelfAdjointEigenSolver` from the Eigen lib to compute eigenpairs of a matrix. The eigenvalues are computed correctly but the eigenvectors it produces are wrong. To progress, I switched to the generic `EigenSolver` which does the job although it is not the optimal choice for symmetric matrices. I wanted to document this in case someone else comes across it. Maybe it gets fixed with Eigen 5.0 (I am currently using Eigen-3.4.0).

### An example
In `MaterialTest_cpp`, a random tensor is generated, then adjusted to become positive-definite. On my machine, the random matrix is:
```math
C = \begin{bmatrix} 
6.03838452029702 & -0.500804028823228 & 1.0692913506807 \\ 
-0.500804028823228 & 7.01386290516793 & 2.22215388836855  \\
1.0692913506807 & 2.22215388836855 & 1.53182086321567 
\end{bmatrix}
```

For this matrix, the eigenpairs ($\lambda, \vec{v}$) should satisfy the equation $\lvert C \vec{v} - \lambda\vec{v} \rvert = 0$. The `SelfAdjointEigenSolver` fails the check.

| `SelfAdjointEigenSolver` | `EigenSolver` |
| - | - |   
| $\lvert C \vec{v_1} - \lambda_1\vec{v_1} \rvert \approx 4.1$ | $\lvert C \vec{v_1} - \lambda_1\vec{v_1} \rvert < 1e-15$ |
| $\lvert C \vec{v_2} - \lambda_2\vec{v_2} \rvert \approx 0.7$ | $\lvert C \vec{v_2} - \lambda_2\vec{v_2} \rvert < 2e-15$ |
| $\lvert C \vec{v_3} - \lambda_3\vec{v_3} \rvert \approx 7.4$ | $\lvert C \vec{v_3} - \lambda_3\vec{v_3} \rvert < 1e-15$ |

## Code to reproduce this
    Eigen::Matrix<double,3,3> CEigen;
    CEigen(0,0) = 6.03838452029702; 
    CEigen(0,1) = -0.500804028823228; 
    CEigen(1,0) = -0.500804028823228; 
    CEigen(1,1) = 7.01386290516793;
[...]

PR #5748, opened 2025-10-24
[HyperElastic] Add a working version of Ogden's material law

th-skam:
The material model in `Ogden.h` was failing the `Material_test.cpp` unit tests. This PR adds a working version that passes the tests.

[ci-depends-on https://github.com/sofa-framework/Regression/pull/87]

## TL;DR
- [X] Add unit tests for Ogden's model
- [X] Include the model in the examples scene with beams
- [X] Disable the `SelfAdjointEigenSolver` and use `EigenSolver` for computing eigenvectors (see #5733 ); this might get fixed with eigen 5 and then we should get back to the `SelfAdjointEigenSolver` (more appropriate).
- [ ] `applyElasticityTensor` computes the elasticity tensor and multiplies a vector; a 'smarter' computation is needed.
- [X] Add some documentation for the model (starting from this PR)

The new model runs slower than before. I kept the old functions for now since they are faster (although incorrect). 
To get the model working, I had to reimplement the functions related to the elasticity tensor, as I couldn’t derive the mathematical expressions backwards from the old implementation. However, I might be able to do this with a working version in hand.

### What the model functions provide:

The strain energy: 
```math
W = F(J) \frac{\mu}{\alpha^2} \sum_i^3 \left[\lambda^{ \frac{\alpha}{2} } \right] 
- 3 \frac{\mu}{\alpha^2} 
+ k_0 \ln^2 (J)
```

where $F(J) = J^{ - \frac{\alpha}{3} }$ with $J = \sqrt{det(C)}$ and $C$ is the Cauchy deformation tensor. Also, $\alpha$ and $\mu$ are variables of the model for fitting. $\lambda$ are the eigenvalues of $C$. This expression is different than many of the formulas I found in literature for Ogden's model. Here, $\alpha$ is squared. It can be found in ABAQUS' domumentation: https://classes.engineering.wustl.edu/2009/spring/mase5513/abaqus/docs/v6.6/books/usb/default.htm?startat=pt05ch17s05abm07.html

second Piola–Kirchhoff stress tensor: 
```math
S = 2 \frac{\partial W}{\partial C} = F(J) \frac{\mu}{\alpha} 
\left[ C^{ \frac{\alpha}{2} - 1} 
- \frac{1}{3} tr\left( C^{\frac{\alpha}{2}} \right) C^{-1} \right] 
+ k_0 \ln (J) C^{-1}
```

Elasticity tensor: 
```math 
C = 4 \frac{\partial^2 W}{\partial C \partial C} 
= 2 F(J) \frac{\mu}{\alpha} 

\left[ \left( \frac{\alpha}{2}  - 1 \right)  

\sum_i^3 \left( \lambda_i^{\frac{\alpha}{2} -2} P_i \otimes P_i \right)

+ \sum_{i\neq j}^3 \left(\frac{ \lambda_i^{\frac{\alpha}{2} -1} - \lambda_j^{\frac{\alpha}{2} -1} } {\lambda_i - \lambda_j} \left( P_i \otimes P_j \right)^{sym}\right)

-\frac{\alpha}{6} C^{\frac{\alpha}{2} - 1} \otimes C^{-1}
+ \frac{\alpha}{3} tr\left( C^{\frac{\alpha}{2}} \right) \left( C^{-1} \otimes C^{-1} \right)^{sym}

-\frac{\alpha}{6} C^{-1} \otimes C^{\frac{\alpha}{2} - 1}
+ \frac{\alpha}{6} \frac{1}{3} tr\left(C^{\frac{\alpha}{2}} \right) \left( C^{-1} \otimes C^{-1} \right)
\right]
+ k_0 \left( C^{-1} \otimes C^{-1} \right) - 2 k_0 ln(J) \left( C^{-1} \otimes C^{-1} \right)^{sym}
```

where $P_i = \vec{N_i} \otimes \vec{N_i}$ and $\vec{N}_i$ are eigenvectors of $C$.


th-skam:
[ci-build][with-all-tests]

fredroy:

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+        const Real mu1 = param.parameterArray[0];
+        const Real alpha1 = param.parameterArray[1];
+        const Real k0 = param.parameterArray[2];
+        const Real fj = pow(sinfo->J, -alpha1/3.);
+        const Real logJSqr = pow(log(sinfo->J), 2.);
+
+        // Solve eigen problem for C
+        Eigen::Matrix<Real, 3, 3> CEigen;
+        for (sofa::Index i = 0; i < 3; ++i)
+            for (sofa::Index j = 0; j < 3; ++j) CEigen(i, j) = C[vId(i, j)];
```
bakpaul:
```suggestion
            for (sofa::Index j = 0; j < 3; ++j) 
                CEigen(i, j) = C[vId(i, j)];
```

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+
+        // Disable temporarilly until fixed /*Eigen::SelfAdjointEigenSolver<EigenMatrix>*/
```
bakpaul: Can you explain why this method is prefered and why it is broken ??

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+        const MatrixSym& C=sinfo->deformationTensor;
+        const Real mu1 = param.parameterArray[0];
+        const Real alpha1 = param.parameterArray[1];
+        const Real k0 = param.parameterArray[2];
+        const Real fj = pow(sinfo->J, -alpha1/3.0);
 
+        // Solve eigen problem for C
+        Eigen::Matrix<Real, 3, 3> CEigen;
+        for (sofa::Index i = 0; i < 3; ++i)
+            for (sofa::Index j = 0; j < 3; ++j) CEigen(i, j) = C[vId(i, j)];
```
bakpaul:
```suggestion
            for (sofa::Index j = 0; j < 3; ++j) 
                CEigen(i, j) = C[vId(i, j)];
```

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -48,6 +48,20 @@ the determinant of the deformation gradient J and the right Cauchy Green deforma
...
+    const std::array<std::tuple<sofa::Index, sofa::Index>, 6> fromVoigt =
+    {
+        std::make_tuple(0,0),
+        std::make_tuple(0,1),
+        std::make_tuple(1,1),
+        std::make_tuple(0,2),
+        std::make_tuple(1,2),
+        std::make_tuple(2,2)
+    };
+
```
bakpaul: Could this be added to the MatrixSym API somehow ?

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
-                +k0/(Real)2.0*_trHC*inversematrix-(Real)(k0*log(sinfo->J))*Firstmatrix;
+        // Disable temporarilly until fixed /*Eigen::SelfAdjointEigenSolver<EigenMatrix>*/
+        Eigen::EigenSolver<Eigen::Matrix<Real, 3, 3> > EigenProblemSolver(CEigen, true);
+        if (EigenProblemSolver.info() != Eigen::Success)
+        {
+            dmsg_warning("Ogden") << "SelfAdjointEigenSolver iterations failed to converge";
+            return;
+        }
+        const EigenMatrix Evect = EigenProblemSolver.eigenvectors().real(); // orthonormal eigenvectors
+        const CoordEigen Evalue = EigenProblemSolver.eigenvalues().real();
```
bakpaul: Could this be saved once per timestep because if C stays the same then this computation is unecessary. This might be a micro optim, but it might be computed more than once per timstep.

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+
+        // trace of C^(alpha1/2)
+        const Real aBy2 = alpha1*0.5;
+        const Real trCaBy2 = pow(Evalue[0], aBy2) + pow(Evalue[1], aBy2) + pow(Evalue[2], aBy2);
+
+        // Transpose (also inverse) of the eigenvector matrix
+        Matrix3 EigenBasis;
+        for (auto m = 0; m < Evect.rows(); ++m)
+            for (auto n = 0; n < Evect.cols(); ++n) EigenBasis(m, n) = Evect(m, n);
```
bakpaul:
```suggestion
            for (auto n = 0; n < Evect.cols(); ++n) 
                EigenBasis(m, n) = Evect(m, n);
```

bakpaul: By the way why do you need explicitly this matrix in transpose form ? Wouldn't be enought to use the 'transposed()' method ? It might be more optimized thant copying the values

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+        
+        // Construct C^(alpha1/2 - 1) from eigenbasis: V * D * V^T; D_i = lambda_i^(alpha1/2 - 1)
+        const Real aBy2Minus1 = aBy2 - 1.;
+        const MatrixSym D = MatrixSym(pow(Evalue[0], aBy2Minus1), 0, pow(Evalue[1], aBy2Minus1), 0, 0, pow(Evalue[2], aBy2Minus1));
+        const Matrix3 Ca = EigenBasis*D.SymMatMultiply(EigenBasis.transposed());
```
bakpaul: If you keep the explicit transposed matrix, please replace the `EigenBasis.transposed()` by `Evect`

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+        const MatrixSym& C = sinfo->deformationTensor;
+        const Real mu1 = param.parameterArray[0];
+        const Real alpha1 = param.parameterArray[1];
+        const Real k0 = param.parameterArray[2];
+        const Real fj = pow(sinfo->J, -alpha1/3.0);
+
+        // Solve eigen problem for C
+        Eigen::Matrix<Real, 3, 3> CEigen;
+        for (sofa::Index i = 0; i < 3; ++i)
+            for (sofa::Index j = 0; j < 3; ++j) CEigen(i, j) = C[vId(i, j)];
```
bakpaul:
```suggestion
            for (sofa::Index j = 0; j < 3; ++j) 
                CEigen(i, j) = C[vId(i, j)];
```

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+
+        // Disable temporarilly until fixed /*Eigen::SelfAdjointEigenSolver<EigenMatrix>*/
+        Eigen::EigenSolver<Eigen::Matrix<Real, 3, 3> > EigenProblemSolver(CEigen, true);
+        if (EigenProblemSolver.info() != Eigen::Success)
+        {
+            dmsg_warning("Ogden") << "SelfAdjointEigenSolver iterations failed to converge";
+            return;
+        }
+        const EigenMatrix Evect = EigenProblemSolver.eigenvectors().real();
+        const CoordEigen Evalue = EigenProblemSolver.eigenvalues().real();
```
bakpaul: Same as before

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+
+        // trace of C^(alpha1/2)
+        const Real aBy2 = alpha1*0.5;
+        const Real trCaBy2 = pow(Evalue[0], aBy2) + pow(Evalue[1], aBy2) + pow(Evalue[2], aBy2);
+
+        // Transpose (also inverse) of the eigenvector matrix
+        Matrix3 EigenBasis;
+        for (auto m = 0; m < Evect.rows(); ++m)
+            for (auto n = 0; n < Evect.cols(); ++n) EigenBasis(m, n) = Evect(m, n);
```
bakpaul:
```suggestion
            for (auto n = 0; n < Evect.cols(); ++n) 
                 EigenBasis(m, n) = Evect(m, n);
```
+ same comment as before

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
+
+        // Construct C^(alpha1/2 - 1) from eigenbasis: V * D * V^T; D_i = lambda_i^(alpha1/2 - 1)
+        const Real aBy2Minus1 = aBy2 - 1.;
+        MatrixSym D(pow(Evalue[0], aBy2Minus1), 0, pow(Evalue[1], aBy2Minus1), 0, 0, pow(Evalue[2], aBy2Minus1));
+        MatrixSym CaBy2Minus1;
+        sofa::type::MatSym<3, Real>::Mat2Sym(EigenBasis*D.SymMatMultiply(EigenBasis.transposed()), CaBy2Minus1);
```
bakpaul: Looking at this for the third time I wonder if a part of all this computation could be factorized and somewhat optimized by keeping the value in one timestep.

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
```diff
@@ -61,178 +75,216 @@ class Ogden: public HyperelasticMaterial<DataTypes>
...
     }
 
+    void applyElasticityTensor(StrainInformation<DataTypes> *sinfo, const MaterialParameters<DataTypes> &param,
+                               const MatrixSym& inputTensor, MatrixSym& outputTensor) override
+    {
+        // For now, let's just multiply matrices using the ElasticityTensor explicitly
+        Matrix6 elasticityTensor;
+        this->ElasticityTensor(sinfo, param, elasticityTensor);
+        auto temp = elasticityTensor * inputTensor;
+        for (size_t i = 0; i < 6; i++) outputTensor[i] = temp[i]/2.;
```
bakpaul:
```suggestion
        for (size_t i = 0; i < 6; i++) 
            outputTensor[i] = temp[i]/2.;
```

bakpaul (commented):
Sorry I cannot check the math, I just checked that what you said that you where doing was what you where actually doing.

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
th-skam: I left a comment on it. Do not know what causes this from eigen's side.

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
th-skam: That's done by updating the parameters in the FEMForceFields (0229b72)

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
th-skam: Cannot use `Evect` directly with sofa's MatSym functions because it's got a type from eigen. 
So here we have to do this assignment to `EigenBasis` or write the multiplication explicitly.

Sofa/Component/SolidMechanics/FEM/HyperElastic/src/sofa/component/solidmechanics/fem/hyperelastic/material/Ogden.h
th-skam: Also done

th-skam:
The regression test failures were in fact caused by incorrect reference data.
I had omitted one of the model parameters in the terms of the elasticity tensor. The mistake went unnoticed by the `Material_test.cpp` test because the parameter values were set equal or close to 1, so the tests still passed. This was the case up until commit 2c02c41. From that commit and backwards, changing the model parameter values makes the unit tests fail.

So, I regenerated the references using commit 362023d (from this stage and onwards, tests are passed for parameters =/= 1). 
Additionally, I switched back to computing terms in the function calls instead of once at every time step because such a change had an effect on the end-state. I will have to dig in more to determine whether it is proper or not.

epernod (approved):

epernod:
[ci-build][with-all-tests]

PR #5754, opened 2025-10-30
[Visual] VisualGrid: do not update bbox

EulalieCoevoet:
**Problem**

We use the `VisualGrid` to understand and visualize the scene distance unit and to quickly measure distances when analyzing a simulation. For aesthetic purposes, our grids are set to be n times larger than the scene bounding box.
Currently, the `VisualGrid` component contributes to the scene bounding box, which I would argue should not be the case in general.

**In this PR**

Removes the bbox update. 
If you prefer, I can add a boolean to choose whether to update the bbox.

**Screenshot**

image

### DIFF
--- a/Sofa/Component/Visual/src/sofa/component/visual/VisualGrid.cpp
+++ b/Sofa/Component/Visual/src/sofa/component/visual/VisualGrid.cpp
@@ -80,15 +80,6 @@ void VisualGrid::updateGrid()
     }
 
     buildGrid();
-
-    //bounding box for the camera
-    auto s = d_size.getValue() * 0.5f;
-    sofa::type::Vec3 min(-s, -s, -s);
-    sofa::type::Vec3 max( s,  s,  s);
-    const auto& plane = d_plane.getValue();
-    min[static_cast<unsigned int>(plane)] = -s * 0.2f;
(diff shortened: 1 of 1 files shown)


damienmarchal:
Thank for the PR. 

I'm not sure about the fix, I have impression that the expected behavior for any VisualXXX object (like a visual model) is to update the bbox.

I wonder if the problem you are fixing is not a consequence of using scene component to implement viewer's Gizmos.  If this is the case, then these visual helper shouldn't be part of the scene but should be an utility object of the BaseViewer. To control the rendering of Gizmos of the viewer's should have binding to python or scene component, changing the viewer properties (as done in ViewerSetting). 

I agree that what I'm suggesting would be a more drastic move compared to what you are proposing but it would make thing much simpler as it would apply for every UI gizmo's.

bakpaul:
SOFA-dev discussion: this doesn't comply with how the API is meant to be used. If we break this for this object then the API will not make sense. So adding a boolean is preferred on our side. 

Another solution would be to add the grid inside a node and use parameter `tags="NoBBox"`

EulalieCoevoet:
I understand. I'll implement Damien's suggestion when I have time.

PR #5757, opened 2025-10-30
[Sofa.GUI.Common] Fix the setting of viewer's camera as it is not working properly currently

damienmarchal:
Currently only a camera at the root node of the scene is detected during initialization of the BaseViewer. 
This is problematic as it is very usefull to have a camera embeded in some subnode of the scene.

The problem is because the Node::get() method with default parameter is searching "upward".  As we are at root node this means that only a camera at root node is detected. It would be searching in the node and downward. 


Sofa/GUI/Common/src/sofa/gui/common/BaseViewer.cpp
```diff
@@ -264,18 +264,16 @@ bool BaseViewer::load()
...
+        groot->get(currentCamera, core::objectmodel::BaseContext::SearchDown);
         if (!currentCamera)
         {
             currentCamera = sofa::core::objectmodel::New<sofa::component::visual::InteractiveCamera>();
             currentCamera->setName(groot->getNameHelper().resolveName(currentCamera->getClassName(), sofa::core::ComponentNameHelper::Convention::python));
             groot->addObject(currentCamera);
-            //currentCamera->d_position.forceSet();
-            //currentCamera->d_orientation.forceSet();
             currentCamera->bwdInit();
+            msg_warning("BaseViewer") << "There is no camera in this scene, I created one. To remove this error message, add a camera in your scene.";
```
EulalieCoevoet: When we build a simulation, we expect users to address the warnings. I don't think adding a camera should be mandatory. So I think an info message would be more appropriate: `msg_info("BaseViewer") << A default camera has been added because none was found in the scene graph.`

hugtalbot (approved):

PR #5798, opened 2025-11-28
[FileSystem] Add openFileWithDefaultApplication method

EulalieCoevoet:
Used to open the file where a node or component is implemented or instantiate (see https://github.com/SofaComplianceRobotics/SofaGLFW/pull/92)


Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
```diff
@@ -497,6 +498,34 @@ std::string FileSystem::append(const std::string_view& existingPath, const std::
...
+            msg_error("FileSystem::openFileWithDefaultApplication()") << "File does not exist: " << filename;
+            return success;
+        }
+
+#ifdef WIN32
+        if ((INT_PTR)ShellExecuteA(nullptr, "open", filename.c_str(), nullptr, nullptr, SW_SHOWNORMAL) > 32)
+            success = true;
+#elif defined(__APPLE__)
+        const std::string command = "open \"" + filename + "\"";
+        if (std::system(command.c_str()) == 0)
```
fredroy: `std::system()` is not safe ; as a user could call this function with an input crafted to do malicious things...
So the solution would be to sanitize the inputs or the best solution is to use `fork()` and `execvp()` (for linux at least.)

fredroy (changes requested):

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
EulalieCoevoet: I see, sorry I didn't know it was bad practice to use `std::system`. I'll make the changes for both macOS and linux.

EulalieCoevoet: I have replaced `std::system` with `posix_spawn` for Linux and `popen` for MacOS.
I haven't tested the MacOS implementation.

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
```diff
@@ -497,6 +507,43 @@ std::string FileSystem::append(const std::string_view& existingPath, const std::
...
+            msg_error("FileSystem::openFileWithDefaultApplication()") << "File does not exist: " << filename;
+            return success;
+        }
+
+#ifdef WIN32
+        if ((INT_PTR)ShellExecuteA(nullptr, "open", filename.c_str(), nullptr, nullptr, SW_SHOWNORMAL) > 32)
+            success = true;
+#elif defined(__APPLE__)
+        const std::string command = "open \"" + filename + "\"";
+        FILE* pipe = popen(command.c_str(), "r+");
```
fredroy: Apparently, popen does more or less the same as std::system so still unsecure if filename is badly formatted.
Claude (🫢) tells me that posix_spawn is doable on macOS as well so
```
    pid_t pid;
    char* argv[] = {
        const_cast<char*>("open"),
        const_cast<char*>(filename.c_str()),
        nullptr
    };
    
    int status = posix_spawn(&pid, "/usr/bin/open", nullptr, nullptr, argv, environ);
    if (status == 0) {
        waitpid(pid, &status, 0);
        success = ( WIFEXITED(status) && WEXITSTATUS(status) == 0);
    }

```
(similar to your linux version)

should be OK.

fredroy (changes requested):

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
EulalieCoevoet: Thanks! I made the changes.

fredroy (approved):

PR #5805, opened 2025-12-01
[gui] Cleaning BaseViewer selection

EulalieCoevoet:
**Proposition:**
Use a single color for all selection drawings. In my opinion, it helps you understand what in the 3D scene rendering corresponds to the selection. I vote for orange, as we (Compliance Robotics) already use this color for selections elsewhere.

https://github.com/user-attachments/assets/1a118b30-a3f9-4a2b-83a1-21a9f62ec2f5

**In this PR:**
- Clear `currentSelection` in `load()` (e.g., when reloading the simulation or opening a new scene file)
- Refactor `size` (pixels) and `color`
- Position gizmo sizes are doubled compared to lines (e.g., bbox, meshes)
- Only draw `bbox` when valid
- Only draw rigid positions when `currentCamera` exists (prevents crashes)


Sofa/GUI/Common/src/sofa/gui/common/BaseViewer.cpp
```diff
@@ -352,13 +353,15 @@ void BaseViewer::drawSelection(sofa::core::visual::VisualParams* vparams)
     if(currentSelection.empty())
         return;
 
+    const float size = 2.f;
+    const type::RGBAColor color = type::RGBAColor::orange();
```
hugtalbot: Since orange is the color of CollisionModel, we should use another color.
Suggested by the devs : purple ! (to avoid the current yellow)

EulalieCoevoet: Can I make it a parameter, `m_selectionColor`?   
Like you already did with `m_visualScaling`

hugtalbot: Yes, it could be then even customized by the user in a dedicated window of the GUI

EulalieCoevoet: I've made the changes:
- the color is now a member variable - `m_selectionColor` - of the class `BaseViewer`
- and the default color is purple

EulalieCoevoet: We would like to have this feature in our release of December. Do you think we could merge? Unless you have other concerns.

hugtalbot (approved):

Sofa/GUI/Common/src/sofa/gui/common/BaseViewer.cpp
hugtalbot: thanks @EulalieCoevoet  for letting us know
we milestoned it accordingly

PR #5823, opened 2025-12-11
[Node] A node name can now be surrounded by quotation marks 

remibessard:
(allow node names with / , e.g. \'/universe/base\')

Motivation was to import OpenSim simulation in SOFA, where articulated chain architecture in the scene is described by frames with names such as "/universe/base".
This will be replicated in the SOFA scene using Sofa.RigidBodyDynamics plugin, while trying to stay as close as possible to OpenSim's description for the sake of future comparisons.

### DIFF
--- a/Sofa/framework/Simulation/Core/src/sofa/simulation/Node.cpp
+++ b/Sofa/framework/Simulation/Core/src/sofa/simulation/Node.cpp
@@ -508,10 +508,21 @@ sofa::core::objectmodel::Base* Node::findLinkDestClass(const core::objectmodel::
         }
         else
         {
-            std::size_t p2pos = pathStr.find('/',ppos);
-            if (p2pos == std::string::npos) p2pos = psize;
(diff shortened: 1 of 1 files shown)


damienmarchal:
Hello @remibessard 

I'm curious about this change. 
I wonder if:
- it wouldn't it be better and clear to map the frames hierarchies from OpenSim to our classical Node/Component structure
- or alternatively if an adaptation layer converting "/" to another symbol wouldn't be an safer option
- finally, if the two first are not desirable, if it wouldn't be preferable to use \ to protect from both "spaces" and "/" from parser interpretation... having quotation mark in the middle of the string looks weird to me.

remibessard:
Thanks for your feedback @damienmarchal .

I would go for the 2. as regarding 1. I prefer to keep SOFA simulations's hierarchies closer to Sofa.RigidBodyDynamics  structure compared to OpenSim's hierarchies (pinocchio/SOFA mappings implied in such simulations).

PR #5839, opened 2025-12-22
[SceneChecking] Collision: new checker on the presence of a pipeline and models

fredroy:
Some scenes has a whole collision pipeline but no collision in their scene whatsoever, 99% because of copy paste.

This checker just checks if there is a collision pipeline, and warns if there is no collision models (and vice versa)

[with-all-tests]

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #5794, opened 2025-11-26
[Core] Start renaming BaseObject to BaseComponent

alxbilger:
A not-working PR to start the discussion on renaming BaseObject to BaseComponent. It makes more sense for me as we name the classes in the factory 'Component'.

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #5841, opened 2025-12-23
[Collision] Introduce multi-staged collision pipeline

fredroy:
Based on
- #5839 

Add the ability to have multiple collision pipeline in a scene, instead of one monolithic pipeline (presented at the STC20)
This allows a lot of things:
 - group collision model with {alarm,contact} distances (can mitigate some tunneling-effect artifacts during collisions)
 - different response for a group of models (not tested)
 - mix different type of intersections
 - easily add custom pipelines
 - maybe more

It adds a warning if no collision model are handled by any pipeline.

This PR also adds a not-an-alias for `CollisionPipeline` similar to `BruteForceDetection`, but in this case I would not deprecate it and would be more seen as a convenience for most users, as the whole Multi + Sub collision pipeline is quite verbose in the end. And the former CollisionPipeline was doing a lot of things implicitly.

EDIT: with this, PipelineImpl is not used anymore (and has ... a dubious implementation by the way)

[with-all-tests]

[ci-depends-on https://github.com/sofa-framework/Regression/pull/96]

EDIT2: examples for a showcase
`w/o Composite Pipeline` [same collision distance, the 2nd sphere cannot go through]
demo_composite_pipeline_none

`w/ Composite Pipeline` [different collision distances, the 2nd case has a smaller one so the sphere can go through]
demo_composite_pipeline


fredroy:
[ci-build][with-all-tests]

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
```diff
@@ -0,0 +1,123 @@
...
+* along with this program. If not, see <http://www.gnu.org/licenses/>.        *
+*******************************************************************************
+* Authors: The SOFA Team and external contributors (see Authors.txt)          *
+*                                                                             *
+* Contact information: contact@sofa-framework.org                             *
+******************************************************************************/
+#pragma once
+#include <sofa/component/collision/detection/algorithm/config.h>
+
+#include <sofa/core/objectmodel/BaseObject.h>
```
hugtalbot: is this include required since we include other classes inheriting from BaseObject?
This may have to evolve with regards to https://github.com/sofa-framework/sofa/pull/5794

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CollisionPipeline.cpp
```diff
@@ -66,31 +66,44 @@ CollisionPipeline::CollisionPipeline()
     , d_depth(initData(&d_depth, defaultDepthValue, "depth",
                ("Max depth of bounding trees. (default=" + std::to_string(defaultDepthValue) + ", min=?, max=?)").c_str()))
 {
+    
+    m_subCollisionPipeline = sofa::core::objectmodel::New<SubCollisionPipeline>();
+    m_subCollisionPipeline->d_depth.setParent(&this->d_depth);
+    m_subCollisionPipeline->l_broadPhaseDetection.set(this->broadPhaseDetection);
+    m_subCollisionPipeline->l_narrowPhaseDetection.set(this->narrowPhaseDetection);
+    m_multiCollisionPipeline = sofa::core::objectmodel::New<MultiCollisionPipeline>();
+    m_multiCollisionPipeline->l_subCollisionPipelines.add(m_subCollisionPipeline.get());
```
bakpaul: I am not sure to exactly get what you intend to do here. It seems like you are expecting that if a CollisionPipeline is in the scene, then you expect to find the rest of the components and then you internally create a MultisubCollision + SubCollision. But what if we create another SubcollisionPipeline in the scene ? How t link it to the one created automatically ? 

If it is for compat purpose, why don't you just keep the old behavior ? Why forcing to instantiate a multiCollisionPipeline inside ? I suppose that it is to avoid duplicating code. Wouldn't it be cleaner to make CollisionPipeline inherits from SubCollisionPipeline ? Or at least add some static methods that does the job ?

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/MultiCollisionPipeline.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+void registerMultiCollisionPipeline(sofa::core::ObjectFactory* factory)
+{
+    factory->registerObjects(core::ObjectRegistrationData("Multiple collision pipelines in one.")
+        .add< MultiCollisionPipeline >());
+}
+
+MultiCollisionPipeline::MultiCollisionPipeline()
+    : d_parallelDetection(initData(&d_parallelDetection, false, "parallelDetection", "Parallelize collision detection."))
+    , d_parallelResponse(initData(&d_parallelResponse, false, "parallelResponse", "(DISABLED) Parallelize collision response."))
```
bakpaul: Why DISABLED ?

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CompositeCollisionPipeline.cpp
```diff
@@ -0,0 +1,245 @@
...
+        {
+            pipelineCollisionModels.insert(cm);
+        }
+    }
+
+    for (const auto& cm : sceneCollisionModels)
+    {
+        if (pipelineCollisionModels.find(cm) == pipelineCollisionModels.end())
+        {
+            msg_warning() << "CollisionModel " << cm->getName() << " is not handled by any SubCollisionPipeline.";
```
bakpaul: I would add, regarding my first comment, that this message will become very confusing for used it the old CollisionPipeline uses a multi one inside.

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/MultiCollisionPipeline.cpp
```diff
@@ -0,0 +1,245 @@
...
+void MultiCollisionPipeline::doCollisionReset()
+{
+    msg_info() << "MultiCollisionPipeline::doCollisionReset" ;
+
+    for(const auto& subPipeline : l_subCollisionPipelines)
+    {
+        subPipeline->computeCollisionReset();
+    }
+    
+    // re-order pipelines by order of distance
```
bakpaul: Why ?

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/MultiCollisionPipeline.h
```diff
@@ -0,0 +1,84 @@
...
+    
+    void draw(const core::visual::VisualParams* vparams) override;
+
+    /// Remove collision response from last step
+    virtual void computeCollisionReset() override;
+    /// Detect new collisions. Note that this step must not modify the simulation graph
+    virtual void computeCollisionDetection() override;
+    /// Add collision response in the simulation graph
+    virtual void computeCollisionResponse() override;
+
```
bakpaul:
```suggestion
    virtual void computeCollisionReset() override final;
    /// Detect new collisions. Note that this step must not modify the simulation graph
    virtual void computeCollisionDetection() override final;
    /// Add collision response in the simulation graph
    virtual void computeCollisionResponse() override final;
    
```

bakpaul (changes requested):

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/MultiCollisionPipeline.cpp
fredroy: The answer is in MultiCollisionPipeline::doCollisionResponse() in the comments, where it is stated that there is a potential (and real) race condition on Node

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CompositeCollisionPipeline.cpp
fredroy: Nope, as the CollisionPipeline will fetch all collision models, and set in the link.
https://github.com/fredroy/sofa/blob/fed32d2df4806edaf37f36fd34e47ad24f6c6472/Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CollisionPipeline.cpp#L89

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/MultiCollisionPipeline.cpp
fredroy: Initially it was in the case of if two sub collision pipelines were testing on the same collision models, to avoid doing twice/thrice whatever the same collision tests. But not sure if useful actually

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/MultiCollisionPipeline.h
fredroy: Not sure about that, if one would want to do their own MultiCollisionPipeline (specific order, needs, etc)

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CompositeCollisionPipeline.cpp
bakpaul: YEah, but the message will be prefixed with "[WARNING] [MultiCollisionPipeline(unnamed)]" which will be confusing for someone using an oold scene that doesn't contain any MultiCollisionPipeline. 

Anyway, I really think that the original CollisionPipeline is just a special case of SubCollisionPipeline that is alone and doesn't belong to any multi one. The inheritance seems more appropriate for me.

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/MultiCollisionPipeline.h
bakpaul: You've introduced the template method pattern in those methods of the API. If anyone wants to make their own, they'll rewrite the "do" equivalent.

fredroy: right

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
```diff
@@ -0,0 +1,123 @@
...
+
+#include <sofa/core/visual/VisualParams.h>
+
+#include <set>
+#include <string>
+
+namespace sofa::component::collision::detection::algorithm
+{
+
+class SOFA_COMPONENT_COLLISION_DETECTION_ALGORITHM_API AbstractSubCollisionPipeline : public sofa::core::objectmodel::BaseObject
```
alxbilger: Would it be possible to document this class? Also the main virtual functions. Thanks

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
```diff
@@ -0,0 +1,123 @@
...
+            msg_warning() << "A contact manager component is required to compute collision detection.";
+            validity = false;
+        }
+        
+        if (validity)
+        {
+            this->d_componentState.setValue(sofa::core::objectmodel::ComponentState::Valid);
+        }
+                
+        doInit();
```
alxbilger: Should we call `doInit` when `validity==false`?

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    static std::set< std::string > getResponseList()
+    {
+        std::set< std::string > listResponse;
+        core::collision::Contact::Factory::iterator it;
+        for (it = core::collision::Contact::Factory::getInstance()->begin(); it != core::collision::Contact::Factory::getInstance()->end(); ++it)
```
alxbilger: A range-based for loop with structured bindings would be preferable here.

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
```diff
@@ -0,0 +1,123 @@
...
+    virtual void computeCollisionReset() = 0;
+    virtual void computeCollisionDetection() = 0;
+    virtual void computeCollisionResponse() = 0;
+    virtual void doInit() = 0;
+    virtual void doBwdInit() {}
+    virtual void doHandleEvent(sofa::core::objectmodel::Event* e) = 0;
+    
+    virtual void doDraw(const core::visual::VisualParams*) {}
+    
+    AbstractSubCollisionPipeline()
```
alxbilger: Should be protected.

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
```diff
@@ -0,0 +1,123 @@
...
+{
+public:
+    SOFA_ABSTRACT_CLASS(AbstractSubCollisionPipeline, sofa::core::objectmodel::BaseObject);
+
+    virtual void computeCollisionReset() = 0;
+    virtual void computeCollisionDetection() = 0;
+    virtual void computeCollisionResponse() = 0;
+    virtual void doInit() = 0;
+    virtual void doBwdInit() {}
+    virtual void doHandleEvent(sofa::core::objectmodel::Event* e) = 0;
```
alxbilger: Should be protected

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
```diff
@@ -0,0 +1,123 @@
...
+    
+    virtual void doDraw(const core::visual::VisualParams*) {}
```
alxbilger: Should be protected

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
alxbilger, line 1: The implementations of the methods can be in a cpp file.

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CollisionPipeline.cpp
```diff
@@ -66,31 +66,44 @@ CollisionPipeline::CollisionPipeline()
...
+    
+    this->addSlave(m_subCollisionPipeline);
+    this->addSlave(m_multiCollisionPipeline);
```
alxbilger: Can you confirm that they should be slaves?

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CollisionPipeline.cpp
```diff
@@ -66,31 +66,44 @@ CollisionPipeline::CollisionPipeline()
...
-    if (contactManager == nullptr)
-    {
-        msg_warning() << "A ContactManager component is required to compute collision response and was not found in the current scene";
-    }
-
+    
+    msg_info() << "CollisionPipeline is now a wrapper to MultiCollisionPipeline with a single SubCollisionPipeline.";
+    msg_info() << "If you want more flexibility, use directly the components MultiCollisionPipeline and SubCollisionPipeline, with their respective Data.";
+
+    auto context = this->getContext();
```
alxbilger, line 77:
```suggestion
    auto context = this->getContext();
    assert(context);
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/MultiCollisionPipeline.cpp
```diff
@@ -0,0 +1,245 @@
...
+        return;
+    }
+    
+    if(d_parallelDetection.getValue() || d_parallelResponse.getValue())
+    {
+        m_taskScheduler = sofa::simulation::MainTaskSchedulerFactory::createInRegistry();
+        assert(m_taskScheduler);
+        
+        m_taskScheduler->init();
+    }
```
alxbilger: You can derive from `TaskSchedulerUser`.

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
alxbilger: Shouldn't this class be named `BaseSubCollisionPipeline`? For consistency with all the other component classes?

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
fredroy: A bit of a stretch but maybe the (in)-validity could be solved in the inheriting class 🤔

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CollisionPipeline.cpp
fredroy: They are supposed to be totally controlled by the "master" (aka MultiCollisionPipeline) and thus should be not be used by the common visitor and such

fredroy:
[ci-build][with-all-tests]

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CompositeCollisionPipeline.cpp
hugtalbot: The change has been made by @fredroy 
We could finally simply add the name of the class:
```cpp
    m_subCollisionPipeline = sofa::core::objectmodel::New<SubCollisionPipeline>();
    m_subCollisionPipeline->d_depth.setParent(&this->d_depth);
    m_subCollisionPipeline->d_name.setValue("CollisionPipeline");
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    msg_info()
+         << "doCollisionDetection, compute Bounding Trees" ;
+
+    // First, we compute a bounding volume for the collision model (for example bounding sphere)
+    // or we have loaded a collision model that knows its other model
+
+
+
+    type::vector<CollisionModel*> vectBoundingVolume;
```
hugtalbot:
```suggestion
    // First, we compute a bounding volume for the collision model (for example bounding sphere)
    // or we have loaded a collision model that knows its other model
    
    type::vector<CollisionModel*> vectBoundingVolume;
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+        const int used_depth = (
+                    (l_broadPhaseDetection->needsDeepBoundingTree()) ||
+                    (l_narrowPhaseDetection->needsDeepBoundingTree())
+            ) ? d_depth.getValue() : 0;
+
+        for (auto it = l_collisionModels.begin(); it != l_collisionModels.end(); ++it)
+        {
+            msg_info()
+                << "doCollisionDetection, consider model" ;
```
hugtalbot:
```suggestion
            msg_info() << "doCollisionDetection, consider model" ;
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+                (*it)->computeBoundingTree(used_depth);
+            }
+
+            vectBoundingVolume.push_back ((*it)->getFirst());
+            ++nActive;
+        }
+
+
+        msg_info()
+                << "doCollisionDetection, Computed "<<nActive<<" Bounding Boxes." ;
```
hugtalbot:
```suggestion
        msg_info() << "doCollisionDetection, Computed "<<nActive<<" Bounding Boxes." ;
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+    }
+
+    msg_info()
+            << "doCollisionDetection, BroadPhaseDetection "<<l_broadPhaseDetection->getName();
```
hugtalbot:
```suggestion
    msg_info()  << "doCollisionDetection, BroadPhaseDetection "<<l_broadPhaseDetection->getName();
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+void SubCollisionPipeline::computeCollisionDetection()
+{
+    SCOPED_TIMER_VARNAME(docollisiontimer, "doCollisionDetection");
+
+    if (!this->isComponentStateValid())
+        return;
+
+    msg_info()
+         << "doCollisionDetection, compute Bounding Trees" ;
```
hugtalbot:
```suggestion
    msg_info()  << "doCollisionDetection, compute Bounding Trees" ;
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+        SCOPED_TIMER_VARNAME(broadphase, "BroadPhase");
+        l_intersectionMethod->beginBroadPhase();
+        l_broadPhaseDetection->beginBroadPhase();
+        l_broadPhaseDetection->addCollisionModels(vectBoundingVolume);  // detection is done there
+        l_broadPhaseDetection->endBroadPhase();
+        l_intersectionMethod->endBroadPhase();
+    }
+
+    msg_info()
+        << "doCollisionDetection, NarrowPhaseDetection "<< l_narrowPhaseDetection->getName();
```
hugtalbot:
```suggestion
    msg_info() << "doCollisionDetection, NarrowPhaseDetection "<< l_narrowPhaseDetection->getName();
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    {
+        SCOPED_TIMER_VARNAME(narrowphase, "NarrowPhase");
+        l_intersectionMethod->beginNarrowPhase();
+        l_narrowPhaseDetection->beginNarrowPhase();
+        const type::vector<std::pair<CollisionModel*, CollisionModel*> >& vectCMPair = l_broadPhaseDetection->getCollisionModelPairs();
+
+        msg_info()
+                << "doCollisionDetection, "<< vectCMPair.size()<<" colliding model pairs" ;
```
hugtalbot:
```suggestion
        msg_info()  << "doCollisionDetection, "<< vectCMPair.size()<<" colliding model pairs" ;
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+void SubCollisionPipeline::computeCollisionResponse()
+{
+    if (!this->isComponentStateValid())
+        return;
+
+    core::objectmodel::BaseContext* scene = getContext();
+
+    msg_info()
+        << "Create Contacts " << l_contactManager->getName() ;
```
hugtalbot:
```suggestion
    msg_info() << "Create Contacts " << l_contactManager->getName() ;
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    {
+        SCOPED_TIMER_VARNAME(createContactsTimer, "CreateContacts");
+        l_contactManager->createContacts(l_narrowPhaseDetection->getDetectionOutputs());
+    }
+
+    // finally we start the creation of collisionGroup
+
```
hugtalbot:
```suggestion
    // finally we start the creation of collisionGroup
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,282 @@
...
+            {
+                notStaticContacts.push_back(contact);
+            }
+        }
+    }
+
+    SCOPED_TIMER_VARNAME(createResponseTimer, "CreateMovingObjectsResponse");
+
+    msg_info()
+        << "Linking all contacts to Scene" ;
```
hugtalbot:
```suggestion
    msg_info() << "Linking all contacts to Scene" ;
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/AbstractSubCollisionPipeline.h
fredroy: I usually try to include stuff I am explicitly using (even if is already included somewhere else)
And #5794 does not seem to be ready any time soon 🤔

fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CollisionPipeline.cpp
```diff
@@ -66,33 +66,57 @@ CollisionPipeline::CollisionPipeline()
...
-    {
-        msg_warning() << "A NarrowPhase component is required to compute collision detection and was not found in the current scene";
-    }
-
-    if (contactManager == nullptr)
-    {
-        msg_warning() << "A ContactManager component is required to compute collision response and was not found in the current scene";
-    }
-
+    msg_info() << "CollisionPipeline is now a wrapper to CompositeCollisionPipeline with a single SubCollisionPipeline.";
```
alxbilger:
```suggestion
    msg_info() << "Since v26.06, CollisionPipeline is a wrapper to CompositeCollisionPipeline with a single SubCollisionPipeline.";
```

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+void registerSubCollisionPipeline(sofa::core::ObjectFactory* factory)
+{
+    factory->registerObjects(core::ObjectRegistrationData("Collision pipeline to be used with CompositeCollisionPipeline.")
+        .add< SubCollisionPipeline >());
+}
+
+SubCollisionPipeline::SubCollisionPipeline()
+    : Inherited()
+    , d_depth(initData(&d_depth, s_defaultDepthValue, "depth", +("Max depth of bounding trees. (default=" + std::to_string(s_defaultDepthValue) + ", min=?, max=?)").c_str()))
```
alxbilger: What is the `+` symbol in front of the help message? Not sure min and max are relevant here.

bakpaul (approved):

Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/SubCollisionPipeline.cpp
fredroy: I dont know, it was (still is) in the original CollisionPipeline https://github.com/fredroy/sofa/blob/486b82400ac7122bfc3a27c59f0b89b1d4efdb06/Sofa/Component/Collision/Detection/Algorithm/src/sofa/component/collision/detection/algorithm/CollisionPipeline.cpp#L66

fredroy: but yeah, the '+'must have been a typo... still, it is  funny it compiled tho

bakpaul:
**Removed**

PR #5854, opened 2026-01-08
[Visual] Update BaseCamera init logic when a camera is set in the scene with custom Data inputs

epernod:
This is just to be able to set manually an Interactive/BaseCamera in the scene with wanted Data for position + orientation or position + lookAt and also set ZNear and ZFar values that is not overridden at start.

Quick Note: Another solution to be able to change ZNear and ZFar values:
Do not add Camera in the scene and use the default one. Save the camera position and then edit the .view file. ZNear and ZFar are specified in it.


epernod:
I broke something :/

PR #5873, opened 2026-01-20
Quick test on current regression on the CI

epernod:
Just checking that the regression fails if the state reference file (based on the mechanicalObject name) is not found

[with-all-tests]
[ci-depends-on https://github.com/sofa-framework/Regression/pull/98]

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -168,14 +168,6 @@
           "type": "BOOL",
           "value": "ON"
         },
-        "SOFA_FETCH_SOFAVISCOELASTIC": {
-          "type": "BOOL",
-          "value": "ON"
-        },
-        "PLUGIN_SOFAVISCOELASTIC": {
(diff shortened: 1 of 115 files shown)


epernod:
[ci-build][with-all-tests]

epernod:
I'm afraid to say that my fear were right. There is a mismatch of reference filename against current mechanicalObject names. And the ReadState is not returning false if the ref is not found.

You can see that I change the name of a mechanicalObject and totally mess up the mass of several torus in the chain.scn but the regression is still passing:

>[INFO]    [BaseRegression_test::runStateRegressionTest] Testing /workspace/sofa/examples/Demos/chainAll.scn
> [       OK ] Regression_test/StateRegression_test.sceneTest/chainAll_scn (7416 ms)
> [ RUN      ] Regression_test/StateRegression_test.sceneTest/chainHybrid_scn

PR #5871, opened 2026-01-20
[Lagrangian.Solver] UnbuiltConstraintSolver: Fix resetForUnbuiltResolution (constraint re-ordering)

fredroy:
In a BeamAdapter scene where a catheter is deployed into *deformable* vessels, I got a huge discrepancy between LCPConstraintSolver (unbuilt) and UnbuiltGaussSeidelConstraintSolver (being the slowest).
Whereas the same scene (aka the standard scene [3instruments_collis](https://github.com/sofa-framework/BeamAdapter/blob/master/examples/3instruments_collis.scn) where the catheter is deployed into static vessels, the difference is not that huge (~15-20% slower)

--> very important:  wire_optimization must be *true*  (more on that later)

Bench: 
- 5000 steps on 3instru_collis (static)
```
LCP : 5000 iterations done in 19.9832 s ( 250.21 FPS)
UGS:: 5000 iterations done in 23.5956 s ( 211.904 FPS)

```
- 5000 steps on 3instru_collis (deformable)
```
LCP :5000 iterations done in 43.9782 s ( 113.693 FPS)
UGS: 5000 iterations done in 97.7162 s ( 51.1686 FPS)
```

For reference WITHOUT wire_optimization (set to False)
```
LCP :5000 iterations done in 63.7577 s ( 78.4219 FPS)
UGS: 5000 iterations done in 99.2367 s ( 50.3846 FPS)
```

We can see that  wire_optimization seems to do nothing for UGS.

After investigation (thanks to Claude 🤖) it appeared that the re-ordering of the constraints was cancelled in the case of there is TWO constraint corrections (which is the case for the deformable vessels but not for the static vessels)

The loop calling resetForUnbuiltResolution on each ConstraintCorrection was resetting the ordering on the second iteration of the loop (e.g if the list is { LinearCC, UncoupledCC } then the first iteration will re-order fine but the second loop will reset the list and UncoupledCC do nothing for the re-oredering)

All in all the fix is easy, and set the initialization of the list outside the loop... 🤪

Results:

- 5000 steps on 3instru_collis (deformable)
```
UGS: 5000 iterations done in 49.5163 s ( 100.977 FPS)
```

with the same ~15% difference as with the static vessel case.

✍️ Note that there should not be any difference between LCP and UGS but more to come about that....

[with-all-tests]
______________________________________________________

By submitting this pull request, I acknowledge that  
[...]

PR #5875, opened 2026-01-21
[Lagrangian.Solver] UnbuiltConstraintSolver: implement hotstart

fredroy:
Based on
- #5871 

in #5871 (and more cases) it was noted that (unbuilt) LCP was *always* faster than UnbuiltGaussSeidelConstraintSolver (in average ~10-20 %) and usually converged in less iterations.
Even if the algorithm is rigorously the same (apparently...)

After investigations, Claude 🤖successfully found that the main difference is UnbuiltGaussSeidelConstraintSolver was always starting its GS with the forces at 0 (cold start), whereas LCP was keeping track of the previous forces.
It also provided the implementation (I could never do that by myself 💩) which... looks good to me at least 🙃

Benches (continuation of #5871)
- 5000 steps on 3instru_collis (deformable), wire_optimization on
```
LCP                : 5000 iterations done in 43.9782 s ( 113.693 FPS)
UGS (w/o hotstart) : 5000 iterations done in 49.5163 s ( 100.977 FPS)
UGS (w hotstart)   : 5000 iterations done in 44.0242 s ( 113.574 FPS)
```

- for the fun, wire_optimization off
```
LCP                : 5000 iterations done in 63.7577 s ( 78.4219 FPS)
UGS (w/o hotstart) : 5000 iterations done in 99.2367 s ( 50.3846 FPS)
UGS (w hotstart)   : 5000 iterations done in 62.5122 s ( 79.9844 FPS)
```

This PR could be considered a little bit... controversial (:vibe:) as Claude-code implemented lots of stuff

[with-all-tests]

### DIFF
--- a/Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
+++ b/Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
@@ -70,7 +70,6 @@ GenericConstraintSolver::GenericConstraintSolver()
     , d_regularizationTerm(initData(&d_regularizationTerm, 0.0_sreal, "regularizationTerm", "Add regularization factor times the identity matrix to the compliance W when solving constraints"))
     , d_scaleTolerance(initData(&d_scaleTolerance, true, "scaleTolerance", "Scale the error tolerance with the number of constraints"))
     , d_allVerified(initData(&d_allVerified, false, "allVerified", "All constraints must be verified (each constraint's error < tolerance)"))
-    , d_initialGuess(initData(&d_initialGuess, true, "initialGuess", "Activate constraint force history to improve convergence (hot start)"))
     , d_computeGraphs(initData(&d_computeGraphs, false, "computeGraphs", "Compute graphs of errors and forces during resolution"))
     , d_graphErrors(initData(&d_graphErrors, "graphErrors", "Sum of the constraints' errors at each iteration"))
     , d_graphConstraints(initData(&d_graphConstraints, "graphConstraints", "Graph of each constraint's error at the end of the resolution"))
@@ -188,13 +187,13 @@ bool GenericConstraintSolver::buildSystem(const core::ConstraintParams *cParams,
     applyProjectiveConstraintOnConstraintMatrix(cParams);
 
     // Get constraint info for hot-start BEFORE clear() resets the constraint count check
-    getConstraintInfo(cParams);
+    preClearCorrection(cParams);
 
     //clear and/or resize based on the number of constraints
     current_cp->clear(numConstraints);
 
     // Restore initial guess from previous timestep AFTER clear() zeros the forces
-    computeInitialGuess();
+    postClearCorrection();
 
     getConstraintViolation(cParams, &current_cp->dFree);
 
@@ -387,8 +386,7 @@ void GenericConstraintSolver::storeConstraintLambdas(const core::ConstraintParam
 
 bool GenericConstraintSolver::applyCorrection(const core::ConstraintParams *cParams, MultiVecId res1, MultiVecId res2)
 {
-    // Save forces for hot-start in next timestep
-    keepContactForcesValue();
+    doPreApplyCorrection();
 
     computeAndApplyMotionCorrection(cParams, res1, res2);
     storeConstraintLambdas(cParams);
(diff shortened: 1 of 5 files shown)


fredroy:
[ci-build][with-all-tests]

fredroy:
[ci-build][with-all-tests]

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/UnbuiltConstraintSolver.cpp
```diff
@@ -102,5 +125,110 @@ void UnbuiltConstraintSolver::initializeConstraintProblems()
...
+                }
+            }
+        }
+    }
+}
+
+void UnbuiltConstraintSolver::keepContactForcesValue()
+{
+    if (!d_initialGuess.getValue())
+        return;
```
hugtalbot: should this condition not be rather before line 107 ?
(for readability)

hugtalbot (approved):

fredroy:
Before merging, I would like a review (or at least the opinion) of either @ChristianDuriez or @EulalieCoevoet

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/UnbuiltConstraintSolver.h
```diff
@@ -40,11 +41,50 @@ class SOFA_COMPONENT_CONSTRAINT_LAGRANGIAN_SOLVER_API UnbuiltConstraintSolver :
...
+
+    /// Get constraint info (block info and persistent IDs) for hot-start
+    void getConstraintInfo(const core::ConstraintParams* cparams);
+
+    // Hot-start data storage
+    std::map<core::behavior::BaseLagrangianConstraint*, ConstraintBlockBuf> m_previousConstraints;
+    type::vector<SReal> m_previousForces;
+    VecConstraintBlockInfo m_constraintBlockInfo;
+    VecPersistentID m_constraintIds;
+    unsigned int m_numConstraints{0}; ///< Number of constraints from current/previous timestep
```
bakpaul, line 88: Can you prefix all of these by previous so it is clear that they are used for storing values of previous time step ?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/GenericConstraintSolver.cpp
```diff
@@ -186,9 +186,15 @@ bool GenericConstraintSolver::buildSystem(const core::ConstraintParams *cParams,
     // suppress the constraints that are on DOFS currently concerned by projective constraint
     applyProjectiveConstraintOnConstraintMatrix(cParams);
 
+    // Get constraint info for hot-start BEFORE clear() resets the constraint count check
+    preClearCorrection(cParams);
+
```
bakpaul, line 191: I don't get why this is needed. When you look at the implementation of the doPreClear, clearing the curren_cp wouldn't change a thing in what MechanicalGetConstraintInfoVisitor will get. So for me the pre/post is not necessary. could you try mergin both and calling the merged method after the clear ?

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/UnbuiltConstraintSolver.cpp
```diff
@@ -102,5 +128,107 @@ void UnbuiltConstraintSolver::initializeConstraintProblems()
...
+
+    // Clear previous history (mark all as invalid)
+    for (auto& previousConstraint : m_previousConstraints)
+    {
+        ConstraintBlockBuf& buf = previousConstraint.second;
+        for (auto& it2 : buf.persistentToConstraintIdMap)
+        {
+            it2.second = -1;
+        }
+    }
```
bakpaul, line 210: Wouldn't it be better to simply clear the whole map and start fresh ? Otherwise if the parent keep changing, this map will grow without getting properly cleaned

Sofa/Component/Constraint/Lagrangian/Solver/src/sofa/component/constraint/lagrangian/solver/UnbuiltConstraintSolver.cpp
```diff
@@ -102,5 +128,107 @@ void UnbuiltConstraintSolver::initializeConstraintProblems()
...
+        force[c] = 0.0;
+    }
+    
+    // Then restore forces from previous timestep for matching persistent IDs
+    for (const ConstraintBlockInfo& info : m_constraintBlockInfo)
+    {
+        if (!info.parent) continue;
+        if (!info.hasId) continue;
+
+        auto previt = m_previousConstraints.find(info.parent);
```
bakpaul, line 164: Using the parent as a key is somewhat strange because it it irrelevant @the level of the solver. What is relevant is the element on which the force is applied. This information is only kept in the constraint jacobian matrix. At this level I don't think it is possible to get an actual proper hotstart. What is done here it is looking at the id of the constraint at the level of the component which transform the collision detection output into the constraint jacobian. So what you are comparing here is the place of the force in the original vector of the collision detection. So for me this hotstart is just as good as copying back the content of the old force vector into the new one while clamping or padding if size mismatch. This is only by chance that you get forces that are good because the collision detection is somewhat deterministic.

The only strong difference would be when there are two component responsible of transforming this collision output into contact force, this would keep the force at least in the good part of the vector (the old forces will be used a starting point of forces coming from the same collision detection), but there is no garantee that the force is applied in the same region of the object as before. This happens when you have multiple objects colliding. But really, I don't know how much you gain with respect of a simple copy/clipping/padding work. 

Could you try out your example with this simple implementation to see if you get worst convergence ?

PR #5732, opened 2025-10-13
[CI] Update metis version

bakpaul:
Based on #5698
This PR intends to try if updating METIS works on the CI.

[ci-depends-on https://github.com/sofa-framework/Sofa.Metis/pull/16]

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #5878, opened 2026-01-21
[tools] Remove autoload of Sofa.Metis for windows

bakpaul:
Metis crashes on the release on windows for some scene (e.g. examples/Demos/SofaScene.scn), so we need to avoid auto-loading it on windows. 

This PR is only on the release branch in hope that we fix this for the next release. 

### DIFF
--- a/tools/postinstall-fixup/common.sh
+++ b/tools/postinstall-fixup/common.sh
@@ -41,7 +41,7 @@ function clean_default_plugins()
       ; do
       disabled_plugins=$disabled_plugins'\|'$plugin
   done
-  grep -v $disabled_plugins "$1/plugin_list.conf.default" >> "$1/plugin_list.conf"
+  grep -v $disabled_plugins$2 "$1/plugin_list.conf.default" >> "$1/plugin_list.conf"
 }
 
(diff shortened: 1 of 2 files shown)


olivier-roussel:
I could do further testing and I think the problem comes from the new version of metis (v5.2.1).
My working Pixi build was older than the metis version bump (https://github.com/sofa-framework/sofa/pull/5732). But I could reproduce the bug with fetched metis v5.2.1 from our fork. Scenes run fine with metis v5.1.0.

olivier-roussel:
~~The scenes also run fine when using the metis v5.2.1 package from conda-forge (warning: the current one is broken on Windows, a fix is in progress on conda-forge https://github.com/conda-forge/metis-feedstock/pull/47).
Even if we do not want to use the conda package all the time, this is good news as it could be due to one of the upstream patches of the conda forge package on v5.2.1, which we could integrate in our fork for v5.2.1.
I'll continue investigate on this.~~
Edit: I confirm I reproduce the issue using metis v5.2.1 from conda-forge. So this looks like a regression upstream somewhere between metis v5.1.0 and v5.2.1.

PR #5879, opened 2026-01-22
[Tools] Fix windows postinstall

olivier-roussel:
On Windows, all binaries that are linked at compilation stage should be in the same directory (or in some directory in the `PATH` env var, but we do not want to modify it and we should not).
This is already done for release builds at the postinstall step. This is also the case in build directory. But after a build install (i.e. cmake install call) on non-release builds, this was not done. This PR enables the binaries copy as done in releases for all Windows builds.

It also remove the collection related stuff in the postinstall script which are deprecated.

### DIFF
--- a/tools/postinstall-fixup/CMakeLists.txt
+++ b/tools/postinstall-fixup/CMakeLists.txt
@@ -3,7 +3,7 @@ if(NOT SOFA_BUILD_RELEASE_PACKAGE AND NOT CMAKE_SYSTEM_NAME STREQUAL Windows)
     return()
 endif()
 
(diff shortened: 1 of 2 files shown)


tools/postinstall-fixup/CMakeLists.txt
bakpaul:
```suggestion
if(TARGET Sofa.Qt AND SOFA_BUILD_RELEASE_PACKAGE)
```

bakpaul (changes requested):

epernod (approved):

bakpaul (approved):

PR #5885, opened 2026-01-23
[GitHub] Update install-nix-action to version 31

hugtalbot:
Changes according to https://github.com/cachix/install-nix-action/issues/224

### DIFF
--- a/.github/workflows/nix.yml
+++ b/.github/workflows/nix.yml
@@ -12,7 +12,7 @@ jobs:
         os: [ubuntu, macos]
     steps:
       - uses: actions/checkout@v4
-      - uses: cachix/install-nix-action@v27
+      - uses: cachix/install-nix-action@v31
       # TODO: the "sofa" account on cachix does not exist yet
       #- uses: cachix/cachix-action@v15
         #with:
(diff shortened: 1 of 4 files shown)


fredroy:
@hugtalbot 

I tried adding nholmann_json as a dependency as it was required to build, but now it is stuck on finding `tight_inclusion` which is not shipped by Nix (and certainly wont be).
Would it not be better to simply disable nix builds ?

olivier-roussel (dismissed):

olivier-roussel:
I agree with @fredroy . 
Unless somebody of us takes care of adding `tight_inclusion` to nix packages, it is quite unlikely this happens soon. So if we have nobody that can manage the nix support, may be we should drop it. 
Also, we only have Nix support for SOFA, but not for any plugins right ? In this case it might be not quite useful.

nim65s:
the good long term solution is to add that package in nixpkgs.
the fast short term one is to add it here.

I'm doing both. that is https://github.com/Continuous-Collision-Detection/Tight-Inclusion, right ?

nim65s:
I've opened #5916, which target the branch for this PR

epernod (approved):

Issue #5890, opened 2026-01-27
Component destructors not called when slaves are present


fredroy:
## Description

  Components that have slaves (via `l_slaves`) do not have their destructors called during scene unload, even though `cleanup()` is called correctly.

  ## Root Cause

  `BaseObject::l_slaves` is defined with `FLAG_STRONGLINK`:

  ```cpp
  // BaseObject.h:134
  typedef MultiLink<BaseObject, BaseObject, BaseLink::FLAG_DOUBLELINK|BaseLink::FLAG_STRONGLINK> LinkSlaves;
```
  When a component (e.g., TetrahedronFEMForceField) has slaves (e.g., Stiffness_matrix, Damping_matrix created by the matrix assembly system), these strong references are never released during cleanup. This prevents the reference count from reaching zero, so the destructor is never called.

##  Steps to Reproduce

  1. Create a scene with TetrahedronFEMForceField (or any component that acquires slaves)
  2. Run with batch GUI: runSofa -g batch scene.scn
  3. Quit normally
  4. Observe that TetrahedronFEMForceField::~TetrahedronFEMForceField() is never called, while cleanup() is called

  Expected Behavior

  The destructor should be called after cleanup() when the scene is unloaded.

##  Workaround

  Manually removing slaves in cleanup() fixes the issue:
```cpp
  void cleanup() override
  {
      while (!this->l_slaves.empty())
      {
          this->removeSlave(this->l_slaves.getValue().front());
      }
      Inherit1::cleanup();
  }
```
##  Suggested Fix

  BaseObject::cleanup() should automatically remove all slaves:
```cpp
  void BaseObject::cleanup()
  {
      while (!l_slaves.empty())
      {
          removeSlave(l_slaves.getValue().front());
      }
  }
```

PR #5891, opened 2026-01-27
[Core] BaseObject: remove components from the slave list

fredroy:
Related (and fix) : https://github.com/sofa-framework/sofa/issues/5890

Description: 
Searching for memory leaks, I detected big ones coming from TetrahedronFEMForcefield.
In particular the colorMap, which was supposed to be deleted in the destructor.
I noticed that the destructor was not called and then... asked Claude which found that slaves was keeping a reference to their owner, and so the reference counter of the Sptr was not at the end of the program (so not the dtor was not called).

Easy solution is to unlink the slaves during the cleanup phase (what is done in this PR)
Only problem is if a component is overriding this function and not calling super::cleanup() 💀
(template method design blabla for BaseObject blabla 🫠)

[with-all-tests]


hugtalbot (approved):

fredroy:
[ci-build][with-all-tests]

bakpaul (approved):

alxbilger:
I still cannot make up my mind if the responsibility to clean up slaves is on the base or derived class...

bakpaul:
It depends what does the remove do when called. Does the slave run a callback that might act on the owner ? Is there any way this callback could access the owner ? If this is the case, IMO the safest way to go would be to make the most derived class to do this. But in the meantime if it is the case, I guess there is a design flaw. 

So for me the best way to go is to make sure no such callback exists and the slaves never access their owner in the destructor and keep this in the BaseObject destructor

EulalieCoevoet:
@hugtalbot @fredroy  This breaks the mouse interaction with the spring force field. 
It crashes when we release the spring.

fredroy:
> @hugtalbot @fredroy This breaks the mouse interaction with the spring force field. It crashes when we release the spring.

I tried on my side (master sofa, master SofaGLFW, macOS) and it does not crash when I select, do some pulling and release, multiple times , with the default MouseInteractor (I guess the spring)

EulalieCoevoet:
It crashes when using the `Attach an object to the Mouse using a spring force field`, on release, with:
- Sofa master
- SofaGLFW master
- Tested on Linux and Windows

It works well without this PR. I'm sorry I haven't spent more time trying to find a proper fix.

**Screenshots**

image

image

PR #5904, opened 2026-01-29
[All] Security : memory management, destructors and al

fredroy:
(the misc last part)

- some dtors were not virtual and may not be called
- sanity checks on input 
- memory management with arrays, etc
- thread-safeness by replacing volatile by std::atomic (volatile dont guarantee about function ordering by the cpu)

[with-all-tests]

### DIFF
--- a/Sofa/GL/src/sofa/gl/VideoRecorderFFMPEG.cpp
+++ b/Sofa/GL/src/sofa/gl/VideoRecorderFFMPEG.cpp
@@ -35,6 +35,30 @@ using sofa::helper::system::FileSystem;
 namespace sofa::gl
 {
 
+namespace
+{
+/// Escape a string for safe use in a shell double-quoted context.
+/// Escapes: backslash, backtick, dollar, double-quote, and newline.
+std::string escapeForShell(const std::string& input)
+{
+    std::string result;
+    result.reserve(input.size() + 16);
+    for (char c : input)
+    {
+        switch (c)
+        {
+        case '\\': result += "\\\\"; break;
+        case '"':  result += "\\\""; break;
+        case '$':  result += "\\$";  break;
+        case '`':  result += "\\`";  break;
+        case '\n': result += " ";    break;  // Replace newline with space
+        default:   result += c;      break;
+        }
+    }
+    return result;
+}
+} // anonymous namespace
+
 VideoRecorderFFMPEG::VideoRecorderFFMPEG()
     : m_framerate(25)
     , m_prefix("sofa_video")
@@ -114,7 +138,7 @@ bool VideoRecorderFFMPEG::init(const std::string& ffmpeg_exec_filepath, const st
        << " -pix_fmt " << codec // yuv420p " // " yuv444p "
        << " -crf 17 "
        << " -vf vflip "
-       << "\"" << m_filename << "\""; // @TODO C++14 : replace with std::quoted
+       << "\"" << escapeForShell(m_filename) << "\"";
 
     const std::string& command_line = ss.str();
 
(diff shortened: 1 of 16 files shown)


fredroy:
[ci-build][with-all-tests]

Sofa/framework/Helper/src/sofa/helper/system/thread/CircularQueue.h
```diff
@@ -151,8 +151,8 @@ class SOFA_HELPER_API OneThreadPerEnd
     template<class T, class OutputIterator>
     unsigned pop(T array[], unsigned maxSize, unsigned maxCapacity, OutputIterator out, unsigned outmaxsize, bool clear = true);
     
-    volatile unsigned head;
-    volatile unsigned tail;
+    std::atomic<unsigned> head;
+    std::atomic<unsigned> tail;
```
bakpaul, line 155: Not fully confident on that

Sofa/GL/src/sofa/gl/VideoRecorderFFMPEG.cpp
```diff
@@ -114,7 +138,7 @@ bool VideoRecorderFFMPEG::init(const std::string& ffmpeg_exec_filepath, const st
        << " -pix_fmt " << codec // yuv420p " // " yuv444p "
        << " -crf 17 "
        << " -vf vflip "
-       << "\"" << m_filename << "\""; // @TODO C++14 : replace with std::quoted
+       << "\"" << escapeForShell(m_filename) << "\"";
```
bakpaul: Yeah ok, but the filename is something we create so this sanity check is maybe not required

Sofa/GUI/Common/src/sofa/gui/common/ArgumentParser.cpp
```diff
@@ -121,12 +127,6 @@ void ArgumentParser::parse()
             }
         }
     }
-
-    // delete argv copy
-    for (int i = 0; i < m_argc; i++) {
-        delete[] copyArgv[i];
```
bakpaul, line 128: Isn't this still required because we still have a vector of char* ?

Sofa/GL/src/sofa/gl/VideoRecorderFFMPEG.cpp
fredroy: One can create a VideoRecorderFFMPEG and then call init() with a crafted filename

Sofa/framework/Helper/src/sofa/helper/system/thread/CircularQueue.h
fredroy: me too but after checking the interweb, apparently volatile does not protect from re-ordering instructions from the CPU (aka memory barrier stuff). But not an expert ✌️

Sofa/GUI/Common/src/sofa/gui/common/ArgumentParser.cpp
fredroy: copyArgv does not "own" what is pointed by the pointers (if that is your question)

Sofa/framework/Helper/src/sofa/helper/system/thread/CircularQueue.h
bakpaul: Yes we gain thread safetyness, but we also loose a bit of performance. But I guess for this kind of structure, the safetyness is of the essence.

Sofa/GUI/Common/src/sofa/gui/common/ArgumentParser.cpp
bakpaul: Sorry I was tired, now I see that the pointer is owned by the string. So all's good.

bakpaul (approved):
Cuda compilation failing, except for that, the PR is fine for me.

epernod:
@fredroy some conflicts on this one

PR #5907, opened 2026-01-29
[Type] Array/Vec: small refresh for modern C++

fredroy:
using std algos, clamp, etc and some keywords noexcept.

and fixed_array_algo, removing some useless templates

and some (generated but modified) unit tests for vec ,as there were almost none ? 🤔

Note that fixed_array_algo is only used in RGBAColor, at least in the standard code base.

[with-all-tests]


fredroy:
[ci-build][with-all-tests]

Sofa/framework/Type/src/sofa/type/Vec.h
```diff
@@ -472,10 +472,7 @@ class Vec
     /// Squared norm.
     constexpr ValueType norm2() const noexcept
     {
-        ValueType r = this->elems[0]*this->elems[0];
-        for (Size i=1; i<N; i++)
-            r += this->elems[i]*this->elems[i];
-        return r;
+        return std::inner_product(elems.begin(), elems.end(), elems.begin(), ValueType{});
```
alxbilger: We can ask the question of the performances...

Sofa/framework/Type/test/VecTypes_test.cpp
```diff
@@ -137,3 +137,415 @@ TEST(VecTest, toVecN)
...
+    EXPECT_EQ(v2[2], 3.0);
+}
+
+TEST(VecTest, moveConstructor)
+{
+    sofa::type::Vec<3, double> v1(1.0, 2.0, 3.0);
+    sofa::type::Vec<3, double> v2(std::move(v1));
+    EXPECT_EQ(v2[0], 1.0);
+    EXPECT_EQ(v2[1], 2.0);
+    EXPECT_NEAR(v2[2], 3.0, 1e-6);
```
th-skam:
```suggestion
    EXPECT_EQ(v2[2], 3.0);
```
If it's a move, it should work. If it's suspicious comparison, then the above should be set to NEAR as well.

Sofa/framework/Type/test/VecTypes_test.cpp
```diff
@@ -137,3 +137,415 @@ TEST(VecTest, toVecN)
...
+
+TEST(VecTest, setFromOtherVec)
+{
+    sofa::type::Vec<3, double> v1(1.0, 2.0, 3.0);
+    sofa::type::Vec<5, double> v;
+    v.set(v1);
+    EXPECT_EQ(v[0], 1.0);
+    EXPECT_EQ(v[1], 2.0);
+    EXPECT_EQ(v[2], 3.0);
+}
```
th-skam, line 217: set here is purposefully tested with a smaller vector?

Sofa/framework/Type/src/sofa/type/Vec.h
fredroy: > We can ask the question of the performances...

Usually people (and IA...) advise to use stl functions because it may help the compiler more (vectorization and stuff). But of course it depends of the implementation of the libstl itself... I will try to do some bench tho

Sofa/framework/Type/test/VecTypes_test.cpp
fredroy: Yes but I can add other set() cases as well

Sofa/framework/Type/src/sofa/type/Vec.h
fredroy: https://quick-bench.com/q/WGEPcgHll4O2cAl8Ms-VoyRBn7I

At -O3, all config (clang/gcc & libstdc++/libc++) give same result. But will less optimization the loop is faster.
Should I roll back the implementation ?

Sofa/framework/Type/src/sofa/type/Vec.h
```diff
@@ -583,10 +580,7 @@ class Vec
     /// sum of all elements of the vector
     constexpr ValueType sum() const noexcept
     {
-        ValueType sum = ValueType(0.0);
-        for (Size i=0; i<N; i++)
-            sum += this->elems[i];
-        return sum;
+        return std::accumulate(elems.begin(), elems.end(), ValueType{});
```
fredroy: ditto

https://quick-bench.com/q/qEv7N1ki6u1g68GI2SLs6UkfL1E

bakpaul (approved):

PR #5908, opened 2026-01-29
[Collision.Geometry] Use StateAccessor and Topology accessor

alxbilger:
This is to factorize the behavior, have the same common variables and expose the links.

[with-all-tests]

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/behavior/SingleStateAccessor.inl
+++ b/Sofa/framework/Core/src/sofa/core/behavior/SingleStateAccessor.inl
@@ -31,15 +31,21 @@ void SingleStateAccessor<DataTypes>::init()
 {
     Inherit1::init();
 
-    d_componentState.setValue(sofa::core::objectmodel::ComponentState::Valid);
+    if (!this->isComponentStateInvalid())
+    {
+        this->validateMState();
+    }
+}
 
+template <class DataTypes>
+void SingleStateAccessor<DataTypes>::validateMState()
+{
     if (!mstate.get())
     {
         mstate.set(dynamic_cast< MechanicalState<DataTypes>* >(getContext()->getMechanicalState()));
 
         if(!mstate)
         {
-
             msg_error() << "No compatible MechanicalState found in the current context. "
                 "This may be because there is no MechanicalState in the local context, "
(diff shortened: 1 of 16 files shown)


alxbilger:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/behavior/SingleStateAccessor.inl
```diff
@@ -31,15 +31,21 @@ void SingleStateAccessor<DataTypes>::init()
 {
     Inherit1::init();
 
-    d_componentState.setValue(sofa::core::objectmodel::ComponentState::Valid);
+    if (!this->isComponentStateInvalid())
+    {
+        this->validateMState();
```
bakpaul, line 36: validateMState() doesn't set the component state to be valid when the mstate exists. We either must change tje validateMSatet method or add a component state modification here

alxbilger: Why does it have to be valid? The only useful information is knowing if the state is invalid or not, but the validity is not important. Specially because this class in designed to be a base class. In this case, we don't have any rule to set the state to valid. The most illustrative example is `BaseObject::init`.

fredroy:
`Build OK. FIXME: 212 units, 447 scenes, 80 regressions` 🤕

Sofa/framework/Core/src/sofa/core/behavior/SingleStateAccessor.inl
bakpaul: Because it was the case before. Your PR is changing this component and fully rely on derived class to set the valid state. 

IMO, we shouldn't expect only the derived class to set valid when it is finally the case, it is too difficult to maintain, we never know if a class is a 'real' leaf or not. With this design we can endup having object that are totally fine, all check passed, but because the last class didn't actually validate it, the object is in an invalid state. What I would advocate for is that all class set to valid as long as its test passes. And let the following class invalidate the state if necessary. This way no matter what is the class place in the inheritance tree, we know that the component state will always be defined. 

But maybe this is more a meta question ?

PR #5909, opened 2026-01-30
[Type] Mat: add unit tests and fix isDiagonal()

fredroy:
While adding some unit tests, I noticed that isDiagonal was wrong 😑
The problem occurs when i=0 (first row). The loop condition is while `j<i-1` but when i=0 it will become `j<std::max<sofa::Size>` because i and j are unsigned.
The fix is also testing if the matrix is rectangular but not sure if we want that 🤔 (like for determinant?)

[with-all-tests]


alxbilger:
Wikipedia says:

The term diagonal matrix may sometimes refer to a rectangular diagonal matrix, which is an m-by-n matrix with all the entries not of the form di,i being zero. For example:

$$
\begin{bmatrix}
1&0&0\\
0&4&0\\
0&0&-3\\
0&0&0\\
\end{bmatrix}
\quad \text{or}\quad 
\begin{bmatrix}
1&0&0&0&0\\
0&4&0&0&0\\
0&0&-3&0&0
\end{bmatrix}
$$

alxbilger (approved):

Sofa/framework/Type/test/MatTypes_test.cpp
```diff
@@ -443,3 +443,136 @@ TEST(MatTypesTest, determinant3x3)
...
+
+//TEST(MatTypesTest, determinantRectangular)
+//{
+//    const Mat<2,3,double> m232{{1., 2., 3.}, {4., 5., 6.}};
+//    EXPECT_DOUBLE_EQ(sofa::type::determinant(m232), -3.);
+//
+//    const Mat<3,2,double> m322{{1., 2.}, {3., 4.}, {5., 6.}};
+//    EXPECT_DOUBLE_EQ(sofa::type::determinant(m322), -12.);
+//}
+
```
hugtalbot, line 455: Shall we keep it the determinant for non-square matrices @alxbilger?

alxbilger: https://github.com/sofa-framework/sofa/pull/5876

Sofa/framework/Type/src/sofa/type/Mat.h
```diff
@@ -528,12 +528,14 @@ class Mat
...
-        for (Size i=0; i<L; i++)
+        constexpr Size minDim = std::min(L, C);
+        for (Size i=0; i<L; ++i)
         {
-            for (Size j=0; j<i-1; j++)
-                if( rabs( (*this)(i,j) ) > EQUALITY_THRESHOLD ) return false;
-            for (Size j=i+1; j<C; j++)
+            for (Size j=0; j<C; ++j)
+            {
+                if (j == i && i < minDim) continue;
```
bakpaul: I don't think that `&& i < minDim` is required because once one of the two indices is higher than the minDim, they can never be equal.

bakpaul (approved):

fredroy:
[ci-build][with-all-tests]

PR #5915, opened 2026-02-02
[Config] Runs clang-tidy on specific targets during the build

alxbilger:
Unfortunately, I couldn't test it on MSVC. According to [the documentation](https://cmake.org/cmake/help/latest/prop_tgt/LANG_CLANG_TIDY.html), this option is available only for makefile and ninja generators. Therefore, I couldn't test if this option is useful or not. The tool can be configured with `.clang-tidy` file that can be placed at the root of the project.


bakpaul:
I would prefere to have an option `SOFA_ENABLES_CLANG-TIDY` that is off by default and can be put to ON.

alxbilger:
I made this PR just to illustrate that this tool exist. Since it is not portable, I prefer not to merge it.

PR #5916, opened 2026-02-02
[CI] CMake & Nix: add tight_inclusion

nim65s:
This is a quick and dirty temporary workaround, following discussions with @hugtalbot,
but proper packaging should really be contributed upstream.

Do you have plans for this @olivier-roussel @bakpaul ?

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,7 +28,8 @@
           devShells.default = pkgs.mkShell { inputsFrom = [ self'.packages.default ]; };
           packages = {
             default = self'.packages.sofa;
-            sofa = pkgs.callPackage ./package.nix { };
+            sofa = pkgs.callPackage ./package.nix { inherit (self'.packages) tight_inclusion; };
+            tight_inclusion = pkgs.callPackage ./tight_inclusion.nix { };
           };
         };
     };
--- a/package.nix
+++ b/package.nix
@@ -12,6 +12,7 @@
   metis,
   nlohmann_json,
   stdenv,
+  tight_inclusion,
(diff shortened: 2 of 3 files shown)


bakpaul:
Hello ! Yes there is a fork of us with patches fixing packaging (this has been made for Conda) 

The patch list is here : https://github.com/conda-forge/tight-inclusion-feedstock/tree/main/recipe/patches 

We plan to make PR on the original fork.

olivier-roussel:
The introduced CMake find module for tight inclusion will be used instead of the CMake config file, such as the one provided by the conda package in our Pixi build chain.
We should either find a way to give priority to CMake Config file (which I thought to be the case in modern CMake, but I tested and it was not), either this CMake module file should be adapted to follow our general needs, such as defining targets and install properties (see https://github.com/sofa-framework/sofa/blob/master/cmake/Modules/FindTinyXML2.cmake for example). They should also support DLL properly for installation on Windows in our releases (https://github.com/olivier-roussel/sofa/blob/a8547ce0cb53576927efe84379b4829dd27d2a75/cmake/Modules/FindTinyXML2.cmake), but it's out of the scope of this PR and it will be handled in #5252 (for tracking).

nim65s:
I'll workaround that by using directly your conda patches in nix, but clearly the good solution is to contribute the cmake exports upstream

olivier-roussel:
Thanks for the update. Yes, we usually fix upstream for all patches we have to make for conda packages and here we've plan to do it as well asap. We'll let you know once done.

PR #5924, opened 2026-02-03
[Helper] Match ubuntu implementation of create dir for Windows

bakpaul:
Before : On windows, if the folder creation fails because the folder already exists then an error is thrown. Which is not the case on ubuntu
Now: If the path already exists and is indeed a folder then everything is fine.  

[with-all-tests]

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #5929, opened 2026-02-04
[Helper] FileSystem::createDirectory:  fix missing return warning

fredroy:
I guess it was not implemented because it should never happen (the error would be detected at the stat() stage)

My beef is more about the fact that createDirectory() return **true** if there is an error 🤔
which is counter-intuitive ; I would expect return **false** in case of an error... 🫠


bakpaul:
Linked to this : https://github.com/sofa-framework/sofa/pull/5924

epernod (approved):

bakpaul (approved):

PR #5943, opened 2026-02-06
[Mapping] Fix rigid mapping init 

bakpaul:
Use right vector size for checking. This fixes a crash when the mapping was used with only "useX0=True". 

I know this component needs way more fixing/refactoring. But this is just a fix keeping the ols implementation untouched for now. 


Sofa/Component/Mapping/NonLinear/src/sofa/component/mapping/nonlinear/RigidMapping.inl
```diff
@@ -135,7 +135,8 @@ template <class TIn, class TOut>
 sofa::Index RigidMapping<TIn, TOut>::getRigidIndex(sofa::Index pointIndex ) const
 {
     // do we really need this crap?
```
hugtalbot: Possibly ... no?
```suggestion
```

bakpaul: This comment is from 2015 and the following code has been modified multiple times afterwards. So I would say that we might need it.

alxbilger (approved):

hugtalbot (approved):

PR #5963, opened 2026-02-24
[LinearSystem] Update the "M, B, K factors" of the linear system when necessary

leobois67:
Problem: calling `mop.setSystemMBKMatrix(mFact, bFact, kFact, l_linearSolver.get());` actually ignores the factors except for the first time it is called.

Use case: In order to compute the gradient of a loss wrt some parameters, I want to solve linear systems with my own "M, B, K factors".

Solution: check for a change in the factors and update if necessary. This update is done in `MatrixLinearSystem<TMatrix, TVector>::associateLocalMatrixTo()`.

Details: The factors passed to `setSystemMBKMatrix()` follow the path below:
```
MechanicalOperations::setSystemMBKMatrix()
> BaseMatrixLinearSystem::buildSystemMatrix()
    > BaseMatrixLinearSystem::preAssembleSystem() [virtual]
    > TypedMatrixLinearSystem<TMatrix, TVector>::preAssembleSystem()
        > TypedMatrixLinearSystem<TMatrix, TVector>::associateLocalMatrixToComponents() [virtual]
        > MatrixLinearSystem<TMatrix, TVector>::associateLocalMatrixToComponents()
            > MatrixLinearSystem<TMatrix, TVector>::associateLocalMatrixTo()
```
In this final method, the factors are only considered the first time the system is assembled because of this check:
```
auto it = componentLocalMatrix.find(pairs);
if (it == componentLocalMatrix.end())
{
    // use M, B, K factors
}
```
Which is why, after this check, we add the following code:
```
if (std::abs(localMatrix->m_cachedFactor - factor) > std::numeric_limits<SReal>::epsilon())
{
    localMatrix->setFactor(factor);
}
```


leobois67:
Also: now that I understand better how things work, I may actually be able to do what I want without this PR ;)

alxbilger:
Your analysis makes sense. But the solution is wrong. `m_cachedFactor` is protected, therefore a direct access leads to a compilation error.

Sofa/Component/LinearSystem/src/sofa/component/linearsystem/MatrixLinearSystem.inl
```diff
@@ -863,6 +863,11 @@ void MatrixLinearSystem<TMatrix, TVector>::associateLocalMatrixTo(
             dmsg_fatal() << "Local matrix is invalid";
         }
 
+        if (std::abs(localMatrix->m_cachedFactor - factor) > std::numeric_limits<SReal>::epsilon())
+        {
+            localMatrix->setFactor(factor);
+        }
```
alxbilger:
```suggestion
        localMatrix->setFactor(factor);
```
I suggest to always set the factor even if it does not change since last time.

leobois67:
Yep, that is what we did before adding the condition without checking that it compiled... I made the change.

bakpaul (approved):

alxbilger (approved):

PR #5969, opened 2026-02-25
[Collision] Set Data & Links to public in Collision models. Adopt state accessor for mstate

th-skam:
A bit of an inconsistency since the topology link (and Data) are mostly public in other model classes. Same with Sphere coli model.
Was this by design? Let's keep one consistent style for all models.
What do you think? should they be public or private?


alxbilger (commented):
Yes, Links and Data should be public. In any case, they can be accessed publicly (for example using `findData`, or in Python), so it makes sense that they are also public in C++.

As a side note, `mstate` should be a link too.

th-skam:
Ended up with a bit of cleaning:

- Data and Links were moved to public
- topology pointer and Link coexisting in `LineCollisionModel`
- Use SingleStateAccessor for the mstate

fredroy (approved):

bakpaul (approved):

PR #5961, opened 2026-02-24
[SofaCUDA] Call register functions in the plugin initialization function

fredroy:
and cleanups in the config/init

I can see that nobody seems to be using SofaCUDA these days 🫢 
as it was unusable since 
- #5827 

so SofaCUDA might be not usable in the v25.12 release 💀 

[with-all-tests]
______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #5973, opened 2026-02-26
[SofaCUDA] SofaCUDA is now a meta-plugin of two plugins: SofaCUDA.Core and SofaCUDA.Component

fredroy:
the big boi of these PRs

The main point of this PR was to split the tools/core stuff of SofaCUDA and the different component (either specializations or just new instanciations) (+ the GUI done in a previous PR)

A nice side-effect is that plugins depending on SofaCUDA and implementing new components of whatever can now just depends on SofaCUDA.Core.

Screenshot 2026-02-26 at 11 07 47

There are still some confusions such as 
 - the namespaces and directories: I kept the old sofa::gpu::cuda for the core ones for not breaking too much. But I can do it 🤫 
 - uppercase/lowercase for the (ported) includes directories. This is a plugin so the (non-official) guidelines should be "SofaCUDA/Core/blabla.h" and "SofaCUDA/Component/blabla.h", but this is emulating Sofa so it could be "sofacuda/core/blabla.h" and "sofacuda/component/blabla.h" . In the current situation, I kept the mix of the two "SofaCUDA/core" and "SofaCUDA/component" because   "SofaCUDA/component"  was already used... 🤪

It is breaking obviously, but the runtime behavior *should* be preserved (i.e compat with the scenes)


bakpaul:
So I tried multiple scenes. the python ones you provided work. In the examples some worked and some didn't, so I checkout only [this PR](https://github.com/sofa-framework/sofa/pull/5961/changes) to verify if we have regression or not. And it seems to be no regression.

bakpaul (approved):

PR #5976, opened 2026-02-26
[Visual] Scene center and radius are not longer cached

alxbilger:
This PR removes the cache mechanism around the scene center and radius in the `BaseCamera` class.

Removed:
- Data member `sceneCenter` and `sceneRadius` are removed. This is breaking for derived class using them.

Added:
- Methods to get the scene center and radius when needed.

Changes:
- The cached scene center and radius are replaced by direct calls to the newly added methods.

Effects:
- If fixes a bug producing nans in the projection matrix of the camera. The cached value `sceneCenter` was used, but it was not already computed and cached. Therefore, the default value was used (likely zero). The default value was then used to compute Znear and Zfar, resulting both to be set to zero. At some point there is a division by `(currentZFar - currentZNear)`, leading to a nan.

Before:
image

After:
image

Note that I don't understand why having nan is not problematic. I am not sure neither if having valid values helps.

I took the opportunity to add easy constness to some methods.


fredroy (approved):

Lucas-TJ (approved):

alxbilger:

PR #5987, opened 2026-03-03
[Geometry] Bugfixes and add unit tests

fredroy:
Fix various Edge and triangles functions
- intersectionWith* (test against abs value to check both sides) and collinearity
- fix 2d and 3d cases
- isPointInTriangle: add checks if the point is in the triangle plane but also add a boolean to bypass this check (as it was before the "fix")
- mat_LCP (used for proximity):  the solver returned true even it fails to converge (max iterations reached)

and more unit tests for Sofa.Geometry module

[with-all-tests]

### DIFF
--- a/Sofa/framework/Geometry/src/sofa/geometry/Edge.h
+++ b/Sofa/framework/Geometry/src/sofa/geometry/Edge.h
@@ -307,15 +307,19 @@ struct Edge
             const T alphaNom = AC[1] * CD[0] - AC[0] * CD[1];
             const T alphaDenom = AB[1] * CD[0] - AB[0] * CD[1];
 
-            if (alphaDenom < std::numeric_limits<T>::epsilon()) // collinear
+            if (std::abs(alphaDenom) < std::numeric_limits<T>::epsilon()) // collinear
             {
                 intersectionBaryCoord = sofa::type::Vec<2, T>(0, 0);
                 return false;
             }
 
             const T alpha = alphaNom / alphaDenom;
+            // beta: pC + beta * CD = pA + alpha * AB
+            const T beta = (std::abs(CD[0]) > std::abs(CD[1]))
+                ? (AC[0] + alpha * AB[0]) / CD[0]
+                : (AC[1] + alpha * AB[1]) / CD[1];
 
-            if (alpha < 0 || alpha > 1)
+            if (alpha < 0 || alpha > 1 || beta < 0 || beta > 1)
             {
                 intersectionBaryCoord = sofa::type::Vec<2, T>(0, 0);
                 return false;
--- a/Sofa/framework/Geometry/src/sofa/geometry/Edge.h
+++ b/Sofa/framework/Geometry/src/sofa/geometry/Edge.h
@@ -197,7 +197,7 @@ struct Edge
 
         //compute intersection between line and plane equation
         const T denominator = planeNorm * (n1 - n0);
-        if (denominator < EQUALITY_THRESHOLD)
+        if (std::abs(denominator) < EQUALITY_THRESHOLD)
         {
             return false;
         }
--- a/Sofa/framework/Geometry/src/sofa/geometry/Edge.h
+++ b/Sofa/framework/Geometry/src/sofa/geometry/Edge.h
@@ -165,7 +165,7 @@ struct Edge
             N2 = sofa::type::cross(AB, AC).norm2();
         }
 
-        if (N2 > EQUALITY_THRESHOLD || N2 < 0)
+        if (std::abs(N2) > EQUALITY_THRESHOLD)
             return false;
         
         return true;
--- a/Sofa/framework/Geometry/src/sofa/geometry/Edge.h
+++ b/Sofa/framework/Geometry/src/sofa/geometry/Edge.h
@@ -316,8 +316,8 @@ struct Edge
             const T alpha = alphaNom / alphaDenom;
             // beta: pC + beta * CD = pA + alpha * AB
             const T beta = (std::abs(CD[0]) > std::abs(CD[1]))
-                ? (AC[0] + alpha * AB[0]) / CD[0]
-                : (AC[1] + alpha * AB[1]) / CD[1];
+                ? (-AC[0] + alpha * AB[0]) / CD[0]
+                : (-AC[1] + alpha * AB[1]) / CD[1];
 
             if (alpha < 0 || alpha > 1 || beta < 0 || beta > 1)
             {
--- a/Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
+++ b/Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
@@ -88,26 +88,54 @@ struct Triangle
         static constexpr auto getBarycentricCoordinates(const Node& p0, const Node& n0, const Node& n1, const Node& n2)
     {
         // Point can be written: p0 = a*n0 + b*n1 + c*n2
-        // with a = area(n1n2p0)/area(n0n1n2), b = area(n0n2p0)/area(n0n1n2) and c = area(n0n1p0)/area(n0n1n2) 
-        const auto area = Triangle::area(n0, n1, n2);
-        if (fabs(area) < std::numeric_limits<T>::epsilon()) // triangle is flat
+        // with a = area(n1n2p0)/area(n0n1n2), b = area(n0n2p0)/area(n0n1n2) and c = area(n0n1p0)/area(n0n1n2)
+        if constexpr (std::is_same_v<Node, sofa::type::Vec<3, T>>)
         {
-            return sofa::type::Vec<3, T>(-1, -1, -1);
-        }
-        
-        const auto A0 = Triangle::area(n1, n2, p0);
-        const auto A1 = Triangle::area(n0, p0, n2);
+            // In 3D, use signed areas via dot product with triangle normal
+            // to get correct sign for outside-triangle points
+            const auto N = sofa::type::cross(n1 - n0, n2 - n0);
+            const auto NdotN = sofa::type::dot(N, N);
 
-        sofa::type::Vec<3, T> baryCoefs(type::NOINIT);
-        baryCoefs[0] = A0 / area;
-        baryCoefs[1] = A1 / area;
-        baryCoefs[2] = 1 - baryCoefs[0] - baryCoefs[1];
+            if (NdotN < std::numeric_limits<T>::epsilon() * std::numeric_limits<T>::epsilon()) // triangle is flat
+            {
+                return sofa::type::Vec<3, T>(-1, -1, -1);
+            }
 
-        if (fabs(baryCoefs[2]) <= std::numeric_limits<T>::epsilon()){
-            baryCoefs[2] = 0;
+            sofa::type::Vec<3, T> baryCoefs(type::NOINIT);
+            baryCoefs[0] = sofa::type::dot(N, sofa::type::cross(n2 - n1, p0 - n1)) / NdotN;
+            baryCoefs[1] = sofa::type::dot(N, sofa::type::cross(n0 - n2, p0 - n2)) / NdotN;
+            baryCoefs[2] = 1 - baryCoefs[0] - baryCoefs[1];
+
+            if (fabs(baryCoefs[2]) <= std::numeric_limits<T>::epsilon()){
+                baryCoefs[2] = 0;
+            }
+
+            return baryCoefs;
         }
-        
-        return baryCoefs;
+        else
+        {
+            // In 2D, Triangle::area() returns a signed value (shoelace formula),
+            // so the ratio of sub-areas to total area preserves correct signs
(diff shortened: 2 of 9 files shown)


fredroy:
[ci-build][with-all-tests]

Sofa/framework/Geometry/src/sofa/geometry/Edge.h
bakpaul, line 153: Funny that the AI didn't catch this, but here because AB is not normalized the check is wrong. 
```suggestion
        if (Ln0p0 < 0 || Ln0p0 > std::pow(Ln0n1,2) ) // out of bounds [n0; n1]
```

Sofa/framework/Geometry/src/sofa/geometry/Edge.h
```diff
@@ -165,7 +165,7 @@ struct Edge
             N2 = sofa::type::cross(AB, AC).norm2();
         }
 
-        if (N2 > EQUALITY_THRESHOLD || N2 < 0)
+        if (std::abs(N2) > EQUALITY_THRESHOLD)
```
bakpaul, line 168: Ok, the original code didn't deal properly with the case of a 2D segment with the point being under the edge instead of over it.  This is correct now.

Sofa/framework/Geometry/src/sofa/geometry/Edge.h
```diff
@@ -197,7 +197,7 @@ struct Edge
 
         //compute intersection between line and plane equation
         const T denominator = planeNorm * (n1 - n0);
-        if (denominator < EQUALITY_THRESHOLD)
+        if (std::abs(denominator) < EQUALITY_THRESHOLD)
```
bakpaul, line 200: Ok fine, when n0 was further away to the origin of the world, the old check would always return false.  This is correct.

Sofa/framework/Geometry/src/sofa/geometry/Edge.h
```diff
@@ -307,15 +307,19 @@ struct Edge
...
             {
                 intersectionBaryCoord = sofa::type::Vec<2, T>(0, 0);
                 return false;
             }
 
             const T alpha = alphaNom / alphaDenom;
+            // beta: pC + beta * CD = pA + alpha * AB
+            const T beta = (std::abs(CD[0]) > std::abs(CD[1]))
+                ? (AC[0] + alpha * AB[0]) / CD[0]
+                : (AC[1] + alpha * AB[1]) / CD[1];
```
bakpaul:
```suggestion
                ? (-AC[0] + alpha * AB[0]) / CD[0]
                : (-AC[1] + alpha * AB[1]) / CD[1];
```
Important to check the arithmetic

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
```diff
@@ -88,26 +88,54 @@ struct Triangle
...
+            sofa::type::Vec<3, T> baryCoefs(type::NOINIT);
+            baryCoefs[0] = sofa::type::dot(N, sofa::type::cross(n2 - n1, p0 - n1)) / NdotN;
+            baryCoefs[1] = sofa::type::dot(N, sofa::type::cross(n0 - n2, p0 - n2)) / NdotN;
+            baryCoefs[2] = 1 - baryCoefs[0] - baryCoefs[1];
+
+            if (fabs(baryCoefs[2]) <= std::numeric_limits<T>::epsilon()){
+                baryCoefs[2] = 0;
+            }
+
+            return baryCoefs;
```
bakpaul, line 113: Ok, before this the barycentric mapping was a bit wrong when the point wasn't on the triangle. Because of 104, the value of the third barycentric coordinate was always underestimated because the other ones where overestimated.

bakpaul: @epernod this fix the cases when the point is not on the same plane as the triangle. But does it ever happen ? Anyway I think that the api should not expect anything and check because otherwise we would rely on the user to make the check before.

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
```diff
@@ -163,10 +191,30 @@ struct Triangle
...
+                const auto normal = Triangle::normal(n0, n1, n2);
+                const auto normalNorm2 = sofa::type::dot(normal, normal);
+                if (normalNorm2 > std::numeric_limits<T>::epsilon())
+                {
+                    const auto d = sofa::type::dot(p0 - n0, normal);
+                    if (d * d / normalNorm2 > std::numeric_limits<T>::epsilon())
+                        return false;
+                }
+            }
+        }
```
bakpaul, line 243: I would put this code in a function "isPointOnPlane", it might be useful.

Sofa/framework/Geometry/src/sofa/geometry/Edge.h
fredroy: but `Ln0n1 = sofa::type::dot(AB, AB);` so I dont see why you do `std::pow(Ln0n1,2)`....

bakpaul: Because I messed everything up in my calculation. I am not used to work with not normalized vector, it is way easier when everything is normalized ! So yeah, it is normal that the AI didn't catch this, as it is correct.

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
```diff
@@ -148,25 +176,75 @@ struct Triangle
...
+    * @param assumePointIsOnPlane: optional bool to avoid testing if the point is on the plane defined by the triangle
     * @return	bool result if point is inside Triangle.
     */
     template<typename Node,
         typename T = std::decay_t<decltype(*std::begin(std::declval<Node>()))>,
         typename = std::enable_if_t<std::is_scalar_v<T>>
     >
-        static constexpr bool isPointInTriangle(const Node& p0, const Node& n0, const Node& n1, const Node& n2, sofa::type::Vec<3, T>& baryCoefs)
+    [[nodiscard]]
+    static constexpr bool isPointInTriangle(const Node& p0, const Node& n0, const Node& n1, const Node& n2, sofa::type::Vec<3, T>& baryCoefs, bool assumePointIsOnPlane = true)
```
epernod, line 230: shouldn't be the default value of assumePointIsOnPlane  to false to always test it by default?

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
```diff
@@ -148,25 +176,75 @@ struct Triangle
...
     {
         baryCoefs = Triangle::getBarycentricCoordinates(p0, n0, n1, n2);
```
epernod: I would call this computation after the if/else to avoid some unnecessary calculation if point is not on plane. and init the baryCoef to same invalid value sofa::type::Vec<3, T>(-1, -1, -1);

aha you didn't see this one Claude!

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
epernod: yes good catch I wrote all those functions as helper methods taking into account user know what they do but I agree to add all those isInPlane safeguards.

However here, in the Triangle::getBarycentricCoordinates method, if the point is not in a triangle, I think we should not do a projection of this point and send some valid values. 
For me you should just call at start your new method: isPointOnPlane  and if it is false you exit and return incorrect value like return sofa::type::Vec<3, T>(-1, -1, -1); for flat triangle.

epernod: you could also add a bool& isInPlane input parameter, that could be useful, but it means changing the signature of the method

Sofa/framework/Geometry/test/Triangle_test.cpp
```diff
@@ -247,21 +247,29 @@ TYPED_TEST(GeometryVec3DTriangle_test, isPointInTriangle)
...
     p0 = { 1., 0.2, 0.2 };
-    res = sofa::geometry::Triangle::isPointInTriangle(p0, a, b, c, bary);
+    res = sofa::geometry::Triangle::isPointInTriangle(p0, a, b, c, bary, false);
     EXPECT_FALSE(res);
-    EXPECT_NEAR(bary[0], 0.69282, 1e-4);
-    EXPECT_NEAR(bary[1], 0.360555, 1e-4);
-    EXPECT_NEAR(bary[2], -0.0533754, 1e-4);
+    EXPECT_NEAR(bary[0], 0.6, 1e-4);
+    EXPECT_NEAR(bary[1], 0.3, 1e-4);
+    EXPECT_NEAR(bary[2], 0.1, 1e-4);
```
epernod: to be updated to equal -1 -1 -1

Sofa/framework/Geometry/test/Triangle_test.cpp
```diff
@@ -247,21 +247,29 @@ TYPED_TEST(GeometryVec3DTriangle_test, isPointInTriangle)
...
+
+    // out of plan with assumePointIsInPlane=true (default): bary coords are all positive
+    // so the function returns true even though the point is not in the triangle plane
+    res = sofa::geometry::Triangle::isPointInTriangle(p0, a, b, c, bary);
+    EXPECT_TRUE(res);
+    EXPECT_NEAR(bary[0], 0.6, 1e-4);
+    EXPECT_NEAR(bary[1], 0.3, 1e-4);
+    EXPECT_NEAR(bary[2], 0.1, 1e-4);
```
epernod, line 264: for me this should not be allowed

Sofa/framework/Geometry/test/Triangle_test.cpp
```diff
@@ -247,21 +247,29 @@ TYPED_TEST(GeometryVec3DTriangle_test, isPointInTriangle)
...
 
     // in plan but out of triangle
     p0 = { 2., 2., 2. };
     res = sofa::geometry::Triangle::isPointInTriangle(p0, a, b, c, bary);
     EXPECT_FALSE(res);
-    EXPECT_EQ(bary[0], 1);
+    EXPECT_EQ(bary[0], -1);
     EXPECT_EQ(bary[1], 1);
-    EXPECT_EQ(bary[2], -1);
+    EXPECT_EQ(bary[2], 1);
```
epernod, line 272: this change is strange but ok if it passes

epernod (changes requested):
didn't check all the tests but the safeguard on checking if point is on a plane is indeed a good practice. However in this file as we request value regarding the triangle, if the point is out of the plane we should always return invalid value in my opinion

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
fredroy: As it was not tested before, setting to false would potentially decrease the perfs automatically.

An other solution would be to set to false, and adding true to all existing call to isPointInTriangle(). But this would not cover the usage in external projects...
Last solution would be: no bool and two functions (isPointInTriangle and fastIsPointInTriangle or something like that)

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
fredroy: Not sure about the best way if not on plane: setting the {-1, -1, -1 } as you suggest ; or do not touch the Vec3 as it will return false, and it is the caller's responsibility to manage this case

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
fredroy: > you could also add a bool& isInPlane input parameter, that could be useful, but it means changing the signature of the method

I would just assume that if it returns false, the caller (if he wants) would investigate on the reason it is false (either by calling isOnPlane itself, or checking the resulting coeffs)

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
epernod: yes maybe the last solution is better to avoid any missunderstanding.
a 2nd method isPointInTriangleNoPlaneCheck 😆

Sofa/framework/Geometry/src/sofa/geometry/Triangle.h
epernod: the problem here is getBarycentricCoordinates  return the vec3 of barycoefs, not a bool. That's why I was speaking of having a additional bool& parameter.
Maybe changing the method signature to take the vec3 as input/output and returning a bool would be more efficient to avoid checking against -1 -1 -1 value when using the method.

bakpaul: The issue with what you propose Erik is that it will change the behavior of the API. Wasn't this previous "projection-like" mechanism useful in some cases where people used a barycentric mapping of points on a triangular mesh without making sure it was actually on the mesh ? Anyway, I expect great regression with this change. I would do it in another PR.

epernod (approved):

PR #6036, opened 2026-03-19
[Helper] ScopedAdvancedTimer: add a compile-time check on string argument in macros

fredroy:
This will ensure that the developer will use the good macro, even if it is not compiled with Tracy.

The diagnostic message is not the best but I think it is good enough (c)

Screenshot 2026-03-19 at 09 15 23

And breaking because it might break external plugins using these macro.

(NOTE: i wanted to do first a consteval lambda to avoid declaring a function but the compilers on the CI are too "old" 👴 )


hugtalbot (approved):
Approved at SOFA dev

PR #6043, opened 2026-03-19
[SofaMatrix] Try loading extensions based on the GUI

alxbilger:


fredroy (commented):
I guess this will work only if the GUI plugins are loaded *before* this one

bakpaul:
Given that the order of occurrence in `plugin_list.conf.default` is dealt by CMake, it might sometimes work and sometimes not with the default loading. Could it be possible to make a mecanisms like a bakward init  for plugins ?After everything is loaded, we load the extensions of every plugin depending on what is loaded

PR #6073, opened 2026-04-08
[Diffusion] Explicit MechanicalObject link

GaetanDesrues:
Adding an explicit link to the MechanicalObject instead of relying on tags


fredroy (approved):
I see `GaetanDesrues:v25.12` -> `sofa-framework:v25.12`
I think this PR should be merged into master, and might be backported to v25.12 later (or not)

bakpaul:
Hello, all ForceFields are SingleStateAccessor, which are member called `mstate` and already are a link to the mechanical state. Can you check if this fits your need ?

GaetanDesrues:
Hello @hugtalbot, yes PCG commit should not be in that PR, my mistake. I'll recreate both PRs in master.
Regarding the explicit link to the MechanicalObject, it's not targeted at the node's MechanicalObject like mstate, but rather at any MechanicalObject in the scene. It was previously addressed with tags, I'm just adding an explicit link which is cleaner than hard-coded tags.

alxbilger:
@GaetanDesrues what Hugo meant, is that `TetrahedronDiffusionFEMForceField` already has an explicit link to a mstate through its inheritance (base class `SingleStateAccessor`). Calling it with `this->mstate` and removing all the tag-based search should be sufficient.

PR #6079, opened 2026-04-13
[SofaCUDA] CMake: Remove old options

fredroy:
First intention was to fix the cmake configure process with cuBLAS activated (it was missing the `PUBLIC` keyword)

But then cuBLAS and THRUST are fully integrated into CudaToolkit, which is a requirement for SofaCUDA ; and I guess it came from the time it was not supported by all the GPUs
So IMO it should always be enabled.

Same for double, but I am not 100% sure about it, as one may not want the double version maybe.


alxbilger (approved):

bakpaul:
I would prefer to leave the possibility to deactivate the DOUBLE compilation. So IMO, let's keep the option but put it to on by default.

bakpaul (approved):

PR #6089, opened 2026-04-22
[Example] Convert Torusfall XML scene to SofaPython3 scene

fredroy:
I wanted to test different solver or model for each torus but it was a pain in the xml file to apply to all the toruses.
Moreover it was quite unreadable because of the length of the scene, the long strings, etc.

So I ask Claude to rewrite and refactor in python

The cylinders part can be customized (increasing/decreasing the numbers). 
I wanted to do the same with the toruses but it was a pain to find a programmatically way to compute them so I let as it is for now.

[with-all-tests]


bakpaul:
[ci-build][with-all-tests]

bakpaul:
To add it to the perf tests, we first need to enable running python scripts, which is not an issue. And then we will only need to add the filename inside the .performance-tests file inside of the repo. I'll do it one this PR is ready to be merged.

PR #6090, opened 2026-04-22
[Visual] draw more lines for LineAxis and VisualGrid

EulalieCoevoet:
**Problem**
When using gl equations that are evaluated at vertices only (ex: GL_FOG), drawing large `LineAxis` and `VisualGrid` produces artefacts. 

**In this PR**
Adds additional vertices and lines when drawing `LineAxis` and `VisualGrid`, instead of only two points (start and end of lines). I chose 6 lines, which is arbitrary but sufficient for my use case.

**Example of results (using GL_FOG)**

https://github.com/user-attachments/assets/0f233109-70bc-42f0-b233-bbea77c3f92f

**Artefacts when drawing only two points**

https://github.com/user-attachments/assets/1d2a0968-5b7d-4ca5-883e-32c4d23b80fd


bakpaul:
Hello ! 

Because those two objects are components, we would prefer this number to be added as a data. Plus, maybe a video or picture of the artifact to better understand the issue would be good. 

Thank you !

EulalieCoevoet:
Thanks for the review. I have just added a video showing the artefacts. 
I'll add the data as well. Should I set 1 or 6 as the default value? I vote for keeping it as is, meaning 1. What do you think?

alxbilger (approved):

bakpaul:
I agree for 1 as default !

EulalieCoevoet:
@bakpaul I did all the changes. Thanks again for the review!

bakpaul (approved):

PR #6095, opened 2026-04-24
[Helper] FileSystem: add more unittests

fredroy:
Should test everything except `openFileWithDefaultApplication`

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #6096, opened 2026-04-24
[Helper] FileSystem: refactor using std::filesystem

fredroy:
Based on
- #6095 

Title says all.
`computeParentDirectory` and `stripDirectory` have quite different behaviors than their STL counterparts (See the comments) so it still using the previous string-based implementation.
Note: different LLM (Qwen) than for the tests (Claude) 🤖 for fairness

[with-all-tests]

### DIFF
--- a/Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
+++ b/Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
@@ -24,45 +24,38 @@
 #include <sofa/helper/StringUtils.h>
 #include <sofa/helper/Utils.h>
 
-#if __has_include(<filesystem>)
-  #include <filesystem>
-  namespace fs = std::filesystem;
-#elif __has_include(<experimental/filesystem>)
-  #include <experimental/filesystem> 
-  namespace fs = std::experimental::filesystem;
-#else
-  error "Missing the <filesystem> header."
-#endif
-
+#include <filesystem>
 #include <fstream>
 #include <iostream>
 #ifdef WIN32
 # include <windows.h>
 # include <winerror.h>
 # include <strsafe.h>
-# include "Shlwapi.h"           // for PathFileExists()
-#include <shellapi.h>
+# include "Shlwapi.h"
+# include <shellapi.h>
 #else
 # include <dirent.h>
 # include <sys/stat.h>
 # include <sys/types.h>
 # include <errno.h>
-# include <string.h>            // for strerror()
+# include <string.h>
 # include <unistd.h>
 #endif
 
 #if defined(__APPLE__)
-#include <stdio.h>
-#include <spawn.h>
+# include <stdio.h>
+# include <spawn.h>
 #endif
 
 #ifdef linux
-#include <spawn.h>
-#include <sys/wait.h>
+# include <spawn.h>
+# include <sys/wait.h>
 #endif
 
 #include <cassert>
 #include <sofa/helper/system/SetDirectory.h>
+
+namespace fs = std::filesystem;
 
 namespace sofa
 {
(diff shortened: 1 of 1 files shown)


epernod:
aha I didn't saw this one. I was testing my Qwen setup exactly on the same class...

bakpaul:
[ci-build][with-all-tests]

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
```diff
@@ -81,284 +114,132 @@ std::string FileSystem::getExtension(const std::string& filename)
...
 #else
     if (rmdir(path.c_str()))
     {
-        msg_error("FileSystem::removedirectory()") << path << ": " << strerror(errno);
+        msg_error("FileSystem::removeDirectory()") << path << ": " << strerror(errno);
         return true;
     }
 #endif
     return false;
 }
```
bakpaul, line 183: Should use std::filesystem::remove_all instead

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
```diff
@@ -81,284 +114,132 @@ std::string FileSystem::getExtension(const std::string& filename)
...
 
+bool FileSystem::removeAll(const std::string& path){
```
bakpaul, line 185: Ok I don't get why adding this instead of modifying the previous one which is meaningless now.

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
```diff
@@ -81,284 +114,132 @@ std::string FileSystem::getExtension(const std::string& filename)
...
+        return true ;
+    }
+    catch(const fs::filesystem_error& /*e*/)
+    {
+        return false ;
+    }
+}
 
-bool FileSystem::exists(const std::string& path, [[maybe_unused]] bool quiet)
+bool FileSystem::removeFile(const std::string& path)
```
bakpaul, line 191: Ok so basically it is wrapping all call to std with try catch, do we want that ?

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
```diff
@@ -561,7 +470,7 @@ bool FileSystem::openFileWithDefaultApplication(const std::string& filename)
         if ((INT_PTR)ShellExecuteA(nullptr, "open", filename.c_str(), nullptr, nullptr, SW_SHOWNORMAL) > 32)
             success = true;
 #elif defined(__APPLE__)
-        pid_t pid; // points to a buffer that is used to return the process ID of the new child process.
```
bakpaul, line 564: Why ?

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
```diff
@@ -551,7 +460,7 @@ bool FileSystem::openFileWithDefaultApplication(const std::string& filename)
 
     if (!filename.empty())
     {
-        if (!FileSystem::exists(filename))
+        if (!fs::exists(filename))
```
bakpaul: Why not use the previously created FileSystem::exists ?

fredroy: - it will print twice the message that it does not exist
- IMO we should use the fs::xxxx() as much as possible

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
fredroy: if you look at the comment in the header :
- removeDirectory = deleting an *empty* directory (like `rmdir <dir>`)
- removeAll = deleting a *non-empty* directory (like `rm -rf <dir>`)

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
bakpaul: Ok !

bakpaul (approved):

Sofa/framework/Helper/src/sofa/helper/system/FileSystem.cpp
```diff
@@ -81,284 +114,117 @@ std::string FileSystem::getExtension(const std::string& filename)
...
-            && path[1] == ':';
-}
-
-static std::string pathWithoutDrive(const std::string& path) {
-    return path.substr(2, std::string::npos);
-}
-
-static std::string pathDrive(const std::string& path) {
-    return path.substr(0, 2);
+    std::error_code ec;
```
bakpaul, line 226:
```suggestion
    if(!exists(path, quiet))
        return false;
    
    std::error_code ec;
```

bakpaul: When we take a look at the previous implementation, before checking for the file type, the existence of the file was checked. Without this, all of the previous code that didn't check if the file existed before asking if it is a file fails. It seems, given the test results, that most of the current code base isn't doing this check. So I suggest to add this guard back.

bakpaul (changes requested):

bakpaul (approved):

PR #6105, opened 2026-05-05
[Type] Standardize extern template instantiations for vector

alxbilger:
This PR ensures that the template class `vector` is generated only once by the compiler for fundamental types such as `int`, `float`, `std::string` etc.


fredroy (changes requested):
Can you do/post benchmarks with/without the explicit instanciations ?
I remember doing this long time ago with base classes like vector, and got some performance regressions. 
I think this can prevent the compiler to inline functions, if you dont compile with LTO

PR #6108, opened 2026-05-06
[Pixi] Fix pixi compilation for CUDA related plugins

bakpaul:
Pixi compilation of SodaCUDA was broken because of two reasons : 
1. Missing packages now that cublas is required (by the way cusparse was also missing)
2. Now that we force double compilation by default, using the "native" architecture in pixi wasn't working anymore because it fall into an architecture not supporting double. To fix this properly I removed the option SOFACUDA_ENABLE_NATIVE_ARCHITECTURE current and instead use native as the default fallback instead of a random architecture. This makes more sense, a newby wanting to build it will expect it to build it for his architecture when providing no information... 

### DIFF
--- a/.github/workflows/ci-macos-linux-windows-pixi.yml
+++ b/.github/workflows/ci-macos-linux-windows-pixi.yml
@@ -38,7 +38,7 @@ jobs:
         env:
           SOFA_BUILD_TYPE: ${{ matrix.build_type }}
         run: |
-          pixi run -e ${{ matrix.environment }} configure -CMAKE_CUDA_ARCHITECTURES=60;89
+          pixi run -e ${{ matrix.environment }} configure -DCMAKE_CUDA_ARCHITECTURES=60;89
           pixi run -e ${{ matrix.environment }} build
 
       - name: Testing - Run SOFA in batch mode [MacOS/Linux/Windows]
--- a/.github/workflows/ci-macos-linux-windows-pixi.yml
+++ b/.github/workflows/ci-macos-linux-windows-pixi.yml
@@ -38,7 +38,7 @@ jobs:
         env:
           SOFA_BUILD_TYPE: ${{ matrix.build_type }}
         run: |
-          pixi run -e ${{ matrix.environment }} configure -DCMAKE_CUDA_ARCHITECTURES=60;89
+          pixi run -e ${{ matrix.environment }} configure -DCMAKE_CUDA_ARCHITECTURES="60;89"
           pixi run -e ${{ matrix.environment }} build
 
       - name: Testing - Run SOFA in batch mode [MacOS/Linux/Windows]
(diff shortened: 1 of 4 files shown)


applications/plugins/SofaCUDA/Core/CMakeLists.txt
```diff
@@ -5,17 +5,11 @@ set(SOFACUDA_CORE_MAJOR_VERSION 0)
...
-        set(CMAKE_CUDA_ARCHITECTURES native)
-    endif()
-endif()
 
 # set 75 as fallback value
 if(NOT DEFINED CMAKE_CUDA_ARCHITECTURES)
-    message(NOTICE "CMAKE_CUDA_ARCHITECTURES is not set, it will be set by default to 75")
-    set(CMAKE_CUDA_ARCHITECTURES 75)
+    message(NOTICE "CMAKE_CUDA_ARCHITECTURES is not set, it will be set by default to native")
+    set(CMAKE_CUDA_ARCHITECTURES native)
```
fredroy: The problem is that `native` keyword  is only supported for CMake >=3.24 https://cmake.org/cmake/help/latest/prop_tgt/CUDA_ARCHITECTURES.html and this plugin is supposed to be compatible with cmake >=3.22

so thats why the fallback was setting  a hard-coded version.
I guess the solution would be to force cmake 3.24 in the header of this file

fredroy (changes requested):

.github/workflows/ci-macos-linux-windows-pixi.yml
```diff
@@ -38,6 +38,7 @@ jobs:
         env:
           SOFA_BUILD_TYPE: ${{ matrix.build_type }}
         run: |
+          pixi run -e ${{ matrix.environment }} configure -DCMAKE_CUDA_ARCHITECTURES=60;89
```
bakpaul:
```suggestion
          pixi run -e ${{ matrix.environment }} configure -DCMAKE_CUDA_ARCHITECTURES="60;89"
```

fredroy:
```
-- Adding module SofaCUDA.Core
CMake Error at applications/plugins/SofaCUDA/Core/CMakeLists.txt:1 (cmake_minimum_required):
  CMake 3.24 or higher is required.  You are running version 3.22.1
```

@bakpaul

alxbilger:
The report says "Let's wait for the CI then ready". Apparently it's not ready, right?

bakpaul:
Nope, because I don't think that I want to think of a different way of fixing this just because of CMake version issue as we are going to update the builders in a few months. I'll take 10 minutes to think of it but this is clearly not in the priority.

bakpaul:
So now it is working but the behavior will not be the same depending on your CMake version. But it was the case before so...

fredroy (approved):

hugtalbot (approved):

PR #6117, opened 2026-05-20
[All] Introduce integration schemes

bakpaul:
This PR replaces the ODE solvers by Integration schemes. It enriches the API of Implicit integration schemes by exploding the timestep solving in sub phases useful for Newton applications

Now instead of single monolithic way of expressing an integration scheme, two families of integration schemes are introduced the Implicit and Explicit ones. 

Explicit integration schemes get no improvement from being run into a Newton solver, thus their implementation remain unchanged and they can still live in a one solve function. Enabling local optimization such as lumping the mass matrix to compute acceleration.   

Implicit solvers are divided here in to three main families : Velocity-based, acceleration-based and the rest. By doing so, the velocity and acceleration based integration schemes share a common implementation enabling a clear API to introduce new ones. As it can be seen in the PR, for any new Velocity based IS we need to implement 4 simple methods. For acceleration based it is 5. But the implementation sticks to the basics of what makes an IS : the position update, the velocity update, the acceleration computation and some derivative terms. A full documentation + scientific explanation will be added using the following this custom note https://typst.app/project/rVNQw8slVmqX7aHTCUrewe 

To list them, the current velocity based IS are 
- Linear multistep famillies
- Implicit Euler

And the acceleration based : 
- Newmark Beta

And the others are : 
- StaticEquilibrium
- Simplectic (energy-based)

This is a first required step to implement a global Newton on the simulation. 

There is still a TODO list : 
- [x] Investigate the reduced stability in example scenes
- [ ] ~(re) implement the symplectic IS~
- [x] Re-implement a newton for the static solver (replaced in future PR)
- [x] Implement trapezoidal rule
- [x] Implement first order for implicit solver
- [x] Fix all scenes 
- [x] Add a compatibility layer ? 
- [x] Build a doc using the scientific notes
  - See PR https://github.com/sofa-framework/doc/pull/227 
- [x] Decide where to put the rayleigh parameters

PS: I am sorry for the history being croped to 11 commits. There should be way more but it has been lost because I started with a much complex PR including the work on the Newton solver which in the end was to intricated, so I started fresh with only the modifications on the ODE... 

About first order : 
I've only implemented is for velocity-based IS, I don't feel that it makes sense in acceleration based one. For instance, in Acceleration based one we could also think of second order approximation. But for instance for the Newmark solver it would react as an explicit IS or as if beta = gamma = 0. For me this approximation makes sense for second order IS, but if you use a higher order one, you just want every order. Actually I have the implementation somewhere but it makes the implementation much more complexe for no real use IMO

After struggling with diffusion it seems like the first order implementation was broken since its creation. A 1/h was missing in front of the mass matrix in the LHS. This PR fixes it too.

List of regression tests I had to regenerate because of unconverged CG:
- examples/Demos/collisionMultiple.scn (ERRORBYDOF: 0.000555197)
- examples/Demos/fallingSOFA.scn (ERRORBYDOF: 2.29746e-05)
- *examples/Demos/SofaScene.scn (Incoming, waiting for https://github.com/sofa-framework/BeamAdapter/pull/237)*
- examples/Component/Collision/Response/RuleBasedContactManager.scn (ERRORBYDOF: 0.000169497)
- examples/Component/Collision/Detection/CompositeCollisionPipeline_none.scn (ERRORBYDOF: 0.000197937)
- examples/Component/LinearSolver/Preconditioner/FEMBAR_PCG_BlockJacobiPreconditioner.scn (ERRORBYDOF: 2.33959e-05)
- examples/Component/LinearSolver/Preconditioner/FEMBAR_PCG_JacobiPreconditioner.scn (ERRORBYDOF: 2.27517e-05)
- examples/Component/LinearSolver/Preconditioner/FEMBAR_PCG_NoPreconditioner.scn (ERRORBYDOF: 9.06356e-06)
- examples/Component/LinearSolver/Preconditioner/FEMBAR_PCG_SSORPreconditioner.scn (ERRORBYDOF: 1.30806e-05)
- examples/Component/Mapping/NonLinear/DistanceMapping.scn ( ERRORBYDOF: 0.00858526)
- examples/Component/Mapping/NonLinear/DistanceMultiMapping.scn (ERRORBYDOF: 0.436686)
- examples/Component/Mapping/NonLinear/SquareDistanceMapping.scn (ERRORBYDOF: 0.0063835)
- examples/Component/SolidMechanics/FEM/TriangularFEMForceFieldOptim.scn (ERRORBYDOF: 0.000298034)
- examples/Component/SolidMechanics/Spring/MeshSpringForceField.scn (ERRORBYDOF: 0.0004329)
- examples/Benchmark/Analysis/Pendulum.scn (ERRORBYDOF: 0.000971208)
- applications/plugins/SofaDistanceGrid/examples/RigidDistanceGridCollisionModel_liver_DefaultAnimationLoop.scn (ERRORBYDOF: 7.92855e-08)
- applications/plugins/SofaDistanceGrid/examples/RigidDistanceGridCollisionModel_liver_FreeMotionAnimationLoop.scn (ERRORBYDOF: 1.70512e-09)

--> Most of these errors are due to CG not converged except for "Pendulum, SofaScene, fallingSofa, and distance grid ones". But for most of those the error is imperceptible visually : the expected precision might be wrongly set, or explaned in this list.  
--> All scenes on non-linear mappings where simply broken because the CG did only one step. 
--> The pendulum is now working better because the trapezoidal rule was a bit broken before
--> The SofaScene has huge difference because of the BeamAdapter addMDx  method that is broken. The regression files will be regenerated when the PR is merged if necessary.

Finally the forceFeedback component in SOFA wasn't using the contact force but the impulsion dt*lambda) because the legacy implementation of the EulerImplicitSolver was multiplying the RHS and the LHS by dt. This is not the case anymore, so now the forces are real forces, and so you will need to carefully update the force factors in your scene to cope for this missing factor.  Plus, it has to be noted that the unit test values of forces will be updated with forces instead of impulsion.

Last point on the static solver: the old implementation of the Newton-Raphson algorithm has been droped as it will be reimplemented in a more generic way in future PR. But the new version of the static solver still propose a newton algorithm based on Armijo condition ofr linesearch. It is 'optimized' by considering the system matric as symetric only for computin the line search stopping criterion. This will be studied more profoundly in the future generic implmentation. 

I've added some todos here and there for future work on Newton. Some of them are instruciton some are still questions to be answered beut are not needed until a proper Newton algorithm is introduced. 

One especially is to be noted : the propagation of states. It is currently done by the animation loop. The issue is that the freemotion part is wrongly done as it is not the freePos that is directly propagated buyt first the freeVel then we explicitly integrate it using eulerImplicit scheme which doesn't make sense. This will be fixed in another PR. 

[with-all-tests]
[force-full-build]
[ci-depends-on https://github.com/sofa-framework/SofaPython3/pull/615]
[ci-depends-on https://github.com/sofa-framework/SofaGLFW/pull/283]
[ci-depends-on https://github.com/sofa-framework/BeamAdapter/pull/232]
[ci-depends-on https://github.com/sofa-framework/CSparseSolvers/pull/8]
[ci-depends-on https://github.com/sofa-framework/ManifoldTopologies/pull/16]
[ci-depends-on https://github.com/SofaDefrost/ModelOrderReduction/pull/176]
[ci-depends-on https://github.com/sofa-framework/PluginExample/pull/23]
[ci-depends-on https://github.com/sofa-framework/Registration/pull/35]
[ci-depends-on https://github.com/sofa-framework/SofaSphFluid/pull/22]
[ci-depends-on https://github.com/SofaDefrost/SoftRobots/pull/346]
[ci-depends-on https://github.com/SofaDefrost/SoftRobots.Inverse/pull/86]
[ci-depends-on https://github.com/SofaDefrost/STLIB/pull/139]
[ci-depends-on https://github.com/sofa-framework/Regression/pull/114]


bakpaul:
[ci-build][with-all-tests]

Sofa/framework/Core/src/sofa/core/behavior/BaseIntegrationScheme.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include <sofa/core/behavior/MultiVec.h>
+
+namespace sofa::core::behavior
+{
+
+/**
+ *  \brief Component responsible for timestep integration, i.e. advancing the state from time t to t+dt.
+ *
+ *  This class currently control both the integration scheme (explicit,
```
alxbilger: There is something wrong in this sentence.

Sofa/framework/Core/src/sofa/core/behavior/IntegrationScheme.h
```diff
@@ -0,0 +1,77 @@
...
+ *  While all computations required to do the integration step are handled by
+ *  this object, they should not be implemented directly in it, but instead
+ *  the solver propagates orders (or Visitor) to the other components in the
+ *  scenegraph that will locally execute them. This allows for greater
+ *  flexibility (the solver can just ask for the forces to be computed without
+ *  knowing what type of forces are present), as well as performances
+ *  (some computations can be executed in parallel).
+ *
+ */
+class SOFA_CORE_API IntegrationScheme : public virtual objectmodel::BaseComponent
```
alxbilger: Usually the abstract base classes in `Core`  are prefixed with `Base`. I recommand `BaseIntegrationScheme` for consistency.

Sofa/framework/Core/src/sofa/core/behavior/BaseIntegrationScheme.h
```diff
@@ -0,0 +1,77 @@
...
+    SOFA_ABSTRACT_CLASS(IntegrationScheme, objectmodel::BaseComponent);
+    SOFA_BASE_CAST_IMPLEMENTATION(IntegrationScheme)
+
+protected:
+    IntegrationScheme();
+    ~IntegrationScheme() override;
+
+public:
+
+    virtual void integrate(const core::ExecParams* params, SReal dt, sofa::core::MultiVecCoordId xResult, sofa::core::MultiVecDerivId vResult) = 0;
```
alxbilger, line 60: Could you add a documentation for this function?

Sofa/framework/Core/src/sofa/core/behavior/BaseIntegrationScheme.h
```diff
@@ -0,0 +1,77 @@
...
+
+    /// Given the solution dx of the linear system inversion, how much will it affect the velocity
+    ///
+    /// This method is used to compute the compliance for contact corrections
```
alxbilger: Sorry, but the current documentation does not tell me what this function returns. Same for `getPositionIntegrationFactor`

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
```diff
@@ -0,0 +1,131 @@
...
+    virtual void solveLinearEquation() = 0;
+
+    /**
+     * Once (x^{i+1}-x^i) has been computed, the result is used internally to update the current
+     * guess. It computes x^{i+1} += alpha * dx, where dx is the result of the linear system. It is
+     * not necessary to share the result with the Newton-Raphson method.
+     */
+    virtual void updateStatesFromLinearSolution(SReal alpha, bool firstIteration = false) = 0;
+
+    virtual void integrate(const core::ExecParams* params, SReal dt, sofa::core::MultiVecCoordId xResult, sofa::core::MultiVecDerivId vResult) ;
```
alxbilger: You should do a pass on the whole PR checking missing `override`  keyword.

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
```diff
@@ -0,0 +1,131 @@
...
+     /**
+     * compute the current RHS.
+     */
+    virtual void computeRHS(bool firstIteration = false) = 0;
+
+
+    /**
+     * Returns the evaluation of the residue
+     */
+    virtual SReal evaluateResidue() = 0;
```
alxbilger: I think it should be `residual` instead of `residue`.

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
```diff
@@ -0,0 +1,131 @@
...
+                            public  sofa::core::behavior::IntegrationScheme,
+                            public sofa::core::behavior::LinearSolverAccessor
+{
+public:
+    SOFA_ABSTRACT_CLASS(ImplicitIntegrationScheme, sofa::core::behavior::IntegrationScheme);
+
+    // WARNING we expect the linear integrator to initialize the working vecs. Meaning that if we
+    // work in FreeMotion, the xResult should already be equal to the actual position.
+    // Same for the velocity. This is expected when updating the position, only a += will be done.
+    virtual void setupIntegrationStep(const core::ExecParams* params, SReal dt, sofa::core::MultiVecCoordId xResult, sofa::core::MultiVecDerivId vResult);
```
alxbilger: Is it necessary to make it virtual since you use the template method pattern?

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
```diff
@@ -0,0 +1,131 @@
...
+class SOFA_SIMULATION_CORE_API ImplicitIntegrationScheme :
+                            public  sofa::core::behavior::IntegrationScheme,
+                            public sofa::core::behavior::LinearSolverAccessor
+{
+public:
+    SOFA_ABSTRACT_CLASS(ImplicitIntegrationScheme, sofa::core::behavior::IntegrationScheme);
+
+    // WARNING we expect the linear integrator to initialize the working vecs. Meaning that if we
+    // work in FreeMotion, the xResult should already be equal to the actual position.
+    // Same for the velocity. This is expected when updating the position, only a += will be done.
```
alxbilger: Could you clarify a bit?

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/AccelerationBasedImplicitIntegrationScheme.h
```diff
@@ -0,0 +1,93 @@
...
+#include <sofa/core/behavior/LinearSolverAccessor.h>
+#include <sofa/simulation/integrationschemes/ImplicitIntegrationScheme.h>
+
+#include <sofa/simulation/MechanicalOperations.h>
+#include <sofa/simulation/VectorOperations.h>
+
+namespace sofa::simulation::integrationschemes
+{
+
+class SOFA_SIMULATION_CORE_API AccelerationBasedImplicitIntegrationScheme :
```
alxbilger, line 46: Can you document this class?

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
```diff
@@ -0,0 +1,131 @@
...
+protected:
+
+    virtual sofa::Size getIntegrationSchemeTimeOrder() const = 0;
+
+
+
+    const core::ExecParams* m_params;
+    sofa::core::MultiVecCoordId m_xResult;
+    sofa::core::MultiVecDerivId m_vResult;
+    sofa::core::MultiVecDerivId m_unknown;
```
alxbilger: I was confused by the name of this variable. I recommend `m_systemUnknown`

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
bakpaul: Yes, but it was to put the final keyword at the end, and of course I forgot it

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
bakpaul: Actually this is not true anymore, no assumption is made, we copy the right vector into the result before starting

Sofa/framework/Core/src/sofa/core/behavior/BaseIntegrationScheme.h
bakpaul: I reworked it. It came from the old component...

bakpaul:

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ExplicitIntegrationScheme.h
alxbilger, line 1: the source file in the diff is questionable...

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ExplicitIntegrationScheme.h
```diff
@@ -20,50 +20,46 @@
...
-    }
 
-    SReal squaredAbsoluteDifference = 0;
-    SReal squaredPreviousEvaluation = 0;
+    /**
+     * Template method pattern is implemented here to ensure some data/member are rightly initialized
+     * before solving. This method now internally calls doIntegrate.
+     */
+    virtual void integrate(const core::ExecParams* params, SReal dt, sofa::core::MultiVecCoordId xResult, sofa::core::MultiVecDerivId vResult) override final;
+    virtual void doIntegrate(const core::ExecParams* params, sofa::core::MultiVecCoordId xResult, sofa::core::MultiVecDerivId vResult) = 0 ;
```
alxbilger, line 56: Should be protected no?

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
```diff
@@ -0,0 +1,168 @@
...
+                            public sofa::core::behavior::LinearSolverAccessor
+{
+public:
+    SOFA_ABSTRACT_CLASS2(ImplicitIntegrationScheme, sofa::core::behavior::BaseIntegrationScheme, sofa::core::behavior::LinearSolverAccessor);
+
+    /**
+     * Template method pattern is implemented here to ensure some data/member are rightly initialized
+     * before solving. This method internally calls doSetupIntegrationStep that can be overridden.
+     */
+    virtual void setupIntegrationStep(const core::ExecParams* params, SReal dt, sofa::core::MultiVecCoordId xResult, sofa::core::MultiVecDerivId vResult) final;
```
alxbilger, line 53: do you need those methods to be public? Can't they be protected?

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/VelocityBasedImplicitIntegrationScheme.h
```diff
@@ -0,0 +1,125 @@
...
+    virtual void finalizeIntegrationStep() override;
+
+    virtual SReal getVelocityIntegrationFactor() const override final;
+    virtual SReal getPositionIntegrationFactor() const override final;
+
+protected:
+    virtual sofa::Size getIntegrationSchemeTimeOrder() const = 0;
+    /** ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ **/
+
+    /**   New API methods of VelocityBasedImplicitIntegrationScheme   **/
```
alxbilger, line 80:
```suggestion
    /**   API methods of VelocityBasedImplicitIntegrationScheme   **/
```
"New" will no longer be valid if your API is still here in 15 years.

Sofa/framework/Simulation/Core/src/sofa/simulation/Colors.cpp
```diff
@@ -64,7 +64,7 @@ std::map<const std::string, int> colors = {
         {"InteractionForceField",   IFFIELD},
         {"ForceField",              FFIELD},
         {"BaseAnimationLoop",       SOLVER},
-        {"OdeSolver",               SOLVER}, // même valeur que ci-dessus
+        {"IntegrationScheme",               SOLVER}, // même valeur que ci-dessus
```
alxbilger, line 67:
```suggestion
        {"IntegrationScheme",               SOLVER},
```

Sofa/framework/Core/src/sofa/core/behavior/OdeSolver.h
```diff
@@ -21,107 +21,11 @@
...
-    }
-
-
-    bool insertInNode( objectmodel::BaseNode* node ) override;
-    bool removeInNode( objectmodel::BaseNode* node ) override;
-
-};
-
-} // namespace sofa::core::behavior
+    using OdeSolver = BaseIntegrationScheme;
```
alxbilger, line 30: Should this alias be deprecated?

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ExplicitIntegrationScheme.h
```diff
@@ -20,50 +20,46 @@
...
 {
+class MechanicalOperations;
+class VectorOperations;
+}
 
+namespace sofa::simulation::integrationscheme
+{
 
-class RelativeEstimateDifferenceMeasure : public NewtonRaphsonConvergenceMeasureWithSquaredParameter
+class SOFA_SIMULATION_CORE_API ExplicitIntegrationScheme :
```
alxbilger, line 41: I feel that this class is not really specialized for explicit methods. And if we compare to `ImplicitIntegrationScheme`, we can see a lot of similarities in `setupIntegrationStep`. The only advantage I see is for strong typing (which can be a valid reason to keep this class). What's your opinion?

bakpaul: This is exactly the reason I needed this type : strong typing will allow to treat differently implicit and explicit IS in the Newton implementaiton.

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ExplicitIntegrationScheme.h
bakpaul: Yes...

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ExplicitIntegrationScheme.h
bakpaul: yeap

Sofa/framework/Simulation/Core/src/sofa/simulation/integrationscheme/ImplicitIntegrationScheme.h
bakpaul: The setup will be called by the newton at one point, so I prefere it to stay public.

Sofa/framework/Core/src/sofa/core/behavior/OdeSolver.h
bakpaul: The header is deprecated, isn't it sufficient ?

fredroy:
```
[02](https://github.com/sofa-framework/sofa/actions/runs/35115461042/job/105021548782?pr=6117#step:12:103)
    Message content: /workspace/sofa/examples/Component/Mapping/NonLinear/DistanceMultiMapping.scn:  
    TOTALERROR: 1.07033e-05  
    ERRORBYDOF: 1.33792e-07
```
I guess you would need to regenerate the references for DistanceMultiMapping.scn

bakpaul:
Indeed, as explained last time, this was one of the unit test that was not converged in the master branch. So the new PR might break it while mine will catch those errors. I should have done another PR to fix those tests directly in master. But let's regenerate it an move on !

bakpaul:
[ci-build]
[force-full-build]

fredroy:
[ci-build][with-all-tests]

hugtalbot (approved):
Further to SOFA dev

PR #6118, opened 2026-05-20
[SofaCarving] Clean the handleEvent method to have less conditions and also remove some Data

epernod:
All those different events to activate the carving do not make sense, as the Data can be changed directly from python controller or code or linking it to other Data.

The middle mouse event was hidden and was randomly activating/deactivating the carving when you were zooming in the 3D view leading to buggy situation.

### DIFF
--- a/applications/plugins/SofaCarving/src/SofaCarving/CarvingManager.cpp
+++ b/applications/plugins/SofaCarving/src/SofaCarving/CarvingManager.cpp
@@ -188,6 +188,7 @@ void CarvingManager::handleEvent(sofa::core::objectmodel::Event* event)
 
     if (sofa::core::objectmodel::KeypressedEvent::checkEventType(event))
     {
+        sofa::core::objectmodel::KeypressedEvent* ev = dynamic_cast<sofa::core::objectmodel::KeypressedEvent*>(event);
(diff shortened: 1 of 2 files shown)


applications/plugins/SofaCarving/src/SofaCarving/CarvingManager.cpp
```diff
@@ -188,9 +186,8 @@ void CarvingManager::handleEvent(sofa::core::objectmodel::Event* event)
     if (d_componentState.getValue() != sofa::core::objectmodel::ComponentState::Valid)
         return;
 
-    if (sofa::core::objectmodel::KeypressedEvent::checkEventType(event))
+    if (sofa::core::objectmodel::KeypressedEvent* ev = dynamic_cast<sofa::core::objectmodel::KeypressedEvent*>(event))
```
fredroy: https://github.com/sofa-framework/sofa/pull/6109

fredroy (changes requested):

applications/plugins/SofaCarving/src/SofaCarving/CarvingManager.cpp
epernod: ah bad merge, good catch

fredroy (approved):

PR #6120, opened 2026-05-21
[defaulttype] Fix VecN type when compiling with float

alxbilger:
`type::Vec3` uses `SReal`. But when you compile `StdRigidTypes` with something different from `SReal`, compilation fails because types are not compatible.


fredroy (approved):

fredroy:
But it was not very clear that in this case "Vec3" (without type::) referred to 

https://github.com/sofa-framework/sofa/blob/master/Sofa/framework/DefaultType/src/sofa/defaulttype/RigidTypes.h#L55

😅

PR #6124, opened 2026-05-23
[FEM] Refactor templates in latest FEM classes

alxbilger:
The latest FEM classes defined custom names and custom template names so that the class is registered in the factory with this custom name. However, this is not a good practice. This PR restores the regular pattern where the class is registered using its own class. Because of this, I could also remove the `Element` prefix on all the FEM classes.

Before:

```xml
<PrismFEMMass/>
```

After:

```xml
<FEMMass template="Vec3,Prism"/>
```

[with-all-tests]


Sofa/Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/FEMForceField.h
```diff
@@ -106,6 +106,8 @@ class FEMForceField :
         sofa::type::vector<ElementForce>& df,
         const sofa::VecDeriv_t<DataTypes>& dx);
 
+    virtual void beforeElementForceDeriv(const sofa::core::MechanicalParams* mparams) {}
```
th-skam, line 109: Curiosity, what is the plan for this function?

alxbilger: Precomputation only useful when the Hessian is needed. For example:

```cpp
if (!m_isHessianValid)
{
    computeHessian(*m_coordinates);
}
```

Note that there is already an equivalent in `addForce`: `beforeElementForce`

fredroy:
[ci-build][with-all-tests]

alxbilger:
The CUDA force fields are not a template specialization. They are new classes deriving from the CPU class. Therefore, the component name takes the class name. That is why there is a `Cuda` prefix.

@fredroy

fredroy:
> The CUDA force fields are not a template specialization. They are new classes deriving from the CPU class. Therefore, the component name takes the class name. That is why there is a `Cuda` prefix.
> 
> @fredroy

too bad we have both the Cuda prefix on the component + the template (CudaVec3X). Worth a look in the future to remove this inconsistency with the other Cuda stuff (its either Cuda prefix XOR the template usually)

fredroy (approved):

Issue #6151, opened 2026-06-15
DrawMesh<Hexahedron> does not check the consistency of the topology


fredroy:
### Problem

When using a SparseGrid and something drawing the associated hexahedra using DrawMesh<Hexa> (for example HexahedronFEMForcefield and rendering the FEM), then it is crashing because it is expecting having the associated quads composing the hexahedra. It seems that SparseGrid does only create the hexahedra.

**Steps to reproduce**
Simple Scene:
```
<Node name="root" gravity="0 0 0"  dt="0.04">
    <Node name="plugins" >
        <RequiredPlugin pluginName="Sofa.Component.Constraint.Projective"/> <!-- Needed to use components [FixedProjectiveConstraint] -->
        <RequiredPlugin pluginName="Sofa.Component.Engine.Select"/> <!-- Needed to use components [BoxROI] -->
        <RequiredPlugin pluginName="Sofa.Component.IO.Mesh"/> <!-- Needed to use components [MeshOBJLoader] -->
        <RequiredPlugin pluginName="Sofa.Component.LinearSolver.Iterative"/> <!-- Needed to use components [CGLinearSolver] -->
        <RequiredPlugin pluginName="Sofa.Component.Mapping.Linear"/> <!-- Needed to use components [BarycentricMapping] -->
        <RequiredPlugin pluginName="Sofa.Component.Mass"/> <!-- Needed to use components [UniformMass] -->
        <RequiredPlugin pluginName="Sofa.Component.ODESolver.Backward"/> <!-- Needed to use components [EulerImplicitSolver] -->
        <RequiredPlugin pluginName="Sofa.Component.SolidMechanics.FEM.Elastic"/> <!-- Needed to use components [HexahedronFEMForceField] -->
        <RequiredPlugin pluginName="Sofa.Component.StateContainer"/> <!-- Needed to use components [MechanicalObject] -->
        <RequiredPlugin pluginName="Sofa.Component.Topology.Container.Grid"/> <!-- Needed to use components [SparseGridTopology] -->
        <RequiredPlugin pluginName="Sofa.GL.Component.Rendering3D"/> <!-- Needed to use components [OglModel] -->
    </Node>

    <DefaultAnimationLoop />

    <Node name="DynamicStuff" >
        <EulerImplicitSolver name="cg_odesolver"  />
        <CGLinearSolver name="linear solver" iterations="25" tolerance="1e-15" threshold="1e-15" />

        <Node name="Cylinder" >
[...]

PR #6152, opened 2026-06-15
[Core] DrawMesh<Hexahedron>: add quick test for quads when drawing hexa

fredroy:
More a hotfix than a real fix:
-> #6151 

as I think sparsegrid and hexahedronFEMForceField are quite used here and there.

The real fix(es) should be:
- sparsegrid should also generate quads ?
- DrawMesh (all the primitives) should check the consistency of the topology ?
- even better, DrawMesh should just generates its own facets but slower of course

and maybe, doing the same for triangles and Tetrahedron ?

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/visual/DrawMesh.h
+++ b/Sofa/framework/Core/src/sofa/core/visual/DrawMesh.h
@@ -328,6 +328,17 @@ private:
         const auto& elements = topology->getTetrahedra();
         const auto& facets = topology->getTriangles();
 
+        if(facets.empty())
+        {
+            static bool firstTime = true;
+            if (firstTime)
+            {
+                msg_error("DrawElementMesh<Tetrahedron>") << "Drawing tetrahedra needs the associated triangles in the topology.";
+                firstTime = false;
+            }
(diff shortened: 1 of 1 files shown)


bakpaul (approved):

bakpaul:
I would advocate going for the third option when no facet is available in the topology. And it would only be slower for the first timestep.

fredroy:
> I would advocate going for the third option when no facet is available in the topology. And it would only be slower for the first timestep.

The problem with that solution is that:
 - we need to store the generated quads
 - we will still need to keep track of the hexas' state in case of a topology change (removal of an hexa for example)
 - will quite change the "spirit" of this class
 
So in this case, the best would be that the topology itself manages the quad (the problem in the first place actually) with a function like `BaseMeshTopology::generateSubElements<Hexa,Quad>` or something like that.

Finally my preferred solution actually would be simply to call `drawHexahedra()` instead of `drawQuads` from the `DrawTool`, which is taking care of creating the visual quads itself. But maybe @alxbilger had a reason to not call it.

alxbilger (approved):

PR #6159, opened 2026-06-16
[examples] Use CCD for example scene

bakpaul:
Use the CCD collision detection in the example scene to remove the bad looking skinning effect (not very good for showcasting).

The downside is that we have a 20% reduce in performances (on my computed we went from 240 fps to 190)

[with-all-tests]
[ci-depends-on https://github.com/sofa-framework/Regression/pull/115]


bakpaul:
[ci-build][with-all-tests]

hugtalbot (changes requested):
It would be nice to have a dedicated scene example for CCD
Moreover, your change added a warning in the scene :

```
[WARNING] [CCDTightInclusionIntersection(CCDTightInclusionIntersection)] Required data "alarmDistance" has not been set. Falling back to default value: 1
```

bakpaul:
I confirm that there is no added warning with this scene !

hugtalbot (approved):

PR #6129, opened 2026-05-27
[CORE] Refactoring add functions from BaseLink and Link

Lucas-TJ:
Hello,

This PR is there to make the code of the add functions clearer in BaseLink and Link.

Later,  I will add some unit tests.

Thank you

[with-all-tests]
______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #6168, opened 2026-07-02
[All] Introduce Snapshot feature

Lucas-TJ:
Based on 
- #6129 
- #6130 

## What is a Snapshot ?

A snapshot is an object that stores all information required to save the state of a simulation, so that it can be later restored.

### General 

A snapshot is a class structured like this :

- Data and Links are stored in dedicated structures called DataInfo and LinkInfo
- If the object is a node, then it creates a SnapshotNode is created. It contains data, links, components (SnapshotObject) and child nodes (also SnapshotNode instances).
- If the object is a component, a SnapshotObject is created. It contains only data and links, and is stored inside a SnapshotNode.

Once the snapshot has been built, it can be exported to the format chosen by the user. For example, calling saveSnapshot with the JSON exporter serializes the snapshot into a JSON file. Loading works the same way in reverse: the snapshot is first imported from the chosen format, then used to restore the simulation state.

#### Structure of a Snapshot

```cpp
public:
    struct DataInfo
    {
        std::string name;
        std::string type;
        std::string value;
    };

    struct LinkInfo
    {
        std::string name;
        std::string type;
        std::string value;
    };

    struct SnapshotObject
    {
        std::string m_name;
        std::string m_className;
        std::string m_pathName;
        std::vector<DataInfo> m_dataContainer;
        std::vector<LinkInfo> m_linkContainer;
        std::vector<std::shared_ptr<SnapshotObject>> m_objects;

        SnapshotObject() = default;
        explicit SnapshotObject(const std::string& name) : m_name(name){}

        virtual ~SnapshotObject() = default;
    };

    struct SnapshotNode : public SnapshotObject
    {
        std::vector<std::shared_ptr<SnapshotNode>> children;
        
        SnapshotNode() = default;
        SnapshotNode(const std::string& name) : SnapshotObject(name) {}
        SnapshotNode(const SnapshotObject& obj) : SnapshotObject(obj) {}

        ~SnapshotNode() noexcept = default;
    };

    std::shared_ptr<SnapshotNode> m_graphRoot { nullptr };
```

A Snapshot is organized as a graph. The root object, m_graphRoot, is the object that is exported to or imported from a file (or stored directly in memory). The graph is composed of nodes.

- A node corresponds to the `struct SnapshotNode`. Each node contains a name, a classname, a pathname, a data container, a link container, a list of components, and a list of child nodes.
- A component corresponds to the `struct SnapshotObject`. Each component contains a name, a classname, a pathname, a data container, a link container, a slave container).
- Data and links are represented in the same way: each entry stores its name, type, and value
  
For example,

```json
{
    "name" : "root",
    "classname" : "Node",
    "pathname" : "/",
    "data" : [{},{}],
    "links" : [{},{}],
    
    "components" : [
        {
            "data" : [{},{}],
            "links" : [{},{}]
        },
        {
            "data" : [{},{}],
            "links" : [{},{}]
        }
    ],

    "children": []
}
```

Saving and loading snapshots rely on two visitors: SaveSnapshotVisitor and LoadSnapshotVisitor. These visitors traverse the scene graph to collect or restore data and links. Once collected, a snapshot can either be kept in memory or exported to a JSON file using SnapshotJSONExporter. The reverse process is used when loading a snapshot.

#### Save

Saving a snapshot relies on the SaveSnapshotVisitor, which traverses the scene graph. For each visited node or component, it calls:

 ```Base::saveSnapshot(std::vector<std::shared_ptr<SnapshotNode>>& )```. 

Inside ```Base::saveSnapshot ```, the function 

```Base::createSnapshotObject(std::vector<std::shared_ptr<Snapshot::SnapshotNode>>&)```

creates either a SnapshotObject or a SnapshotNode, depending on the type of the current object.

The snapshot object is then populated with:

- ```snapshotObject->m_name = this->getName()``` to store the object's name.
- ```snapshotObject->m_classname = this->getClassName()``` to store the object's className.
- ```snapshotObject->m_pathname = this->getPathName()``` to store the object's pathName.
- a part to serialize the object's data.
- a part to serialize the object's links.
- ```saveInternalStateIn(*snapshotObject)``` to serialize the object's Internal State.

Finally, the newly created SnapshotObject (or SnapshotNode) is inserted into the snapshot graph.

#### Load

Loading a snapshot relies on the LoadSnapshotVisitor, which traverses the scene graph and restores the saved state by calling:
``` Base::findSnapshotObject(const std::shared_ptr<Snapshot::SnapshotNode>& parents, const std::string& objectname, const std::string& classname, const std::string& pathname) ``` : use to find the component/node corresponding to the SnapshotObject

 ``` Base::loadSnapshot(const std::shared_ptr<Snapshot::SnapshotObject>& snapshotObject) ```

 ``` Base::loadInternalStateFrom(const Snapshot::SnapshotObject& snapshot) ```

```loadSnapshot``` use : 
- ``` BaseData::read(const std::string& value) ``` to restore data values from the snapshot.
- ``` BaseLink::readFromSnapshot(const std::string& value) ``` to restore links from the snapshot.

#### Overview

SnapshotManager is responsible for storing and managing snapshots, whether they are kept in memory or stored on disk.

#### API

Snapshots are stored in two separate containers: ```m_snapshotsFromMemory``` for in-memory snapshots and ```m_snapshotsFromFiles``` for snapshots stored on disk. The ```addSnapshotFromMemory``` and ```addSnapshotFromFile``` functions are used to add snapshots to these containers.

Although in-memory and on-disk snapshots are handled separately, their APIs follow the same overall design.

- ```memorySave``` and ```memoryLoad``` save and restore snapshots in memory.
- ```saveTo``` and ```loadTo``` save and load snapshots from disk. ```saveTo``` allows the output file format to be specified.

The ```saveTo``` function also provides an isSet parameter, and the ```loadToSet``` function is available for loading a collection of snapshots. These features are intended for saving and restoring groups of snapshots within a single file.

#### Overview

SnapshotJSONExporter provides the functionality required to export a a snapshot to JSON and import it back from a JSON file. It relies on the nlohmann/json library. 

#### API

The implementation is split in two parts : one for exporting snapshots and one for importing them.

- For exporting, the ``` exportToJSON``` function serializes a Snapshot into a JSON object by calling the appropriate ```to_json``` overloads before writing the result to a file.

- For importing, the ```importFrom``` function reads a JSON file and reconstructs a Snapshot by calling the corresponding ```from_json``` overloads.

- Two additional helper functions, ```fileToString``` and ```snapshotToString```, provide serialization and deserialization to and from strings.

### Slaves

As discussed during the review, snapshots can now save and restore slave objects.

To better understand the `Snapshot` API, particularly the `SnapshotObject` and `SnapshotNode` classes, a new vector named `m_objects` has been introduced. Its purpose depends on the type of snapshot being stored:

- In a `SnapshotObject`, `m_objects` contains the slave objects of the component.
- In a `SnapshotNode`, `m_objects` contains the components attached to the node.

### Update on Snapshot's API

To improve the accuracy of snapshot loading, the Snapshot structure has been updated with two new pieces of information: the class name (`m_classname`) and the path (`m_pathname`).

This slightly changes the `findSnapshotObject()` API, making it more accurate when identifying the corresponding component or node in the scene graph.

### loadInternalStateFrom and saveInternalStateIn

As you can see in the PR, I use `loadInternalStateFrom` and `saveInternalStateIn`. Initially, these functions are intended to save/load the internal state of components. However, in some cases, I use them to reset or clear some variables:

- In `CollisionResponse`, I use `loadInternalStateFrom` to clear the `contactMap` pointer.
- In `TetrahedronCollisionModel`, I use `loadInternalStateFrom` to update the topology.
- In `TriangleCollisionModel`, I use `saveInternalStateIn` and `loadInternalStateFrom` to save and load the `m_triangles` data.

[with-all-tests]

### CODE REVIEW

File: Sofa/framework/Core/src/sofa/core/objectmodel/Base.cpp [hunk 1]
```diff
@@ -680,7 +684,117 @@ int Base::getInstanciationSourceFilePos() const
...
+        linkInfo.type = link->getValueTypeString();
+        linkInfo.value = link->getValueString();
+
+        std::string replaceValue = "//";
+        std::size_t pos = linkInfo.value.find(replaceValue);
+        while (pos != std::string::npos)
+        {
+            linkInfo.value.replace(pos, replaceValue.length(), "");
+            pos = linkInfo.value.find(replaceValue, pos);
+        }
```
alxbilger: look if you can use `sofa::helper::replaceAll` instead to be more concise

Sofa/framework/Core/src/sofa/core/objectmodel/Base.cpp
```diff
@@ -680,7 +684,117 @@ int Base::getInstanciationSourceFilePos() const
...
+{
+    for (const auto& p : parents->components)
+    {
+        if (p.m_name == objectname)
+        {
+            auto snapshotObject = std::make_shared<Snapshot::SnapshotObject>(p);
+            return snapshotObject;
+        }
+    }
+    msg_error("findSnapshotObject") << "SnapshotObject "<< objectname << " not found";
```
alxbilger:
```suggestion
    msg_error() << "SnapshotObject "<< objectname << " not found";
```

alxbilger: and same for all the msg_error in the same file

Sofa/framework/Core/src/sofa/core/objectmodel/Base.cpp
```diff
@@ -680,7 +684,117 @@ int Base::getInstanciationSourceFilePos() const
...
+    {
+        if (const auto data = this->findData(dataInfo.name))
+        {
+            if(data->read(dataInfo.value) == 0 )
+                msg_error("LoadDataSnapshot") << "Failed to read " << dataInfo.name << " in " << this->getName()  << " from the snapshot " << dataInfo.value;
+        }
+    }
+}
+
+void Base::loadLinkSnapshot(const std::shared_ptr<Snapshot::SnapshotObject>& snapshotObject) const {
```
alxbilger: Use Allman style braces (valid for the whole PR)

Sofa/framework/Core/src/sofa/core/objectmodel/BaseLink.cpp
```diff
@@ -323,6 +323,88 @@ bool BaseLink::read( const std::string& str )
...
+
+    std::string linkPath = this->getValueString();
+
+    std::string replaceValue = "//";
+    std::size_t pos = linkPath.find(replaceValue);
+    while (pos != std::string::npos)
+    {
+        linkPath.replace(pos, replaceValue.length(), "");
+        pos = linkPath.find(replaceValue, pos);
+    }
```
alxbilger: again `sofa::helper::replaceAll`?

Sofa/framework/Core/src/sofa/core/objectmodel/SnapshotManager.h
```diff
@@ -0,0 +1,48 @@
...
+*  This class contains methods and elements that will help to manage snapshot in memory.
+*/
+class SnapshotManager
+{
+public:
+    SnapshotManager();
+    ~SnapshotManager();
+
+    std::vector<std::string> recentSnapshotFiles;
+    std::map<std::string, std::shared_ptr<sofa::core::objectmodel::Snapshot>> recentSnapshots;
```
alxbilger:
```suggestion
    std::vector<std::string> m_recentSnapshotFiles;
    std::map<std::string, std::shared_ptr<sofa::core::objectmodel::Snapshot>> m_recentSnapshots;
```

Sofa/framework/Core/simutest/objectmodel/Snapshot_test.cpp
```diff
@@ -0,0 +1,399 @@
...
+    SOFA_CLASS(TestComponent, BaseComponent);
+
+    Data<float> d_value;
+    sofa::MultiLink<TestComponent, BaseComponent, BaseLink::FLAG_DOUBLELINK> l_target;
+
+    TestComponent()
+        : d_value(initData(&d_value, 3.14f, "value", "test value"))
+        , l_target(initLink("target","target test"))
+    {
+        this->setName("pi");
```
alxbilger: here you set the name of the component to `pi`, not the Data. But in the test you want to verify that the Data name is `pi`

Sofa/framework/Core/src/sofa/core/objectmodel/BaseLink.h
```diff
@@ -134,6 +134,8 @@ class SOFA_CORE_API BaseLink
     /// Read the command line
     bool read( const std::string& str );
 
+    bool readFromSnapshot( const std::string& str );
```
alxbilger, line 138: Add a documentation please

Sofa/framework/Core/src/sofa/core/objectmodel/BaseNode.h
```diff
@@ -128,6 +129,11 @@ class SOFA_CORE_API BaseNode : public virtual Base
     virtual core::collision::Pipeline* getCollisionPipeline() const;
     virtual core::visual::VisualLoop* getVisualLoop() const;
 
+protected:
+    std::shared_ptr<Snapshot::SnapshotObject> createSnapshotObject(std::vector<std::shared_ptr<Snapshot::SnapshotNode>>& parents) const override;
+public:
+    std::shared_ptr<Snapshot::SnapshotObject> findSnapshotObject(const std::shared_ptr<Snapshot::SnapshotNode>& parents, const std::string& objectname) override;
```
alxbilger: add a documentation please

Sofa/framework/Core/src/sofa/core/objectmodel/Snapshot.h
```diff
@@ -0,0 +1,126 @@
...
+namespace sofa::core::objectmodel
+{
+
+/**
+*  \brief Class for snapshot
+*
+*  This class contains the structure for a snapshot of a simulation in SOFA.
+*  The snapshot contains data and link, and keep the shape of a scene graph
+*/
+
```
alxbilger: remove empty line between documentation and class declaration

Sofa/framework/Core/src/sofa/core/objectmodel/SnapshotJSONExporter.h
```diff
@@ -0,0 +1,51 @@
...
+{
+
+    /// Export a single Snapshot to a JSON file
+    void exportToJSON(Snapshot& snapshot, const std::string& filename);
+
+    /// Import a single Snapshot from a JSON file
+    void importFrom(Snapshot& snapshot, const std::string& filename);
+
+    /// Read a JSON file and returns its content as a string
+    std::string file_To_String(const std::string& filename);
```
alxbilger: use a consistant formatting: `fileToString`

Sofa/framework/Core/src/sofa/core/objectmodel/SnapshotManager.h
```diff
@@ -0,0 +1,48 @@
...
+*  This class contains methods and elements that will help to manage snapshot in memory.
+*/
+class SnapshotManager
+{
+public:
+    SnapshotManager();
+    ~SnapshotManager();
+
+    std::vector<std::string> m_recentSnapshotFiles;
+    std::map<std::string, std::shared_ptr<sofa::core::objectmodel::Snapshot>> m_recentSnapshots;
```
alxbilger: if you only use static methods, I am not sure the non-static data members are used. So I would remove them.

Also, I am not sure about the design of the class. I don't think that this class manages anything. You seem to store all the data elsewhere.

Lucas-TJ: This class is here to store snapshots when I save in memory. I use it in SofaImGUI.
I will work on it again to make it clearer.

Sofa/framework/Core/simutest/objectmodel/Base_test.cpp
```diff
@@ -36,6 +36,9 @@ using sofa::defaulttype::Rigid3Types;
 #include <sofa/defaulttype/VecTypes.h>
 using sofa::defaulttype::Vec3Types;
 
+#include <sofa/simulation/SaveSnapshotVisitor.h>
+using sofa::simulation::SaveSnapshotVisitor;
```
bakpaul: It doesn't seem to be used anywhere in this cpp file. Is it necessary ?

Sofa/framework/Core/simutest/objectmodel/Snapshot_test.cpp
```diff
@@ -0,0 +1,414 @@
...
+ *
+ * This test verifies that saveSnapshot save the data to a previously saved snapshot.
+ *
+ * Test steps:
+ * 1. Create a component (Component) and a snapshot
+ * 2. Save Component1's data in the snapshot
+ * 3. Verify if the snapshot contains all the data from Component1
+ *
+ */
+TEST_F(Snapshot_test, saveSnapshot)
```
bakpaul, line 266: I've got an issue with this test, it creates a new object, snapshot it and verifies the values. The issue being that those values are the default ones. This would work if snapshot took the default values and not the actual ones. Could you modify a set of values before the snapshot ?

Sofa/framework/Core/src/sofa/core/objectmodel/BaseLink.cpp
```diff
@@ -323,6 +323,83 @@ bool BaseLink::read( const std::string& str )
...
+            linkPathsFromSnapshot.push_back(linkString);
+    }
+
+    /// From the link
+    std::vector<std::string> linkPathsFromLink;
+
+    std::string linkPath = this->getValueString();
+
+    std::string search = "//";
+    sofa::helper::replaceAll(linkPath, search,"");
```
bakpaul: Why replacing // by nothing ?

Sofa/framework/Core/src/sofa/core/objectmodel/BaseLink.cpp
```diff
@@ -323,6 +323,83 @@ bool BaseLink::read( const std::string& str )
...
+    
+    std::istringstream istrLink(linkPath);
+    std::string linkStringBis;
+
+    while (istrLink >> linkStringBis)
+    {
+        if (!linkStringBis.empty() && linkStringBis[0] == '@')
+            linkPathsFromLink.push_back(linkStringBis);
+    }
```
bakpaul, line 361: No error or warning when a link doesn't have @ and is ignored ?

Sofa/framework/Core/src/sofa/core/objectmodel/Snapshot.h
```diff
@@ -0,0 +1,125 @@
...
+    struct LinkInfo
+    {
+        std::string name;
+        std::string type;
+        std::string value;
+    };
+
+    struct SnapshotObject
+    {
```
bakpaul, line 59: MAybe a stupid question, but I am wondering what happnds to slave objects ? Are they also children of the node of the owner ? Or they are skipped by this design ?

Sofa/framework/Simulation/Core/src/sofa/simulation/LoadSnapshotVisitor.cpp
```diff
@@ -0,0 +1,58 @@
...
+Visitor::Result LoadSnapshotVisitor::processNodeTopDown(simulation::Node* node)
+{
+    const auto snapshotObject = node->findSnapshotObject(m_snapshotContainer.m_graphRoot, node->getName());
+    if (snapshotObject)
+    {
+        const auto SnapshotNode = std::dynamic_pointer_cast<core::objectmodel::Snapshot::SnapshotNode>(snapshotObject);
+        node->loadSnapshot(SnapshotNode);
+        for (simulation::Node::ObjectIterator it = node->object.begin(); it != node->object.end(); ++it)
+        {
+            this->processObject(it->get(), SnapshotNode);
```
bakpaul, line 61: Regarding my comment on the slave object, it could be solved here. But it might require a little modification of the objectSnapshot structure.

Sofa/framework/Simulation/Core/src/sofa/simulation/SnapshotManager.h
```diff
@@ -0,0 +1,95 @@
...
+ * both follow the same overall design:
+ *
+ * - doMemorySave() and doMemoryLoad() save and restore snapshots in memory.
+ * - doSaveTo() and doLoadTo() save and load snapshots from disk. The output file format can be specified when calling doSaveTo().
+ *
+ * The doSaveTo() function also provides an isSet parameter, while doLoadToSet() is used to load a set of snapshots.
+ * These functions allow sets of snapshots to be saved to and restored from a single file.
+ *
+ */
+class SOFA_SIMULATION_CORE_API SnapshotManager
```
bakpaul, line 58: What's the use of this class ? It doesn't seem to be used anywhere and it is not a SOFA component.

bakpaul (commented):
Nice work ! I have two main comments : 
1. This PR seems to be based on #6129 and #6130. If so, please stfate it in the PR description and add the right tag to the PR. 
2. I saw many places where you didn't use constness indicator but the object was never modified. I know this might not be the most heavy load work of all time, but it might help some optimization when dealing with strings in loops.

Sofa/framework/Core/simutest/objectmodel/Base_test.cpp
Lucas-TJ: Oh, no it is not necessary... I correct it !

Sofa/framework/Core/simutest/objectmodel/Snapshot_test.cpp
Lucas-TJ: In fact, the goal of this test is just to verify that ```saveSnapshot()``` correctly saves the data into a snapshot. 
At the beginning, the snapshot is completely empty. After calling ```saveSnapshot()```, the data values are stored in the snapshot. That is what I want to verify.
Of course, I can also change values before ```saveSnapshot()```, and verify if the snapshot takes the changes into account.
I add some assertion macros to verify if the snapshot exist and is empty.

Sofa/framework/Core/src/sofa/core/objectmodel/BaseLink.cpp
Lucas-TJ: At the beginning, this was a design choice related to how snapshots were saved. I chose to save paths without the  // (for example, @Object1 instead of @//Object1). However, maybe it's now fine to keep the //, so I'll update it and try if it works correctly.

Sofa/framework/Simulation/Core/src/sofa/simulation/SnapshotManager.h
Lucas-TJ: SnapshotManager is currently used by SofaImGui. It provides all the functions required to manage snapshots, but it is intended to be used from a user interface such as SofaImGui, which is responsible for calling these functions.

Sofa/framework/Core/src/sofa/core/objectmodel/BaseLink.cpp
damienmarchal: Hi all,

This discussion remind me a vague issue in sofa where the "root" node is not properly serialize by all code path, because the name of the root node and the initiale "/" are actually the same thing. 

Regarding the PR, I'm curious to know in which circumstance there could be multiple "//" in path serialization (expect the corner case of the root node) ? To me this indicates that some name are "empty", removing the double // to / change the "meaning" of the path. So yes, please keep it.

Sofa/framework/Core/src/sofa/core/objectmodel/Snapshot.h
damienmarchal: To be confirmed but, if slaves are not children of a node, then the snapshot system must handle them with a dedicated code path. In PR description slaves are not mentionned... but I have see other @bakpaul comment about the slave topic, so I will have the solution in a subsequent file.

Sofa/framework/Core/simutest/objectmodel/Snapshot_test.cpp
bakpaul: Maybe adding another test to test this might be better then.

Sofa/framework/Core/src/sofa/core/objectmodel/Snapshot.h
Lucas-TJ: Actually the design doesn't take slave objects account. I will modify and apply your suggestions.
Thank you for your remarks !

Sofa/framework/Simulation/Core/src/sofa/simulation/SnapshotManager.cpp
```diff
@@ -0,0 +1,145 @@
...
+* Contact information: contact@sofa-framework.org                             *
+******************************************************************************/
+#include <sofa/simulation/SnapshotManager.h>
+
+#include <sofa/simulation/LoadSnapshotVisitor.h>
+#include <sofa/simulation/SaveSnapshotVisitor.h>
+#include <fstream>
+#include <sofa/helper/logging/Messaging.h>
+#include <algorithm>
+#include <sofa/core/objectmodel/SnapshotJSONExporter.cpp>
```
fredroy, line 29: 😬
```suggestion
#include <sofa/core/objectmodel/SnapshotJSONExporter.h>
```

Sofa/framework/Simulation/Core/src/sofa/simulation/LoadSnapshotVisitor.cpp
```diff
@@ -0,0 +1,67 @@
...
+    {
+        obj->loadSnapshot(snapshotObject);
+        obj->loadInternalStateFrom(*snapshotObject);
+
+        if (!snapshotObject->m_objects.empty())
+        {
+            for (auto& it : snapshotObject->m_objects)
+            {
+                auto objSlave = obj->getSlave(it->m_name);
+                objSlave->loadSnapshot(it);
```
fredroy: need a check if nullptr

fredroy (changes requested):

fredroy (approved):

PR #6181, opened 2026-07-13
[Config] Cmake: Skip dependency re-fetch when already at requested reference

fredroy:
`sofa_fetch_dependency now checks the existing checkout's HEAD against the requested GIT_TAG and skips the nested CMake configure+build (used to drive ExternalProject) when they match, re-fetching only when a dependency's reference actually changes. Cuts a warm reconfigure from ~41s to ~13s without relying on FETCHCONTENT_FULLY_DISCONNECTED.`

I asked Claude to do it because I always forget that I set FETCHCONTENT_FULLY_DISCONNECTED (off or on) and then having some fetch problems later.
So now it is almost the same speed as with FETCHCONTENT_FULLY_DISCONNECTED but will still update/fetch if needs be.


bakpaul:
Ok now this works, but is way more complexe. What do you think ?

bakpaul:
Ok I simplified it by removing the tag part, if it is not a hash (I also remove the introduced constraint on hash length) then it'll always check the remote tip of the current ref. 

Only downside is that now a git ls-remote is always performed when the local ref correspond to what's asked, even for bare hash (this was to avoid forcing hash to be of full 40 digits length, which might not be compatible with all git servers). 

A way to avoid this is to check if the tag is only made of hex chars. But this is weak and might lead to false positive (branch name only using hex char and will thus be skipped)

Tell me if this reduces sufficiently your CMake process. In the end this is a good compromise between code readability/maintainability and feature completeness.

hugtalbot (approved):

Issue #6182, opened 2026-07-15
Sofa.Defaulttype_test crashes


fredroy:
and it is not raised as an error by the CI ⚠️

```
free(): invalid pointer

########## SIG 6 - SIGABRT: usually caused by an abort() or assert() ##########
  sofa::helper::BackTrace::sig(int)
  pthread_kill
  raise
  abort
  free
  on_exit+0
  __libc_start_main
```

The crash happens when the DatatypeInfoRegistry wants to delete some typeinfo(?) which has been (presumably) set by the Datatypeinfo Mock (or something like that)

PR #4838, opened 2024-07-19
[All] Reduce use of raw pointers

alxbilger:
By doing that I believe that it also removes some memory leaks

______________________________________________________

By submitting this pull request, I acknowledge that  
**I have read, understand, and agree [SOFA Developer Certificate of Origin (DCO)](https://github.com/sofa-framework/sofa/blob/master/CONTRIBUTING.md#sofa-developer-certificate-of-origin-dco)**.
______________________________________________________

**Reviewers will merge this pull-request only if**  
- it builds with SUCCESS for all platforms on the CI.
- it does not generate new warnings.
- it does not generate new unit test failures.
- it does not generate new scene test failures.
- it does not break API compatibility.
- it is more than 1 week old (or has fast-merge label).

PR #6183, opened 2026-07-15
[Defaulttype] TypeInfoRegistry: Fix crash at exit in TypeInfoRegistry deleter

fredroy:
Fix Issue #6182 

By Claude:

The root cause is the storage deleter in `TypeInfoRegistry::getStorage()`, which calls `delete info` on **every** stored pointer except `NoTypeInfo::Get()`. However the registry holds two kinds of pointers:
 - `NameOnlyTypeInfo` objects it allocates itself with `new` (in   `AllocateNewTypeId`) ‚ genuinely heap-owned.
- Pointers registered via `Set()`, e.g. `DataTypeInfoDynamicWrapper<...>::get()`  (and `VirtualTypeInfo<T>::get()` used by `Data::getValueTypeInfo()`), which  return **function-local statics**.

Deleting those static-storage singletons is undefined behavior (invalid`delete` + double destruction, since the statics also self-destruct via their own `atexit` handlers), which triggers the trap. The deleter already special-cased `NoTypeInfo::Get()` as a singleton but missed that every info registered via `Set()` is a singleton too.

Fix: Only free what the registry actually allocates:

- `getStorage()` now holds **non-owning** borrowed pointers (no delete-all deleter).
- A separate `getOwnedStorage()` (`vector<unique_ptr<const AbstractTypeInfo>>`)
  owns the `NameOnlyTypeInfo` objects created in `AllocateNewTypeId`.

This preserves the leak-free intent of the original deleter (#4838) while never deleting the borrowed static singletons registered through `Set()`.

[with-all-tests]

### DIFF
--- a/Sofa/framework/DefaultType/src/sofa/defaulttype/TypeInfoRegistry.cpp
+++ b/Sofa/framework/DefaultType/src/sofa/defaulttype/TypeInfoRegistry.cpp
@@ -21,6 +21,7 @@
 ******************************************************************************/
 #include <typeindex>
 #include <map>
+#include <memory>
 #include <iostream>
 #include <sofa/helper/logging/Messaging.h>
 #include <sofa/defaulttype/AbstractTypeInfo.h>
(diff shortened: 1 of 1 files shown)


bakpaul:
[ci-build][with-all-tests]

hugtalbot (approved):

PR #6196, opened 2026-07-21
[FEM] Add component for source term integration based on new FEM framework - constant source terms

### LINKED ISSUE #6166: Add new SourceTermFEMForceField
This would allow to add a constant field integrated over FE
In parallel a NodalConstantForceField could also be added

th-skam:
### PR Based on #6225 

The PR adds component with which to integrate source terms using the FEM framework ported from Elasticity. 

There is a central `FEMSourceTermIntegrator` component placed in the scene responsible for gathering all source terms present, and integrating them into loads. Added examples demonstrate this. 

I added 3 distinct classes that can implement the following loads:
| The type of the term | Formula | User input |
| - | - | - |
| Pressure | $L_p = \int_{S} \vec{v} \cdot (-p \vec{n}) dS$ | The pressure $p$ on the boundary |
| Stress | $L_{\sigma} = \int_{S} \vec{v} \cdot (\sigma \vec{n}) dS$ | The stress tensor $\sigma$ |
| Force | $L_{t} = \int_{S} \vec{v} \cdot \vec{t} dS$ | The force density $t$ directly |

The last object can also mimic body forces (gravity etc).

Currently in SOFA, these loads were spread out among various components, such as `SurfacePressureForceField`, `QuadPressureForceField`, `EdgePressureForceField`, `TrianglePressureForceField` among others. Each component implements their own integration loop and uses geometric data (such as normals) directly from its linked topology. 
The loop is unified in the `FEMSourceTermIntegrator` and data such as normals are computed from the mapping jacobian of the reference domain. Extending them to quadratic elements is straightforward by using a quadratic topology in the scene; no more code extensions.

The scenes that have been added demonstrate how the new source term classes can duplicate the existing functionality. 
In this PR, the loads are applied on the reference geometry. That's to simply introduce them and I'll add the ability to follow the geometry in subsequent PR.

Fixes https://github.com/sofa-framework/sofa/issues/6166

[force-full-build]
[with-all-tests]


bakpaul:
I am not a FEM expert so I don't feel confident to review this PR. Nevertheless I checked to make sure this is compatible with higher order element and it seems ton be the case. So it is fine by me given that @alxbilger and/or @hugtalbot review it.

Could you add an example ?

th-skam:
PR now depends on: https://github.com/sofa-framework/sofa/pull/6225

th-skam:
Two examples were added. 
 - `ConstantSourceTerm.scn` just adds gravity as a source term
 - `FEMSourceTermIntegrator` comes from the Method of Manufactured Solutions. It applies a load to make the solver converge to the manufactured solution. It also shows the difference between the quadrature degree:

White is the manufactured solution, green/red use quadrature degree = 2 and 1, respectively, to integrate the source term into nodal loads. 
Result is trivial here but the option to choose the degree is important when computing errors in MMS. 

Screenshot 2026-09-01 161712

The source terms are considered constant here. Non-constant ones in upcoming PR. 
Tests were generated by AI.

PR #6204, opened 2026-07-24
[RigidContactMapper] Take radius into account for spheres

bakpaul:
I don't know why but this was commented since 2013. Thus, the SphereCollisionModel was broken for rigids since then. 
This restores it. I suspect that many regression scene will fail, let's see... 

[with-all-tests]


alxbilger (approved):
On top of that, it's not easy to understand what `r` is, and why it should be the radius.

alxbilger:
It appears that no test is affected. This code is not covered?

bakpaul:
Well it seems like no regression test is covering this. I stumbled upon this by messing around with this scene https://github.com/sofa-framework/sofa/blob/master/examples/Component/Collision/Detection/CompositeCollisionPipeline_none.scn, I would have never saw this if I didn't have another issue with this one.

PR #6209, opened 2026-07-27
[Type] Introduce clamped scalar ensuring value not outside bounds

alxbilger:
This pull request introduces a new foundational type, sofa::type::ClampedScalar<T>. This class template wraps a standard scalar value and guarantees that the stored value cannot fall outside predefined minimum and maximum bounds (min/max).

### Key Changes & Implementation Details

1.  **New Type (`ClampedScalar`):**
    *   A templated class that encapsulates a scalar value and two immutable bounds ($\text{minBound}, \text{maxBound}$).
    *   The class implements clamping logic in several key areas:
        *   **Construction:** Initial values are automatically clamped upon object creation.
        *   **Assignment/Setting (`operator=`):** Any attempt to set the value uses clamping, ensuring state integrity.
        *   **Input/Output Operators:** Stream extraction (`>>`) reads a raw value and clamps it before updating the internal state. Stream insertion (`<<`) only outputs the valid, clamped value.
        *   **Prediction:** A `calculateClamped(T rawValue)` method allows users to predict what the bound-constrained result would be without modifying the object's actual state.

2.  **Integration into `VisualMesh` Component:**
    *   The `d_elementSpace` data member in `sofa::component::visual::VisualMesh` has been updated from `Data<SReal>` to **`Data<type::ClampedScalar<SReal>>`**.
    *   This change ensures that the element space parameter (which must logically be between 0 and 1) is enforced and clamped correctly, making the mesh visualization component more robust.

3.  **Framework Updates:**
    *   Added `ClampedScalar.h`, `ClampedScalar.cpp`, and corresponding test files (`ClampedScalar_test.cpp`) to `Sofa/framework/Type`.
    *   Updated CMake files across the framework to correctly include and build this new type.

⚠️ Breaking because of the change of type in `VisualMesh`.

Companion PR for the GUI: https://github.com/sofa-framework/SofaGLFW/pull/292

https://github.com/user-attachments/assets/464b4a3c-e3fe-408c-a023-59c54d9ae801

Local gemma-4-e4b used for documentation and unit tests.

[with-all-tests]

### DIFF
--- a/Sofa/Component/Visual/src/sofa/component/visual/VisualMesh.cpp
+++ b/Sofa/Component/Visual/src/sofa/component/visual/VisualMesh.cpp
@@ -35,7 +35,7 @@ void registerVisualMesh(sofa::core::ObjectFactory* factory)
 
 VisualMesh::VisualMesh()
     : d_position(initData(&d_position, "position", "The position of the vertices of mesh"))
-    , d_elementSpace(initData(&d_elementSpace, 0.15_sreal, "elementSpace",
+    , d_elementSpace(initData(&d_elementSpace, {0.15_sreal, {0_sreal, 1_sreal}}, "elementSpace",
                               "The space between element (scalar between 0 and 1)"))
     , d_lighting(initData(&d_lighting, true, "lighting", "If true, light is simulated on the mesh. Otherwise, no lighting effect."))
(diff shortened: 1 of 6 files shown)


fredroy (approved):
Should a ClampedIntegral be implemented also ?

alxbilger:
> Should a ClampedIntegral be implemented also ?

I think that `int` is supported as a template parameter of `ClampedScalar`.

PR #6216, opened 2026-07-31
[FEM] Compute and draw von Mises stress

alxbilger:
This PR introduces the computation and visualization of von Mises stress, based on the FEM. It consists in the local least-squares projection of stress values from quadrature points to nodes. The projection is local: the value at the node depends on the considered element, i.e. a global node has multiple values depending on its neighbour element. The result is a discontinuous stress field. It's on purpose! A discontinuity would indicate a problem of discretization. Without any problem, the field should appear continuous.

## The theory

In FEM, the Cauchy stress is approximated in any point of an element by:

$$
\sigma_\text{approx}(\xi) = \sum_i N_i(\xi) \sigma_i
$$

where $\sigma_i$ are the values at nodes, $N_i$ are the shape functions.

$\sigma_i$ are unknowns and we want to compute them. For that, we introduce the error function:

$$
\rho(\sigma_0, \sigma_1, ...) = \frac{1}{2} \int_\Omega( \sum_i N_i(\xi) \sigma_i - \sigma )^2 d\Omega
$$

We want this error to be minimal. Its minimum is when its derivative is zero:

$$
\int_\Omega N_i(\xi)  \left( \sum_i N_i(\xi) \sigma_i - \sigma \right) d\Omega = 0
$$

We rearrange:

$$
\sum_j \left( \int_\Omega N_i N_j d\Omega \right) \sigma_j = \int_\Omega N_i \sigma d\Omega
$$

that we can write as $M \sigma_j = b$. It's a linear system to solve. The solution gives the stress values at the nodes.

The $M$ matrix is the Gram matrix (mentioned in the code). It's constant over time, so it's pre-computed. In order to solve the linear system, we store the inverse of this matrix. The algorithm simply consists in assembling $b$ and multiply it by $M^{-1}$.

Note: for the future, one can think about smoothing the field to make it continuous. There are two ways:
1) Average the stress values around each node. It can be weighted by the volume of the element.
2) Instead of solving a local system, we assemble a global system to project the stress values in a global manner. It means assembling the global Gram matrix, which is sparse. It would require a linear solver, or a pre-computation of its decomposition for $M^{-1} * b$.

https://github.com/user-attachments/assets/92c4d35e-bcb2-4f74-b0a9-d9b24cfe8723

VonMisesStress_00001

For future PRs:

- `DrawElementColoredMesh` is not implemented for all types of elements.
- ~Only `LinearSmallStrainFEMForceField` derives from `CauchyStressEvaluator`. The next candidate to support stress evaluation is naturally `CorotationalFEMForceField`. And `HyperelasticMaterial` once https://github.com/sofa-framework/sofa/pull/6206 is merged.~

https://github.com/user-attachments/assets/dc226eac-736d-4d14-bd48-546c5cab1931

[with-all-tests]


alxbilger:
[ci-build][with-all-tests]

Sofa/Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/VonMisesStress.inl
```diff
@@ -0,0 +1,369 @@
...
+
+                // shape functions in the reference element evaluated at the quadrature point
+                const auto N = FiniteElement::shapeFunctions(quadraturePoint);
+
+                const auto NT_N = sofa::type::dyad(N, N);
+
+                elementMassMatrix += (weight * detJ) * NT_N;
+            }
+
+            sofa::type::invertMatrix(elementMassMatrix, elementMassMatrix);
```
fredroy, line 236: Nice finding from Claude: invertMatrix() is not safe when giving the same matrix for in and out (dest/from),  for 2x2 and 3x3 (specialized versions).
https://github.com/alxbilger/sofa/blob/a1395f7adce413f1ce1966e9e4055c2ad0334545/Sofa/framework/Type/src/sofa/type/Mat.h#L1198
So ElementType = Edge or Triangle it is wrong
(the generic invertMatrix is fine)

Sofa/Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/LinearSmallStrainFEMForceField.inl
```diff
@@ -180,4 +180,32 @@ void LinearSmallStrainFEMForceField<DataTypes, ElementType>::addKToMatrix(
...
+auto LinearSmallStrainFEMForceField<DataTypes, ElementType>::computeStress(
+    const DeformationGradient& F, sofa::Size elementId) -> StressVoigtVector
+{
+    const auto strainTensor = static_cast<Real_t<DataTypes>>(1)/2 * (F + F.transposed())  - DeformationGradient::Identity();
+
+    sofa::type::Vec<type::NumberOfIndependentElements<trait::spatial_dimensions>, Real_t<DataTypes> > strainVoigt;
+    for (sofa::Size i = 0; i < type::NumberOfIndependentElements<trait::spatial_dimensions>; ++i)
+    {
+        const auto [p, q] = type::toTensorIndices<trait::spatial_dimensions>(i);
+        strainVoigt[i] = strainTensor(p, q);
```
fredroy:
```
computeStress() feeds tensorial (halved) shear strain into a Voigt elasticity matrix the rest of the codebase calibrates for engineering shear strain. Every off-diagonal stress component comes out at exactly half its physical value. This affects every element under any shear/bending/torsion — the one loading case (pure uniaxial/hydrostatic) that would look fine in a quick demo is the exception, not the rule.
```
(not an expert so copying Claude verbatim)

Sofa/Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/VonMisesStress.inl
```diff
@@ -0,0 +1,369 @@
...
+
+        this->precomputeData();
+        this->calculateElementInverseGramMatrices(elements, m_elementInverseGramMatrices);
+    }
+}
+
+template <class DataTypes, class ElementType>
+void VonMisesStress<DataTypes, ElementType>::handleEvent(core::objectmodel::Event* event)
+{
+    if (simulation::AnimateEndEvent::checkEventType(event))
```
fredroy, line 78:
```suggestion
    if (this->isComponentStateInvalid())
    {
         return;
    }
    if (simulation::AnimateEndEvent::checkEventType(event))
```

fredroy (changes requested):
Did a quick pass through Sonnet....

EDIT: https://github.com/sofa-framework/sofa/issues/6273

Sofa/Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/VonMisesStress.inl
alxbilger: Good catch! Fixed in https://github.com/sofa-framework/sofa/pull/6275, so I resolve this comment.

Sofa/Component/SolidMechanics/FEM/Elastic/src/sofa/component/solidmechanics/fem/elastic/LinearSmallStrainFEMForceField.inl
alxbilger: I made the changes https://github.com/sofa-framework/sofa/pull/6216/changes/2c8cd712bc54a57e435384cf58eefa34b3c64c7c

fredroy (approved):

PR #6223, opened 2026-08-07
[Constraint][Correction] Enable PrecomputedConstraintCorrection to operate with preconditioned systems

th-skam:
This is a component that can be very useful for real-time simulations, however it has been neglected. 

It pre-computes the compliance matrix by perturbing the dofs with a unit force. 

Problems:
- For accurate computations, when using an iterative solver, stricter tolerances must be set during the perturbations. This was only done with `CGLinearSolver` during the perturbations. It now covers any solver that has `tolerance` and `iterations` as Data. Still fragile. A common API for iterative solvers could be useful. 
- With preconditioned solvers (such as PCG) the precomputation might break because it relies on finding the linearSolver through `getContext`. This can return either the preconditioner or the actual solver, depending on how the user set them up in the scene graph. I added a link instead to explicit this and a scene to demonstrate it.

[with-all-tests]
[ci-depends-on https://github.com/sofa-framework/Regression/pull/127]

### DIFF
--- a/Sofa/Component/Constraint/Lagrangian/Correction/src/sofa/component/constraint/lagrangian/correction/PrecomputedConstraintCorrection.h
+++ b/Sofa/Component/Constraint/Lagrangian/Correction/src/sofa/component/constraint/lagrangian/correction/PrecomputedConstraintCorrection.h
@@ -23,7 +23,7 @@
 #include <sofa/component/constraint/lagrangian/correction/config.h>
 
 #include <sofa/core/behavior/ConstraintCorrection.h>
-#include <sofa/core/behavior/OdeSolver.h>
+#include <sofa/component/odesolver/backward/EulerImplicitSolver.h>
 #include <sofa/core/behavior/LinearSolver.h>
 #include <sofa/core/objectmodel/DataFileName.h>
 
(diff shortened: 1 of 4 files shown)


alxbilger (commented):
Thanks for the PR. It makes sense. Excellent that you added an example scene. In the future, I recommand to have regression tests on it.

In the scene test, since you are working with hexas, I suggest to use the components `FEMMass` and `CorotationalFEMForceField`.

fredroy:
for windows
[ci-build][with-all-tests]

bakpaul:
I understand the intend but using the base class for ODE will not make it compatible will all ODE. This is because of the way it computed the compliance : it looks at the velocity to get one column of the inverted matrix, meaning that it expect the ODE solver to work in velocity, and currenlty the only one doing so are EulerImplicit and the LinearMultiStep ones. The latter ones are  also incopatible because they keep an internal track of the states up to n timestep, n being their order. So given the current implementation, the only real compatible ODE is indeed the Implicit Euler. It coul dbe solved, but the issue is that to be agnostic of the type state (either solved in acceleration or velocity) you need a bt more changes than that I expect, because you'll need coherency in all the factors that are everywhere coping for the integration order and so on. I intend to refactor all the constraint corrections after my work on the newton, but if you want to tackle it before you're welcome ! 

So this is possible but I would suggest doing it in another PR and just keep here the mechanism that deal with any kind of linear solver.

bakpaul (approved):

fredroy:
[ci-build][with-all-tests]

PR #6225, opened 2026-08-10
[FEM] Introduce higher accuracy quadrature rules

th-skam:
For each Element, an option is added to use a higher-accuracy quadrature rule. The default behaviour of existing force fields is preserved. Lower-order rules are also included for completeness.

The purpose of this change is to enable more accurate error-norm computations in the V&V suite using the Method of Manufactured Solutions.

If the accuracy is insufficient, the integration introduces errors in the computed norms. This we observe either as the convergence rate approaching the expected one more slowly or as apparent superconvergence that counters the purpose of the test. 

The degrees are:
| Element | Up to degree | Default degree |
| - | - | - |
| Edge | 3 | 1 |
| Quad | 3 | 2 |
| Tri | 2 | 1 |
| Tetra | 2 | 1 |
| Hex | 5 | 3 |

[with-all-tests]


alxbilger (commented):
I would be curious to know how you use `quadratureRule`

th-skam:
@alxbilger 
It was used like so in my cases: https://github.com/sofa-framework/sofa/pull/6196/changes/97581c56a994ad944f8382a12372f1ce5b7eba14

alxbilger (commented):
Could you just add the new quadrature points to the unit tests `testSumWeights` ? It's not enough, but there are not other tests for now...

alxbilger:
I think you can remove the `ci-depends-on`. It's actually the other PR that needs this one.

th-skam:
Now that we added quadratic elements, I refactored the quadrature rules. Linear or quadratic, the integration is done over the same reference domain. The rules are now re-used, implemented in `QuadratureRules.h`. AI was used to help with std usage to have compile-time/runtime paths. 

The literals were also substituted by std:: variants, wherever possible. This has triggered a unit test failure in `ElementMass__test`.

New rules can be added in the existing tables. There are two paths: 
- Compile-time: using the function `quadraturePoints<Degree>`. It has a default value, and it is the one used by the existing FEM force fields
- Runtime: using `quadratureRule` and requesting a degree.  I use this path for the source terms in https://github.com/sofa-framework/sofa/pull/6196

Each element has a default and minimum degree, so quadratic elements cannot use 1st order rules. 
Will work on populating tests, either here or in subsequent PR.

PR #6258, opened 2026-08-25
[LinearSolver.Direct] EigenDirectSparseSolver: Fix out-of-bounds read on matrices with empty rows

fredroy:
advised by Claude while doing
- #6256 

its point:
```
copyNonZeros() does not store the rows which are entirely zero, so the filtered matrix held fewer row offsets than the Eigen::Map built from it reads, and the factorization accessed memory past the CRS arrays. fullRows() now restores those rows before the mapping.

Any remaining mismatch means the matrix contains entries outside its declared size ‚ CompressedRowSparseMatrix grows silently to store them , which is now reported instead of being mapped, the solver being left invalid so that solve() does not use an unfactorized Eigen solver.
```

I also instructed him to make unit tests to cover this case.

[with-all-tests]


alxbilger (commented):
I located a function in `CompressedRowSparseMatrixGeneric.h` that could be widely used wherever a CRS is manipulated, including in the solver of this PR: `check_matrix`. Currently, it's used nowhere. The problem is that this function implementation is suspect, there is no test and no documentation. I suggest to have a common tool (such as this function) to avoid testing code wherever a matrix is used.

Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/EigenDirectSparseSolver.inl
```diff
@@ -48,6 +48,12 @@ void EigenDirectSparseSolver<TBlockType, EigenSolver>
 {
     SOFA_UNUSED(A);
 
+    if (this->isComponentStateInvalid())
+    {
+        // the matrix could not be factorized: the Eigen solver is not usable
```
hugtalbot:
```suggestion
```

hugtalbot (approved):

bakpaul (approved):

PR #6259, opened 2026-08-25
[Sofa.Simulation.Core] MappingGraph: fix memory leak in from parent/child shared_ptr cycle

### LINKED ISSUE #6230: Memory leak
It seems like a memory leak has been introduced in https://github.com/sofa-framework/sofa/pull/6087. 
Just run the default scene in batch mode and you'll see the problem.

fredroy:
From Claude (sonnet 😅):

`BaseMappingGraphNode` stored both `m_parents` and `m_children` as owning `shared_ptr`s. `MappingGraph::addEdge()` linked every connected pair in both directions, creating a reference cycle between each parent and child node.
`MappingGraph::clear()` (called at the start of every `build()`) only drops the graph's own `m_allNodes` vector; it never breaks these cycles. Since `EulerExplicitSolver::solve()` and `EulerImplicitSolver::solve()` call
`m_mappingGraph.build()` unconditionally on every timestep, the entire previous graph leaked on every single animation step for as long as the solver was active.
Fix: store `m_parents` as `std::weak_ptr` instead of `shared_ptr`. Node ownership already belongs to `MappingGraph::m_allNodes` (and to each node's owning parent via `m_children`), so the back-reference doesn't need to be owning.

| Solver step | Live nodes (before fix) | Live nodes (after fix) |
|---:|---:|---:|
| 1 | 0 | 0 |
| 101 | 1200 | 24 |
| 1001 | 12000 | 24 |
| 1901 | 22800 | 24 |

--- 

Basically the shared_ptr retained a ref so the parents/childs were owning each other, so the shared_ptr/gc mechanism were effectively dismissed.

Fix #6230 

[with-all-tests]

### DIFF
--- a/Sofa/framework/Simulation/Core/src/sofa/simulation/MappingGraph.cpp
+++ b/Sofa/framework/Simulation/Core/src/sofa/simulation/MappingGraph.cpp
@@ -91,11 +91,14 @@ MappingGraph::MappingInputs MappingGraph::getTopMostMechanicalStates(
                 current->accept(visitor);
             }
 
-            for (auto& parent : current->m_parents)
+            for (auto& weakParent : current->m_parents)
             {
-                if (parent->m_pendingCount == 0)
+                if (const auto parent = weakParent.lock())
                 {
-                    nodes.push(parent.get());
+                    if (parent->m_pendingCount == 0)
+                    {
+                        nodes.push(parent.get());
+                    }
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


fredroy:
for windows 
[ci-build][with-all-tests]

alxbilger (commented):
Should the children be weak pointers too ? They are supposed to be in `m_allNodes` too.

fredroy:
> Should the children be weak pointers too ? They are supposed to be in `m_allNodes` too.

weak_ptr does apparently involve 1 test and 1 atomic refcount for every use so from a pure POV of perf, it would be better to keep like this.
But behavior-wise, it  would be similar yes

alxbilger (approved):

PR #6260, opened 2026-08-25
[Simulation] Mapping graph: restrict construction of edges between groups and mappings

alxbilger:
This PR fixes a crash in `examples/Component/LinearSystem/MatrixLinearSystem.scn`.

image

This is complex scene showcasing the correct behavior of mappings in the matrix assembly.

This is the path to locate the bug:

1. The solver crashes because of a bad CSR matrix construction (which leads me to think that we don't have enough tools to analyze those matrices, and that this data structure is still fragile). 
⬇️ 
2. The matrix is not well constructed because of bad matrix projections (`MatrixProjectionMethod.inl`).
⬇️ 
3. The matrices are not well projected because `mappingGraph.getTopMostMechanicalStates` gives wrong inputs. For example, in the scene, it adds an unrelated input to `red/FEM/DOFs`. The unrelated input still contributes to the matrix, potentially to wrong locations.
⬇️ 
4. The wrong inputs are due to a wrong construction of the mapping graph. In the following screenshot, we can observe that the RigidMapping is a child of a group node, itself a child of the blue DoFs (which is unrelated to the mapping).

image

The fix:
Mappings must be child of a group only if all the mapping inputs are all equal to the group's states. Previously, only one common state was enough.

After the fix:

image

The rigid mapping is no longer related to the blue DoFs.

This fix is in addition to https://github.com/sofa-framework/sofa/pull/6222 and https://github.com/sofa-framework/sofa/pull/6256

[with-all-tests]

### DIFF
--- a/Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolverImpl.h
+++ b/Sofa/Component/LinearSolver/Direct/src/sofa/component/linearsolver/direct/SparseLDLSolverImpl.h
@@ -109,7 +109,9 @@ inline void CSPARSE_numeric(int n,int * M_colptr,int * M_rowind,Real * M_values,
         kk = perm[k];  // kth original, or permuted, column 
         for (p = M_colptr[kk] ; p < M_colptr[kk+1] ; p++)
         {
-            i = invperm[M_rowind[p]];	// get A(i,k) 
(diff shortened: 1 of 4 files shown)


alxbilger:
[ci-build][with-all-tests]

fredroy (approved):

PR #6262, opened 2026-08-25
[Example] Add a scene checking the projection of mapped matrices

fredroy:
Claude:

Adds a self-checking example for the assembly of force fields acting on mapped states.

The scene runs the same mechanical system twice: once with a spring acting directly on two top-level states, once with the same spring acting on two identity-mapped children. Since J = I, the projected matrix JTKJ must equal K, so both sub-trees must follow exactly the same trajectory. 
A WriteState on each top-level state records the result, and the two files must be identical.

---

This test is intended to show the validity of #6256 .
The mapped positions are supposed to be identical with the "direct" positions but with master it does not. (and it does with #6256) 

| Configuration | `direct` P1.x | `mapped` Q1.x | Result |
| --- | --- | --- | --- |
| union (before the fix) | 0.457469 | 0.470485 | **DIFFERENT** |
| cartesian product (the fix) | 0.457469 | 0.457469 | identical |
| #6260 + union | 0.457469 | 0.470485 | **DIFFERENT** |
| #6260 + cartesian product | 0.457469 | 0.457469 | identical |

( I did the tests by myself 👀)

The output files does serve as a self-contained regression test but it can be added as a regular regression test as well.

---
EDIT: converted to unit tests, with the following combinations:

| Matrix | Linear system | Projection method | 
| --- | --- | --- | 
| `FullMatrix` | `MatrixLinearSystem` | `MatrixProjectionMethod` | 
| `FullMatrix` | `MatrixLinearSystem` | `ConstantSparsityProjectionMethod` | 
| `CompressedRowSparseMatrixd` | `MatrixLinearSystem` | `MatrixProjectionMethod` | 
| `CompressedRowSparseMatrixd` | `MatrixLinearSystem` | `ConstantSparsityProjectionMethod` | 
| `CompressedRowSparseMatrixd` | `ConstantSparsityPatternSystem` | `MatrixProjectionMethod` | 
| `CompressedRowSparseMatrixd` | `ConstantSparsityPatternSystem` | `ConstantSparsityProjectionMethod` | 

[with-all-tests]
[force-full-build]


examples/Validation/MatrixProjectionEquivalence.scn
```diff
@@ -0,0 +1,90 @@
...
+
+    Because the mappings are identities, J = I, so J^T K J = K. The assembled
+    system matrix of "mapped" must therefore be IDENTICAL to that of "direct",
+    and the two sub-trees must follow exactly the same trajectory.
+
+    Any deviation means the projection of the mapped stiffness added something
+    that is not J^T K J - either a missing coupling term, or a spurious one.
+
+    Run:
+        runSofa -g batch -n 100 projection_equivalence.scn
```
alxbilger:
```suggestion
        runSofa -g batch -n 100 MatrixProjectionEquivalence.scn
```

alxbilger:
Excellent idea to have this scene. I thought I had something similar somewhere. There is `intermediateMapping` in `examples/Component/LinearSystem/MatrixLinearSystem.scn`, but I don't test it numerically.

Can Claude translate it to a C++ unit test? I also would like to apply this test on `ConstantSparsityPatternSystem` instead of `MatrixLinearSystem` and/or `ConstantSparsityProjectionMethod` instead of `MatrixProjectionMethod` (= 4 combinations).

bakpaul (approved):

PR #6270, opened 2026-08-28
[LinearAlgebra] CompressedRowSparseMatrix family: Fix problematic cases

fredroy:
Based on 
- #6269 

Fix the several issues highlighted in the previous PR,

| Commit | Description |
| --- | --- |
| `03b09a946d` | `block()` returned a neighbouring row's value, or read `colsIndex[-1]`, because its first/last-column fast paths never checked the row range for emptiness; extracts the row and column lookups into guarded helpers, in a fast-path and a lean variant |
| `d8cf5dfaa1` | `clearRowBlock()` segfaulted on a matrix with no registered row, calling `rowIndex.back()` before testing for emptiness |
| `308eb34501` | `clearRowColBlock()` ignored its own `foundRowId` flag and passed a stale index to `deleteRow()`, dropping an unrelated row |
| `52652b1731` | `clearRowCol()` indexed `rowBegin` with a block-column number instead of a row position, so it missed entries in the cleared column and zeroed unrelated ones when the symmetric block was absent |
| `628cdfceaa` | `filterValues()` never honoured `keepEmptyRows`, making the flag a no-op across all eight `copy*()` wrappers |
| `b315ab8cfe` | `RowConstIterator`'s members had no initialisers, so a default-constructed iterator read as a valid iterator onto row 0 |
| `11d5ce5c7f` | Routes the dozen remaining open-coded row/column lookups through the guarded helpers, removing unguarded divisions by `nBlockRow`/`nBlockCol` (UB; SIGFPE on x86_64), using the lean variant so assembly is not slowed down |

--> 52652b1731 is the one with real production exposure: clearRowCol is called by FixedProjectiveConstraint, AttachProjectiveConstraint, PointProjectiveConstraint and friends every time step.

[with-all-tests]

### DIFF
--- a/Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixConstraint.h
+++ b/Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixConstraint.h
@@ -430,8 +430,10 @@ public:
 
     private:
 
-        Index m_internal;
-        const CompressedRowSparseMatrixConstraint* m_matrix;
+        /// a default-constructed iterator must report itself as invalid rather
+        /// than alias row 0 of an unspecified matrix
+        Index m_internal = s_invalidIndex;
+        const CompressedRowSparseMatrixConstraint* m_matrix = nullptr;
     };
 
     /// Get the iterator corresponding to the beginning of the rows of blocks
--- a/Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixGeneric.h
+++ b/Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixGeneric.h
@@ -1050,13 +1050,7 @@ public:
         }
 
         Index rowId = 0;
-        if (i == rowIndex.back()) rowId = Index(rowIndex.size() - 1);      /// Optimization to avoid do a find when looking for the last line registred
-        else if (i == rowIndex.front()) rowId = 0;                  /// Optimization to avoid do a find when looking for the first line registred
-        else
-        {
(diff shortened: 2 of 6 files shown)


fredroy:
[ci-build][with-all-tests]

Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixConstraint.h
```diff
@@ -430,8 +430,10 @@ class CompressedRowSparseMatrixConstraint : public sofa::linearalgebra::Compress
 
     private:
 
-        Index m_internal;
-        const CompressedRowSparseMatrixConstraint* m_matrix;
+        /// a default-constructed iterator must report itself as invalid rather
+        /// than alias row 0 of an unspecified matrix
+        Index m_internal = s_invalidIndex;
```
bakpaul: I always though that the SOFA coding style imposes this to be put in the constructor

Sofa/framework/LinearAlgebra/src/sofa/linearalgebra/CompressedRowSparseMatrixGeneric.h
```diff
@@ -422,12 +422,97 @@ public :
...
             if (lastColIndex > maxColIndex) maxColIndex = lastColIndex;
         }
         return maxColIndex;
     }
 
+    /**
+    * \brief Look for column j among the blocks of a single row.
+    *
+    * The first and last registered columns of the row are checked directly, as
+    * they are by far the most common queries; anything else falls back to a
```
bakpaul: I am dubious of this statement, it is more that they are randomly accessible and if they match the column id then it avoids a binary search.

bakpaul (approved):
The code is very hard to read. This is not the fault of the PR but of the original code.

PR #6287, opened 2026-09-01
[Core] MultiVecId tests

alxbilger:
Unit tests mostly generated by Gemini 3.7 Flash.

I had to modify the MultiVecId class:

- Checking the container is not empty in `setId`. Otherwise, an empty container would lead to `hadIdMap` returning `true`, but with an empty map.
- Consistency in `setId` for the class `TMultiVecId` and its specialization for `V_ALL`: they both accept the same type `std::set<State>`. I decided to remove the most generic overload (`template<class StateSet> void setId(const StateSet& states, const MyVecId& id)`), because it causes ambiguity (and compilation errors) when calling `setId` on a `BaseState*` (not a `const BaseState*`).
- `unique()` has been removed in C++20 (https://en.cppreference.com/cpp/memory/shared_ptr/unique), which proves that this code was not even called anywhere. But the unit tests call it.

[with-all-tests]

### DIFF
--- a/Sofa/framework/Core/src/sofa/core/MultiVecId.h
+++ b/Sofa/framework/Core/src/sofa/core/MultiVecId.h
@@ -329,9 +329,12 @@ public:
     template<class State>
     void setId(const std::set<State>& states, const MyVecId& id)
     {
-        IdMap& map = writeIdMap();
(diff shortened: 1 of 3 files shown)


fredroy:
https://github.com/alxbilger/sofa/pull/12

fredroy (approved):

PR #6300, opened 2026-09-09
[Simulation.Core] Differential operations

alxbilger:
This PR introduces fundamental operations in SOFA, formalized to be physics-independent. The vocabulary no longer uses mechanics terms. The terms are more generic. It is also closer to autodiff frameworks.

The functions are based on the mapping graph. As a consequence, they can be used only in codes where the mapping graph is available. As an example, I use them in `MappingGraphMechanicalOperations`. But they should replace all the `propagate*` methods in `MechanicalOperations`. Since those methods are called mainly in integration scheme, I did not want to create conflicts with https://github.com/sofa-framework/sofa/pull/6117. 

To do in a next PR:

- Deprecate all the `propagate*` methods in `MechanicalOperations`.
- Use the methods from `DifferentialOperations` wherever `MechanicalOperations::propagate*` is used.

[with-all-tests]

### DIFF
--- a/Sofa/framework/Simulation/Core/src/sofa/simulation/DifferentialOperations.cpp
+++ b/Sofa/framework/Simulation/Core/src/sofa/simulation/DifferentialOperations.cpp
@@ -1,0 +1,57 @@
+/******************************************************************************
+*                 SOFA, Simulation Open-Framework Architecture                *
+*                    (c) 2006 INRIA, USTL, UJF, CNRS, MGH                     *
+*                                                                             *
+* This program is free software; you can redistribute it and/or modify it     *
+* under the terms of the GNU Lesser General Public License as published by    *
+* the Free Software Foundation; either version 2.1 of the License, or (at     *
+* your option) any later version.                                             *
+*                                                                             *
+* This program is distributed in the hope that it will be useful, but WITHOUT *
+* ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or       *
+* FITNESS FOR A PARTICULAR PURPOSE. See the GNU Lesser General Public License *
+* for more details.                                                           *
+*                                                                             *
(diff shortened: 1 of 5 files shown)


Sofa/framework/Simulation/Core/src/sofa/simulation/DifferentialOperations.cpp
```diff
@@ -0,0 +1,77 @@
...
+    });
+}
+
+void DifferentialOperations::pullbackCotangent(
+    const MappingGraph& mappingGraph, const core::MechanicalParams& mparams,
+    core::MultiVecDerivId cotangentVectorId, bool ignoreMappingFlag)
+{
+    mappingGraph.algorithms.traverseBottomUp_([&](core::BaseMapping& mapping)
+    {
+        if (mapping.areForcesMapped() || ignoreMappingFlag)
```
damienmarchal, line 54: Shouldn't the areForcesMapped() methode renamed it seems a rather mechanical term.

I would also invert the ignoreMappingFlag and mapping.areForcesMapped() because the cost for quering  mapping.areForceMapped() is much more costly as it is virtual function, that is accessing a Data<bool>, so calling a bunch of other virtual function trigger and update the DDG).

alxbilger: yes of course

alxbilger:
[ci-build][with-all-tests]

fredroy (commented):
I guess adding unit tests would be tedious ? 👀

alxbilger:
> I guess adding unit tests would be tedious ? 👀

I will do it. It's a lot of boilerplate code, but that's doable.

bakpaul (approved):

PR #6311, opened 2026-09-17
[framework] Add TypeInfo for MatSym

th-skam:
bla2


fredroy (changes requested):
If the cpp is empty, I guess it is better to just omit it...

th-skam:
Well actually all of the *.cpp in that dir are empty. We can turn this into cleanup PR

fredroy (approved):

